PR #1, opened 2022-09-28
Initial version of print provider models and sequence diagram

kristerbone:
For discussion.


architecture/diagrams/print-sequence.plantuml
```diff
@@ -0,0 +1,60 @@
...
+"EROP Print Service"->"EROP Print Service Database": Get Print Request Statuses
+activate "EROP Print Service Database"
+"EROP Print Service"<-"EROP Print Service Database": Return Print Request Statuses
+deactivate "EROP Print Service Database"
+"EROP Print Service"->"EROP": Return Print Requests Statuses API
+deactivate "EROP Print Service"
+deactivate "EROP"
+
+
+@enduml
```
NathanRussellValtech, line 60: Good sequence diagram; I like 👍 😁

src/main/resources/openapi/PrintProviderAPI.yaml
```diff
@@ -2,120 +2,149 @@ openapi: 3.0.0
 info:
   title: Print ProviderAPI
   version: '1.0'
-  description: API for Print providers to retrieve print requests and to send updates on print request statuses.
+  description: API Specification for sharing Models between EIP and Print suppliers that will be exchanged over SFTP.
```
NathanRussellValtech, line 5: Given that our interface with print providers is SFTP, what "models" do we need to share with them?
What data structures do they need to know about?

Notwithstanding that (and this is something that @Neil-Massey and I are discussing at the moment in context of notifications API), openAPI is not the right tool for this.
openAPI spec is specifically for document REST APIs. Given that you have no `paths` elements, this spec is semantically invalid. Viewing it in swagger editor, or using a codegen tool will fail

Neil and I are considering this exact problem in notifications-api because we dont want the SQS models defined in the openAPI spec (because they are not related to REST APIs)

src/main/resources/openapi/PrintProviderAPI.yaml
```diff
@@ -2,120 +2,149 @@ openapi: 3.0.0
...
-      summary: ''
-      operationId: post-print-requests
-      responses:
-        '201':
-          description: Accepted
-      description: Updates the statuses of the print requests.
-      requestBody:
-        $ref: '#/components/requestBodies/PrintRequestUpdate'
-    parameters: []
 components:
```
NathanRussellValtech, line 11: I don't understand what these models are (in context of an SFTP interface). Who is using them? What for? And where has the design come from?

src/main/resources/openapi/PrintServiceAPI.yaml
```diff
@@ -0,0 +1,155 @@
+openapi: 3.0.0
+info:
+  title: Print Service API
+  version: '1.0'
```
NathanRussellValtech, line 4: We are using semantic versioning for openAPI specs (and we are actively maintaining the version number on any and all updates to a spec). Please can you set this to `1.0.0` please?

src/main/resources/openapi/PrintServiceAPI.yaml
```diff
@@ -0,0 +1,155 @@
...
+    name: Krister Bone
+    email: krister.bone@valtech.com
+servers:
+  - url: 'https://api.printprovider.int.erop.ierds.uk'
+components:
+  schemas:
+    PrintRequest:
+      title: PrintRequest
+      type: object
+      description: Represents a request to print a voter certificate. This is the SQS message to be sent to the Print Service from the Voter Card API Service
```
NathanRussellValtech, line 16: I don't understand what this spec if for, and what this model type is?
There are no `paths` elements in this spec, so is it a REST API?
As I understand it VCA etc will request a print via an SQS message. On that basis I'm guessing this `PrintRequest` model is the SQS message? If so it doesn't live in an openAPI spec like this (ref: the similar work @Neil-Massey and I are doing re: notifications-api)

NathanRussellValtech:
Thanks for starting this repo @kristerbone , I'm guessing its part of https://technologyprogramme.atlassian.net/browse/EIP1-1923 ?
Can we hold fire on this please? We've not yet started the epic (EIP-1226), and I think we need to be very careful about what we commit considering its a public repo/

src/main/resources/openapi/PrintServiceAPI.yaml
Neil-Massey: I'm not sure that this should be considered the `1.0.0` version of the spec - I assume this to be a "starter-for-ten" spec that will change once the features are refined and requirements become clearer.

I would prefer that the spec in `main/resources/api` are maintained by the developers when the API is developed, this ensures that it goes through the PR process when requirements are understood by the team.  Additionally, only endpoints that are actually implemented appear in this specification.

NathanRussellValtech: Good shout Neil, perhaps this should be in a `/migration' directory off of the top level (and not part of the src tree), with the filename being explicit that its is a starter for 10. Much like what we are proposing in notifications-api:
https://github.com/cabinetoffice/eip-ero-notifications-api/blob/feature/EIP1-557-initial-openapi-spec-for-sqs-messages/migration/starter-for-10-NotificationsAPIs.yaml

architecture/diagrams/print-sequence.plantuml
kristerbone: Thank you

src/main/resources/openapi/PrintProviderAPI.yaml
kristerbone: I can add to the sequence diagram.

src/main/resources/openapi/PrintServiceAPI.yaml
kristerbone: WIP - will need paths, just need to question what APIs are actually needed. But this currently is only SQS models.

src/main/resources/openapi/PrintProviderAPI.yaml
kristerbone: Removed and moved to YAML json schema

src/main/resources/openapi/PrintRequestSQS.yaml
```diff
@@ -0,0 +1,89 @@
+#version: 0.0.1
```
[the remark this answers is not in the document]
NathanRussellValtech: Actually, a slight correction/update to what I said above .... I have moved them to my PR, but they are not verbatim copies, and are no longer json/yaml schema.
I have migrated them to openAPI documents, consistent with what we are doing in this area in the notifications-api. Whilst they are not openAPI specs (they are not defining a REST API) we consider this a special use case ... 😁 

Basically @Neil-Massey and I have agreed that of all the technology/format/toolings options available to us, for SQS messages (which are for internal use and not shared externally) then openAPI ticks more boxes than it doesn't tick. What we like is that you can have all the models in one spec, and can therefore distribute one versioned spec file (to the other EROP applications). That one spec then contains all the models for that message type. The tooling is excellent - swagger editor gives a really good visual of the models; plus our codegen "just works"

We have looked at other options for SQS message types. We looked at AsyncAPI. Whilst the spec format is nice, the codegens are not! (they are a combination of nasty hacks that wrap an NPM library!)

We also looked at taking these yamlschema files verbatim. Whilst we got it working, we didn't like the fact that versioning is informal (via a comment), that you need a different spec file for each referenced type (eg: Address, ERO). We found a codegen that worked, but we also found a pretty nasty bug in it, and the code it generates was pretty ugly, especially in the area of the equals method and its constructors.
(having said this about yamlschema, I think we are happy to use it, and the codegen we found, for defining the CSV interface with the print suppliers - this seems a good match for that use case)

NathanRussellValtech (approved):
I'm happy to approve and merge this PR, but I don't think it's fully complete, and I don't think we'll get there within this one PR - its a living spec at the moment, and I think the best way forward is iterative changes.

I think the pragmatic thing to do at this stage would be to merge it, though please be aware that we will almost certainly:
* move `PrintServiceAPI`, `Address` and `ERO` into a "starter for 10" folder outside of the main src root because we already have an openAPI spec (that contains the health check endpoint at the moment), and we are more comfortable adding to the openAPI spec at the point we are working an actual endpoint implementation ticket. We will use the starter for 10 as inspiration/guidance at that point.
* move `PrintProviderSchema` into a sibling folder because it's not an openAPI spec. The folder name is likely to be `jsonschema` or similar, but TBA, and we will do it at the same time that we introduce the codegen for it.

On the assumption that we've acknowledged that this is not yet complete and still a WIP, presumably we'll also be making the Print Provider aware of that? IE. we are not publishing this spec to them yet as a finished article ?

PR #2, opened 2022-10-04
EIP1-1926 - Basic spring boot app

NathanRussellValtech:
This PR adds a very basic vanilla spring boot app with the actuator endpoint only.


build.gradle.kts
```diff
@@ -0,0 +1,166 @@
...
+    implementation("org.springframework.boot:spring-boot-starter-actuator")
+    implementation("org.springframework.boot:spring-boot-starter-web")
+    implementation("org.springdoc:springdoc-openapi-ui:1.6.11")
+    implementation("org.springframework.boot:spring-boot-starter-validation")
+
+    // spring security
+    implementation("org.springframework.boot:spring-boot-starter-security")
+    implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
+
+    // AWS v2 dependencies
```
Neil-Massey, line 66: Looks like we are using DynamoDB on print service too 👍

NathanRussellValtech:
> Are the .github/workflows already added or should they be included in this PR so that the microservice is deployed after this PR is merged?

I've written them (copied them from your fine work in notifications-api), but I deliberately didnt commit them on this PR because they rely on the infra PR.
My thinking was, get this basic code PR'ed and merged; get the infra merged and deployed; do a new PR for the GH workflows which will then use the infra
Does that make sense?

NathanRussellValtech:
Are we OK to approve and merge this @Neil-Massey ?
The infra PR is now approved and merged, but I'd still like to do the GH actions as a separate PR if thats OK?

Neil-Massey (approved):
LGTM - Also like the separation between build and deploy - I guess the build workflow could (and probably should) be included in this PR though

Neil-Massey:
> 

@NathanRussellValtech - yes, I like the separation between build and release and removing dependency from infra.  However the build and test workflow could be included at this stage in this PR (no biggie if already prepped in separate branch/PR)

PR #6, opened 2022-10-10
EIP1-1923: EIP1-1928: Add security configuration and localStack container

adacayi:
This PR is to add security configuration and localStack container to Print service.


NathanRussellValtech (approved):
LGTM 👍

PR #8, opened 2022-10-12
EIP1-1270 - Refining of SQS message SendApplicationToPrintMessage

NathanRussellValtech:
This PR makes a few refinement & corrections to the `SendApplicationToPrintMessage` introduced in my last PR.

Just for context, the `SendApplicationToPrintMessage` is an SQS message that will be consumed by `print-api` (service that owns the queue and therefore the API and this spec); and apps such as VCA (and proxy/postal/overseas as and when we get there) will publish this message. These apps will publish this message to instruct the `print-api` to send the application details to the print provider in order for them to print and post the physical Voter Authority Certificate. Therefore this `SendApplicationToPrintMessage` message needs to contain all the data necessary in order to print the certificate.

The "necessary data" is the data held by VCA (and the other future apps), and is essentially the applicant's name, photo, the date at which the certificate is issued (though there is a question about that - see the comment on the file), the delivery address (which may be the applicant's address, but may also be the ERO's address in the case of pick up from ERO), the ERO details (name, address, phone number, website etc - we don't currently hold this data in `ero-management` but it's coming ....). and the certificate language.

WRT certificate language - 2 languages are supported, English and Welsh. In the case of Welsh, it's actually a dual language certificate, where stuff is printed in English, then also in Welsh on the next line. This doesn't impact us particularly in respect of the data printed on the certificate because we don't provide this "non-placeholder" text. That's the responsibility of the print provider and their template. We simply need to tell them the language.
However, there is one notable exception to this. If the certificate is to be printed in Welsh (dual language) then we need to provide the ERO details (name, address, phone number, website etc) in Welsh so that both the English language and Welsh language name and address etc of the ERO can be printed. This is why there are 2 properties for ERO (snappily named `eroEnglish` and `eroWelsh`)

Data that is specifically not included is:
* DOB - this is not printed on the certificate
* NINo
* Applicant address (except in the case of it being the delivery address - I guess what I'm saying here is that we don't have a specific property/fields for "applicant address". Delivery Address is not the same as Applicant Address, though the data itself could be the same)
* Identity status / documents etc - these are used by the vc-admin to assert the identity of the applicant, but are not needed for the certificate itself
*

### DIFF
--- a/src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
+++ b/src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
@@ -42,19 +42,19 @@ components:
       description: The SQS message for sending an application to print.
       type: object
       properties:
-        voterApplicationId:
+        sourceReference:
           type: string
-          format: uuid
-          description: The Voter Authority Certificate identifier
-          example: 14ab70e8-bd3b-400f-bd95-246caf9e4810
+          description: Reference in the source application that this print request relates to
         applicationReference:
           type: string
           description: The application reference as known by the citizen. Not guaranteed to be unique.
+        sourceType:
+          $ref: '#/components/schemas/SourceType'
         gssCode:
           type: string
           minLength: 9
           maxLength: 9
-          description: GSS Code to ensure permissions can be maintained in Print Service API
+          description: GSS code of the ERO responsible for sending the message
         issuingAuthority:
           type: string
           maxLength: 255
@@ -63,61 +63,61 @@ components:
         issueDate:
           type: string
           format: date
-          description: The issue date of the Voter Authority Card
+          description: The issue date of the Voter Authority Certificate
           example: '2022-06-01'
         requestDateTime:
           type: string
           format: date-time
-          description: The date time the Voter Authority Certificate was requested
+          description: The date time the Voter Authority Certificate print request was requested
           example: '2022-06-01T12:23:03.000Z'
-        cardFirstname:
+        firstName:
           type: string
           maxLength: 255
           description: First name of the Elector
           example: John
-        cardMiddlenames:
+        middleNames:
           type: string
           maxLength: 255
           description: Middle names of the Elector
           example: Malcolm
-        cardSurname:
+        surname:
           type: string
           maxLength: 255
           description: Surname of the Elector
           example: Smith
         certificateLanguage:
           $ref: '#/components/schemas/CertificateLanguage'
-        deliveryOption:
-          $ref: '#/components/schemas/DeliveryOption'
-        photoS3Arn:
+        photoLocation:
           type: string
-          description: S3 ARN to image to send to printers.
+          description: The location of the Elector's photo. Typically an S3 ARN
           maxLength: 1024
-        deliveryName:
-          type: string
-          maxLength: 255
-          description: Name to be printed with delivery address
-        deliveryAddress:
-          $ref: '#/components/schemas/Address'
+        delivery:
+          $ref: '#/components/schemas/CertificateDelivery'
         eroEnglish:
           $ref: '#/components/schemas/ElectoralRegistrationOffice'
         eroWelsh:
           $ref: '#/components/schemas/ElectoralRegistrationOffice'
       required:
-        - voterApplicationId
+        - sourceReference
+        - sourceType
         - applicationReference
         - issuingAuthority
         - issueDate
         - requestDateTime
-        - cardFirstname
-        - cardMiddlenames
-        - cardSurname
+        - firstName
+        - middleNames
+        - surname
         - certificateLanguage
-        - deliveryOption
-        - photoS3Arn
-        - deliveryName
-        - deliveryAddress
+        - photoLocation
+        - delivery
         - eroEnglish
+
+    SourceType:
+      title: SourceType
+      description: An enumeration of sources for this print request
+      type: string
+      enum:
+        - voter-card
 
     CertificateLanguage:
       title: CertificateLanguage
@@ -127,6 +127,24 @@ components:
         - cy
         - en
       default: en
+
+    CertificateDelivery:
+      title: CertificateDelivery
+      description: Object describing how and where the Voter Authority Certificate should be delivered
+      type: object
+      properties:
+        deliveryOption:
+          $ref: '#/components/schemas/DeliveryOption'
+        addressee:
+          type: string
+          maxLength: 255
+          description: The addressee to be printed in addition to the delivery address
+        address:
+          $ref: '#/components/schemas/Address'
+      required:
+        - deliveryOption
+        - addressee
+        - address
 
     DeliveryOption:
       title: DeliveryOption
@@ -170,6 +188,10 @@ components:
         postcode:
           type: string
           maxLength: 10
+        uprn:
+          type: string
+          description: Unique Property Reference Number consisting of up to 12 digits in length
+          pattern: '^\d{1,12}$'
       required:
         - street
         - postcode
--- a/src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
+++ b/src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
@@ -60,11 +60,6 @@ components:
           maxLength: 255
           description: Issuing authority (LA/ERO/VJB)
           example: Camden Borough Council
-        issueDate:
-          type: string
-          format: date
-          description: The issue date of the Voter Authority Certificate
-          example: '2022-06-01'
         requestDateTime:
           type: string
           format: date-time
@@ -102,7 +97,6 @@ components:
         - sourceType
         - applicationReference
         - issuingAuthority
-        - issueDate
         - requestDateTime
         - firstName
         - middleNames
@@ -133,8 +127,8 @@ components:
       description: Object describing how and where the Voter Authority Certificate should be delivered
       type: object
       properties:
-        deliveryOption:
-          $ref: '#/components/schemas/DeliveryOption'
+        deliveryClass:
+          $ref: '#/components/schemas/DeliveryClass'
         addressee:
           type: string
           maxLength: 255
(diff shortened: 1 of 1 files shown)


src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -42,19 +42,19 @@ components:
...
+        sourceReference:
           type: string
-          format: uuid
-          description: The Voter Authority Certificate identifier
-          example: 14ab70e8-bd3b-400f-bd95-246caf9e4810
+          description: Reference in the source application that this print request relates to
         applicationReference:
           type: string
           description: The application reference as known by the citizen. Not guaranteed to be unique.
+        sourceType:
```
NathanRussellValtech, line 51: Consistent with RCA and notifications-api, I have added a property `sourceType` that tells us what type of application this is for (ie. what the source "application" was - though @matt-wills-valtech hates the term source *application* 😁 )
It's an enum (defined below), currently with one value in it (`voter-card`), so we'll be able to use this, along with the `sourceReference` to communicate back to the source *application* 😁 if we need to (or indeed filter by source type when we do the print queue etc)

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -42,19 +42,19 @@ components:
...
+          $ref: '#/components/schemas/SourceType'
         gssCode:
           type: string
           minLength: 9
           maxLength: 9
-          description: GSS Code to ensure permissions can be maintained in Print Service API
+          description: GSS code of the ERO responsible for sending the message
         issuingAuthority:
```
NathanRussellValtech, line 58: `issuingAuthority` - eg: "Camden Borough Council". This is printed on the certificate.
There is an open question that @kristerbone is looking to resolve with the business .... given that we need to support dual language do we need `issuingAuthorityEnglish` and `issuingAuthorityWelsh` ?
Alternatively (and better?), given that we already have `eroEnglish` and `eroWelsh` that have a `name` property, do we need `issuingAuthority` at all? Is the Issuing Authority the same as the ERO name ?

@kristerbone is looking to resolve this question , but I'm not sure what to do about in the immediate term. Half of me says "leave it and we'll update it when we have the answers to the above"; and the other half of me says "delete it now on the basis that I'm hoping the Issuing Authority is the same as the ERO name, as we can always add it back (along with the Welsh version) if we need to"

Thoughts @kristerbone ?

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -63,62 +63,62 @@ components:
         issueDate:
```
NathanRussellValtech, line 63: `issueDate` - a question for @kristerbone ....
Is VCA (and the other future apps) responsible for generating the Issue Date? Or, in the same way that `print-api` is responsible for generating the VAC number and suggested expiry date (issue date + 10 years), should `print-api` be responsible for generating the Issue Date ?

It seems to me that perhaps the date the certificate is issued is the date it was sent to print at the print provider, and not necessarily the date at which the source *application* [sic] sent it to the `print-api`

I'm probably splitting hairs over nothing here, but thoughts @kristerbone ?

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -63,62 +63,62 @@ components:
...
+          description: The issue date of the Voter Authority Certificate
           example: '2022-06-01'
         requestDateTime:
           type: string
           format: date-time
-          description: The date time the Voter Authority Certificate was requested
+          description: The date time the Voter Authority Certificate print request was requested
           example: '2022-06-01T12:23:03.000Z'
-        cardFirstname:
+        firstName:
```
NathanRussellValtech, line 73: @matt-wills-valtech @Neil-Massey ... "firstName", "middleNames", "surname" ... are these property names consistent with similar messages / fields in RCA and notifications-api ?

Consistency is a good thing 😁 👍

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -63,62 +63,62 @@ components:
...
           example: Malcolm
-        cardSurname:
+        surname:
           type: string
           maxLength: 255
           description: Surname of the Elector
           example: Smith
         certificateLanguage:
           $ref: '#/components/schemas/CertificateLanguage'
-        deliveryOption:
```
NathanRussellValtech, line 90: `deliveryOption`, `deliveryName` and `deliveryAddess` have been bundled up into a single property `delivery` of type `CertificateDelivery` (I'm open to suggestions of better names perhaps ?)

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -170,6 +188,10 @@ components:
         postcode:
           type: string
           maxLength: 10
+        uprn:
```
NathanRussellValtech, line 191: `uprn` was missing from the original `Address` type, even though it was described in the examples

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -42,19 +42,19 @@ components:
       description: The SQS message for sending an application to print.
       type: object
       properties:
-        voterApplicationId:
```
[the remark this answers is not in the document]
kristerbone: Yes, applicationId was a reference back to VCA App Id.

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
kristerbone: Happy for you to name this how you see consistent.

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
kristerbone: I will email about this today.

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
kristerbone: Happy for print-api to generate this. As above, if there does become another source, would keep issue date consistent.

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
kristerbone: We don't send this to the printers so it has no use in the print api (yet). Feel free to store it too.

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
Neil-Massey: Notifications API doesn't define these (the template personalisation allows the placeholder name and value to be provided by the caller)

adacayi (approved):
LGTM

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -128,6 +128,24 @@ components:
...
+          maxLength: 255
+          description: The addressee to be printed in addition to the delivery address
+        address:
+          $ref: '#/components/schemas/Address'
+      required:
+        - deliveryOption
+        - addressee
+        - address
+
     DeliveryOption:
```
Neil-Massey, line 149: Is `DeliveryClass` appropriate here?  (As in First class, Second class delivery options)

NathanRussellValtech: Yes, DeliveryClass is a much better name 👍 , thanks I'll do that 👍

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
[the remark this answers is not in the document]
NathanRussellValtech: Cool - so for consistency with our other apps and messages we'll call it `sourceReference` as explained above 👍 
Thanks

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
NathanRussellValtech: 👍

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
NathanRussellValtech: OK, I'll leave it in for now so we can merge this PR and move on. As and when we have an answer to this we can update this message type as necessary 👍

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
NathanRussellValtech: 👍 Cool, thanks, will remove this this spec then 👍

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
NathanRussellValtech: Yeah, thats an interesting one actually. For printing we wont use the UPRN, the printer would never use it, its not printed as part of the address. So there's a strong argument that says we dont need it / shouldnt accept it / not part of the message type.
But, the address description and example suggests we should include it as a field:
```
    Address:
      title: Address
      description: Address format used throughout gov.uk voter services.
      type: object
      x-examples:
        Minimum data:
          street: Street 1
          postcode: PC1 2FB
        Normal Address:
          street: East Lodge
          property: Balruddery
          locality: Invergowrie
          town: Dundee
          area: Angus
          postcode: DD25LF
          uprn: '117095813'
    ....
```
especially the description which says "Address format used throughout gov.uk voter services."

Given that its the same address format used in all the services I think it should include the field, but I take your point that we'll do nothing with it, so we'll consider how/if we store it in the DB when we get to that ticket 👍

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
NathanRussellValtech: I have checked in VCA wrt what it sends to RCA, and its is "firstName", "middleNames" and "surname"; so we are consistent 👍

PR #10, opened 2022-10-17
EIP1-1270: EIP1-2177: Remove issuing authority, eroEnglish and eroWelsh from the send application to print message

adacayi:
As discussed with @Neil-Massey  and @NathanRussellValtech , we don't want each microservice using the `print-api` to duplicate the code to get the relevant ero data when sending an application to print. 
Hence, we leave the responsibility of getting the relevant ero data to the `print-api`. 
`print-api`, with the following story, will get the ero data when consuming the `sendApplicationToPrint` queue messages, by getting the `gssCode` from the message and calling the `/eros` endpoint of the `ero-management-api` with `gssCode` as the query parameter.

Based on this decision, this PR removes the now unnecessary fields from the sqs message.


NathanRussellValtech (approved):
LGTM

PR #12, opened 2022-10-19
EIP1-1878 - codegen print provider schema

NathanRussellValtech:
This PR implements the codegen for the print provider schema.

The initial print provider schema was added in PR https://github.com/cabinetoffice/eip-ero-print-api/pull/1

The schema files have been moved as previously agreed with @kristerbone on PR https://github.com/cabinetoffice/eip-ero-print-api/pull/1 :
* The REST API spec and associated models (openAPI) have been moved to `migration` folder as starter for 10's. We will migrate these into the real spec as and when we do the implementations
* The Print Provider schema file has been moved to `yamlschema` folder (because its not an openAPI spec), and the file has been split into separate files for each type (to make it valid yaml schema)

Codegen has been implemented using [js2d-gradle ](https://github.com/jsonschema2dataclass/js2d-gradle)


migration/starter-for-10-PrintServiceAPI.yaml
```diff
@@ -1,3 +1,12 @@
+#
+# --------------------------------------------------------------------------------
+# This is the original starter-for-10 openAPI spec with initial API definitions
+# for Print Service API. As and when we design and implement each endpoint it
+# should be migrated from this spec into `PrintAPIs.yaml` (even if we change
+# it a lot!), with the end result being this file is empty and deleted.
+# --------------------------------------------------------------------------------
+#
+
 openapi: 3.0.0
```
NathanRussellValtech, line 10: This is the initial openAPI spec that was added in PR https://github.com/cabinetoffice/eip-ero-print-api/pull/1 
As agreed I have moved it here, and added this comment block

src/main/resources/yamlschema/BatchResponse.yaml
```diff
@@ -0,0 +1,25 @@
+#version: 0.0.3
```
NathanRussellValtech, line 1: All of these "new" files are split out from `PrintProviderSchema.yaml` above, which was initially added in PR https://github.com/cabinetoffice/eip-ero-print-api/pull/1

build.gradle.kts
```diff
@@ -143,6 +144,24 @@ tasks.create("generate-models-from-openapi-document-print-api-sqs-messaging.yaml
     packageName.set("uk.gov.dluhc.printapi.messaging")
 }
 
+// Codegen the Print Provider schemas from yamlschema into java pojos
+jsonSchema2Pojo {
```
NathanRussellValtech, line 148: This is the same codegen plugin I showed you the other week @Neil-Massey  , but without that nasty hack I had to do re: reversing properties because of a bug with their builder and constructor args order ...
There was a new release on Oct 17th (4.5.0) which seems to have fixed that, which is good, but it also means the plugin is being actively maintained 👍

src/main/resources/yamlschema/BatchResponse.yaml
```diff
@@ -0,0 +1,25 @@
...
+title: BatchResponse
+type: object
+description: Represents the response from the print provider in response to recieving a batch file from EROP
+properties:
+  batchId:
+    type: string
+    example: 1000-2000
```
Neil-Massey, line 8: Is this a good example?

src/main/resources/yamlschema/PrintRequest.yaml
```diff
@@ -0,0 +1,203 @@
...
+  erodDeliveryAreaCy:
+    type: string
+    maxLength: 255
+  eroDeliveryPostcodeCy:
+    type: string
+    maxLength: 10
+required:
+  - requestId
+  - issuingAuthorityEn
+  - issuingAuthorityCy
```
Neil-Massey, line 183: Is this really required?  (when rest of the welsh transalations are not?)

src/main/resources/yamlschema/PrintRequest.yaml
```diff
@@ -0,0 +1,203 @@
...
+  - photo
+  - deliveryName
+  - deliveryStreet
+  - deliveryPostcode
+  - eroNameEn
+  - eroPhoneNumberEn
+  - eroEmailAddressEn
+  - eroWebsiteEn
+  - eroDeliveryStreetEn
+  - eroDeliveryPostcodeEn
```
Neil-Massey, line 203: I assume this is a CSV, so the header will always include all names and this `required` just relates to the data included in a row

src/main/resources/yamlschema/PrintResponse.yaml
```diff
@@ -0,0 +1,38 @@
...
+    description: The date time the Voter Authority Certificate print request was updated to this status.
+    example: '2022-06-01T14:23:03.000Z'
+  statusStep:
+    type: string
+    description: The current print step this update relates to
+    enum:
+      - PROCESSED
+      - IN-PRODUCTION
+      - DISPATCHED
+      - NOT-DELIVERED
```
Neil-Massey, line 23: Should there be a `DELIVERED`?  (I assume `PROCESSED` is before going into production - like accepted for printing)

build.gradle.kts
Neil-Massey: Beast of options... at least they now set what they should :-)

Neil-Massey (commented):
Few questions/comments

src/main/resources/yamlschema/PrintResponse.yaml
NathanRussellValtech: I think `DISPATCHED` means delivered (effectively). We wont know (and nor will the print provider) that its actually been delivered, unless the postman tells us. I dont think we're doing recorded delivery or similar, so once its been `DISPATCHED` thats the last anyone knows of it ...... except ....

... in the case that the postie can't deliver it, then it gets returned to sender, which is the print provider. In that case the print provider send us another status update saying `NOT-DELIVERED`

src/main/resources/yamlschema/BatchResponse.yaml
NathanRussellValtech: No, it's a terrible example. But it's what was in the original spec file (`PrintProviderSchema.yaml`) that Krister has put together with the Print Provider.
The scope of my work here was getting it to a point of being able to codegen it. I think there will be some subsequent refinement of these specs over the next week or so to get them into a state that we are happy with, but that is not this ticket.

src/main/resources/yamlschema/PrintRequest.yaml
NathanRussellValtech: Yes, absolutely.
It will be a CSV (well, a PSV actually - pipe separated values, but its the same thing! 😁 ). The header row will include all columns (assuming we have a header row - not sure if thats been discussed yet). This `required` attribute is really just there to define which fields are mandatory or not. The codegen will add these fields to the constructor (its a java pojo rather than a kotlin class with non nullable fields). Its about defining the API contract in respect of field cardinality in the CSV

src/main/resources/yamlschema/PrintResponses.yaml
```diff
@@ -0,0 +1,15 @@
+#version: 0.0.3
+title: PrintResponses
```
NathanRussellValtech, line 2: `PrintResponses` ... this is an interesting one @kristerbone ....
Given that it's a type that contains 2 properties which are both arrays of other types ..... how will this get serialized as a flat PSV ? 🤔

Neil-Massey (approved):

PR #13, opened 2022-10-21
EIP1-1364 - Introduction of ID class similar to mongo ObjectId (12 byte ID)

NathanRussellValtech:
This PR introduces a new "id" class which is very similar (heavily based on) Mongo's `ObjectId` class.
The reason for this is as follows:

* `print-api` needs to generate an ID as part of writing a record to it's database (which is why the class is in the `database` package - the rest of the DB stuff will come later)
* The ID itself has some specific requirements in that it has to be less than 25 characters (because it will be printed on an envelope and there is a physical restriction as to the number or printable characters)
* A mongo ID was suggested as a suitable ID because its is a 12 byte / 24 character (hex string) ID

I am not keen on using a mongo ID because it means we have a gradle dependency on their jar , which adds bloat to our distributable, plus loads of classes we wont be using; and of course it feels plain wrong to have a mongo dependency when we are not using mongo!

What I've done here is copy & paste 😁 the relevant parts of mongo's `ObjectId` class into a class of our own. Their class has more methods and functionality than this proposed version. This is very much a cut down version.

Now, there are some bits I'm not entirely happy with and would welcome some thoughts / opinions :

1. Should we even bother? Am I losing too much sleep over using the mongo class and dependency?
2. What to call it and what package? 
Naming is hard!. I've called it `TwelveByteId` because thats what it is (`ObjectId` is only relevant and makes sense in a mongo environment, so I don't want to call it that)
In terms of its use case in `print-api` it will only be used as part of creating a database record, so I've put it in the `database` package. Don't know if that is right? In terms of what it represents in the `print-api` database record it will be a `requestId`, so perhaps the classname should be `RequestId` ?
3. Unit tests! I can't write any for it! (not any meaningful ones anyway)
mongo have written unit tests for their `ObjectId` class .... but they really only work because they've got overloaded public constructors, where you can construct an `ObjectId` from a variety of factors, including the 12 bytes that form the ID!
To me it feels like they've created constructors for the purpose of tests 😢 
4. Constructor overloads .... related to the above, I have implemented a number of constructor overloads because I copy and pasted what mongo do. The difference is that I've made the constructor overloads private; and really I want to get rid of them altogether, and just do everything in the main no-args constructor.
It was very much a case of copy method by method, constructor by constructor; just to get the basic class working. I dont want to implement everything that they do.

There are some other questions that come up from this as well ....
Can we get re-use from this? At the moment we don't have a shared/common library repo; but VCA has a similar requirement. I wonder if @vishalgupta-valtech and @sentioservices could make use of a copy of this class re: their PR https://github.com/cabinetoffice/eip-ero-voter-card-applications-api/pull/232#discussion_r996938364

Open to the floor - what do we think?

---

Mongo classes:
`ObjectId`: https://github.com/mongodb/mongo-java-driver/blob/master/bson/src/main/org/bson/types/ObjectId.java
`ObjectIdTest`:  https://github.com/mongodb/mongo-java-driver/blob/master/bson/src/test/unit/org/bson/types/ObjectIdTest.java

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/database/TwelveByteId.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/database/TwelveByteId.kt
@@ -1,0 +1,157 @@
+package uk.gov.dluhc.printapi.database
+
+import java.nio.ByteBuffer
+import java.security.SecureRandom
+import java.util.Date
+import java.util.concurrent.atomic.AtomicInteger
+
+/**
+ * A globally unique identifier, heavily based on [mongodb's ObjectId class](https://github.com/mongodb/mongo-java-driver/blob/master/bson/src/main/org/bson/types/ObjectId.java)
+ *
+ * Consists of 12 bytes, divided as follows:
+ *
+ * ```
+ *   | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 |
+ *   |---------------|-----------|-------|-------------|
+ *   | time          |rnd 1      | rnd 2 | increment   |
+ * ```
+ *
+ * Where:
+ *   * `time` is the timestamp since epoch rounded to second precision. The first 4 bytes of the integer are used.
+ *   * `rnd 1` is a random integer. This is a static member. The first 3 bytes of the integer are used.
+ *   * `rnd 2` is a random short. This is a static member. The first 2 bytes of the short are used.
+ *   * `increment` is a sequential increment, starting from a random integer. The first 3 bytes of the integer are used.
+ *
+ *  `rnd 1` and `rnd 2` are static members, meaning that all instances of [TwelveByteId] generated in the same JVM will have
+ *  the same `rnd 1` and `rnd 2` values. The net result is that bytes [4, 5, 6, 7, 8] of 2 instances generated in the same JVM
+ *  will be the same.
+ *
+ *  `increment` is an [AtomicInteger] and is a static member seeded from a random number. The use of `getAndIncrement()` means
+ *  this is a thread safe way of generating an incremental counter.
+ *
+ * An instance of [TwelveByteId] is a 12 byte identifier which serializes via it's `toString()` method into a 24 character hex string.
+ */
+class TwelveByteId {
+
+    private val timestamp: Int
+    private val counter: Int
+    private val randomValue1: Int
+    private val randomValue2: Short
+
+    companion object {
+        private const val OBJECT_ID_LENGTH = 12
+        private const val LOW_ORDER_THREE_BYTES = 0x00ffffff
+
+        // Use primitives to represent the 5-byte random value.
+        private val RANDOM_VALUE1: Int = SecureRandom().nextInt(0x01000000)
+        private val RANDOM_VALUE2: Short = SecureRandom().nextInt(0x00008000).toShort()
+
+        private val NEXT_COUNTER = AtomicInteger(SecureRandom().nextInt())
+
+        private val HEX_CHARS = charArrayOf(
+            '0', '1', '2', '3', '4', '5', '6', '7',
+            '8', '9', 'a', 'b', 'c', 'd', 'e', 'f'
+        )
+
+        private fun dateToTimestampSeconds(date: Date): Int =
+            (date.time / 1000).toInt()
+
+        fun get(): TwelveByteId = TwelveByteId()
+    }
+
+    constructor() : this(Date())
+
+    private constructor(date: Date) : this(
+        dateToTimestampSeconds(date),
+        NEXT_COUNTER.getAndIncrement() and LOW_ORDER_THREE_BYTES,
+        false
+    )
+
+    private constructor(timestamp: Int, counter: Int, checkCounter: Boolean) : this(
+        timestamp,
+        RANDOM_VALUE1,
+        RANDOM_VALUE2,
+        counter,
+        checkCounter
+    )
+
+    private constructor(timestamp: Int, randomValue1: Int, randomValue2: Short, counter: Int, checkCounter: Boolean) {
+        require(randomValue1 and -0x1000000 == 0) { "The random value must be between 0 and 16777215 (it must fit in three bytes)." }
+        require(!(checkCounter && counter and -0x1000000 != 0)) { "The counter must be between 0 and 16777215 (it must fit in three bytes)." }
+        this.timestamp = timestamp
+        this.counter = counter and LOW_ORDER_THREE_BYTES
+        this.randomValue1 = randomValue1
+        this.randomValue2 = randomValue2
+    }
+
+    override fun toString(): String = toHexString()
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (javaClass != other?.javaClass) return false
+
+        other as TwelveByteId
+
+        if (timestamp != other.timestamp) return false
+        if (counter != other.counter) return false
+        if (randomValue1 != other.randomValue1) return false
+        if (randomValue2 != other.randomValue2) return false
+
+        return true
+    }
+
+    override fun hashCode(): Int {
+        var result = timestamp
+        result = 31 * result + counter
+        result = 31 * result + randomValue1
+        result = 31 * result + randomValue2
+        return result
+    }
+
+    private fun toHexString(): String {
+        val chars = CharArray(OBJECT_ID_LENGTH * 2)
+        var i = 0
+        for (b in toByteArray()) {
+            chars[i++] = HEX_CHARS[b.toInt() shr 4 and 0xF]
+            chars[i++] = HEX_CHARS[b.toInt() and 0xF]
+        }
+        return String(chars)
+    }
+
+    private fun toByteArray(): ByteArray {
+        val buffer = ByteBuffer.allocate(OBJECT_ID_LENGTH)
+        putToByteBuffer(buffer)
+        return buffer.array()
+    }
+
+    private fun putToByteBuffer(buffer: ByteBuffer) {
+        requireNotNull(buffer) { "buffer cannot be null" }
+        require(buffer.remaining() >= OBJECT_ID_LENGTH) { "buffer.remaining() >=12" }
+
+        buffer
+            .put(int3(timestamp))
+            .put(int2(timestamp))
+            .put(int1(timestamp))
+            .put(int0(timestamp))
+            .put(int2(randomValue1))
+            .put(int1(randomValue1))
+            .put(int0(randomValue1))
+            .put(short1(randomValue2))
+            .put(short0(randomValue2))
+            .put(int2(counter))
+            .put(int1(counter))
+            .put(int0(counter))
+    }
+
+    private fun int3(x: Int): Byte = (x shr 24).toByte()
+
+    private fun int2(x: Int): Byte = (x shr 16).toByte()
+
+    private fun int1(x: Int): Byte = (x shr 8).toByte()
+
+    private fun int0(x: Int): Byte = x.toByte()
+
+    private fun short1(x: Short): Byte = (x.toInt() shr 8).toByte()
+
+    private fun short0(x: Short): Byte = x.toByte()
+}


src/main/kotlin/uk/gov/dluhc/printapi/database/TwelveByteId.kt
```diff
@@ -0,0 +1,157 @@
+package uk.gov.dluhc.printapi.database
+
+import java.nio.ByteBuffer
+import java.security.SecureRandom
+import java.util.Date
+import java.util.concurrent.atomic.AtomicInteger
+
+/**
```
NathanRussellValtech, line 8: IMHO 😁  this kdoc is better than the docs on mongo's `ObjectId` class. It explains the algorithm better I think.

As a side note it was actually quite good doing this work because it meant I had to understand how their class worked and understand the algorithm. It's really quite a neat approach, and is more than just "generate a random number / hex string" 👍

src/main/kotlin/uk/gov/dluhc/printapi/database/TwelveByteId.kt
```diff
@@ -0,0 +1,157 @@
...
+
+        private fun dateToTimestampSeconds(date: Date): Int =
+            (date.time / 1000).toInt()
+
+        fun get(): TwelveByteId = TwelveByteId()
+    }
+
+    constructor() : this(Date())
+
+    private constructor(date: Date) : this(
```
NathanRussellValtech, line 64: These private constructors should all get wrapped up into the main no-args constructors, but it was a case of copying mongo's class to get the basics working 👍

adacayi (commented):
Hi @NathanRussellValtech ,

I have found this project using the same logic to generate a globally unique identifier.
What do you think?

https://github.com/rs/xid

src/main/kotlin/uk/gov/dluhc/printapi/database/TwelveByteId.kt
Neil-Massey: It's also different to: https://www.mongodb.com/developer/products/mongodb/bson-data-types-objectid/

src/main/kotlin/uk/gov/dluhc/printapi/database/TwelveByteId.kt
```diff
@@ -0,0 +1,157 @@
...
+ *  `rnd 1` and `rnd 2` are static members, meaning that all instances of [TwelveByteId] generated in the same JVM will have
+ *  the same `rnd 1` and `rnd 2` values. The net result is that bytes [4, 5, 6, 7, 8] of 2 instances generated in the same JVM
+ *  will be the same.
+ *
+ *  `increment` is an [AtomicInteger] and is a static member seeded from a random number. The use of `getAndIncrement()` means
+ *  this is a thread safe way of generating an incremental counter.
+ *
+ * An instance of [TwelveByteId] is a 12 byte identifier which serializes via it's `toString()` method into a 24 character hex string.
+ */
+class TwelveByteId {
```
Neil-Massey, line 34: Maybe `CompactUuid`

NathanRussellValtech: Ohh yeah , that is a great suggestion 👍  I like that

src/main/kotlin/uk/gov/dluhc/printapi/database/TwelveByteId.kt
```diff
@@ -0,0 +1,157 @@
+package uk.gov.dluhc.printapi.database
```
Neil-Massey, line 1: For comparison `java.util.UUID` - maybe `util` is an appropriate package

src/main/kotlin/uk/gov/dluhc/printapi/database/TwelveByteId.kt
```diff
@@ -0,0 +1,157 @@
...
+
+    private fun int2(x: Int): Byte = (x shr 16).toByte()
+
+    private fun int1(x: Int): Byte = (x shr 8).toByte()
+
+    private fun int0(x: Int): Byte = x.toByte()
+
+    private fun short1(x: Short): Byte = (x.toInt() shr 8).toByte()
+
+    private fun short0(x: Short): Byte = x.toByte()
```
Neil-Massey, line 156: Cannot comment on the correctness of the algorithm - it would be easy to have a typo somewhere.
Although I understand point about unit testing this class is dififcult due to all the randomness - as a one of test say 10,000 random values could be generated and verified no clashes occured.  (I'm sure you must have done some verification already)

NathanRussellValtech:
> Hi @NathanRussellValtech ,
> 
> I have found this project using the same logic to generate a globally unique identifier. What do you think?
> 
> https://github.com/rs/xid

@adacayi - thanks for this.
Yes, it looks to be essentially the same thing 👍 
Its not java though, it's go. Do you know if we can use go libraries in a java project? Can we get it via a gradle dependency somehow?

Neil-Massey:
> 

https://mvnrepository.com/artifact/com.github.0xshamil/java-xid

adacayi:
> > Hi @NathanRussellValtech ,
> > I have found this project using the same logic to generate a globally unique identifier. What do you think?
> > https://github.com/rs/xid
> 
> @adacayi - thanks for this. Yes, it looks to be essentially the same thing 👍 Its not java though, it's go. Do you know if we can use go libraries in a java project? Can we get it via a gradle dependency somehow?

Sure, @NathanRussellValtech ,

This is the gradle implementation: 
<pre>implementation("com.github.0xshamil:java-xid:1.0.0")</pre>

Code:
<pre>
fun getAMongoDbId(): String = Hex.encodeHexString(Xid().toByteArray())
</pre>

vishalgupta-valtech:
> 
@adacayi @NathanRussellValtech Just trying to understand, how is this code from `Xid()` library `Hex.encodeHexString(Random.nextBytes(24))` different from the below code that Kotlin provides ?

`Hex.encodeHexString(Random.nextBytes(24))`

vishalgupta-valtech:
> fun getAMongoDbId(): String = Hex.encodeHexString(Xid().toByteArray())

@adacayi @NathanRussellValtech Just trying to understand, how is this code from Xid() library different from the below code that Kotlin provides ?

Hex.encodeHexString(Random.nextBytes(24))

adacayi:
> > fun getAMongoDbId(): String = Hex.encodeHexString(Xid().toByteArray())
> 
> @adacayi @NathanRussellValtech Just trying to understand, how is this code from Xid() library different from the below code that Kotlin provides ?
> 
> `Hex.encodeHexString(Random.nextBytes(24))`



> > fun getAMongoDbId(): String = Hex.encodeHexString(Xid().toByteArray())
> 
> @adacayi @NathanRussellValtech Just trying to understand, how is this code from Xid() library different from the below code that Kotlin provides ?
> 
> `Hex.encodeHexString(Random.nextBytes(24))`

Hi @vishalgupta-valtech ,
The difference is the way it is generated to ensure the value is globally unique.

Here is the explanation of how it is generated:

> Xid uses the Mongo Object ID algorithm to generate globally unique ids with a different serialization (base64) to make it shorter when transported as a string: https://docs.mongodb.org/manual/reference/object-id/
> 
> 4-byte value representing the seconds since the Unix epoch,
> 3-byte machine identifier,
> 2-byte process id, and
> 3-byte counter, starting with a random value.

Source: https://github.com/rs/xid

NathanRussellValtech:
> > > fun getAMongoDbId(): String = Hex.encodeHexString(Xid().toByteArray())
> > 
> > 
> > @adacayi @NathanRussellValtech Just trying to understand, how is this code from Xid() library different from the below code that Kotlin provides ?
> > `Hex.encodeHexString(Random.nextBytes(24))`
> 
> > > fun getAMongoDbId(): String = Hex.encodeHexString(Xid().toByteArray())
> > 
> > 
> > @adacayi @NathanRussellValtech Just trying to understand, how is this code from Xid() library different from the below code that Kotlin provides ?
> > `Hex.encodeHexString(Random.nextBytes(24))`
> 
> Hi @vishalgupta-valtech , The difference is the way it is generated to ensure the value is globally unique.
> 
> Here is the explanation of how it is generated:
> 
> > Xid uses the Mongo Object ID algorithm to generate globally unique ids with a different serialization (base64) to make it shorter when transported as a string: https://docs.mongodb.org/manual/reference/object-id/
> > 4-byte value representing the seconds since the Unix epoch,
> > 3-byte machine identifier,
> > 2-byte process id, and
> > 3-byte counter, starting with a random value.
> 
> Source: https://github.com/rs/xid

And I don't know why you'd need to wrap the call in `Hex.encodeHexString` ?? eg:
```
fun getAMongoDbId(): String = Hex.encodeHexString(Xid().toByteArray())
```
Why would it not just be
```
fun getAMongoDbId(): String = Xid().toString()
```

NathanRussellValtech:
Some great discussion on this, thanks all 👍 

It looks like `Xid` does exactly what my proposed class does; and I think there's a lot to be said for that, especially as it includes unit tests (that look a lot like the mongo unit tests, mainly because there are public overloaded constructors 😁 )

Question .... 

Are we happy to use "some-bloke.com" code? (ie. code from some random dev - it's not from an official or trusted source). Is this an issue?
Are we happy that it's v1.0.0 and has not been maintained in 2 years? (though perhaps there's no need for continual maintenance on a class like this?)

What do we think?
(In the context of secure/trusted dependencies I specifically told James R the other week that we would never take a "some-bloke.com" dependency ¯\_(ツ)_/¯ )

vishalgupta-valtech:
> What do we think?
> (In the context of secure/trusted dependencies I specifically told James R the other week that we would never take a "some-bloke.com" dependency ¯_(ツ)_/¯ )

+1 Although the library looks promising, however I am unsure if we need to use the library that was updated once by a random bloke 2 years ago?

sentioservices:
Hi @adacayi ,
A minor point but this looks like the old way that the id was generated.

> > Xid uses the Mongo Object ID algorithm to generate globally unique ids with a different serialization (base64) to make it shorter when transported as a string: https://docs.mongodb.org/manual/reference/object-id/
> > 4-byte value representing the seconds since the Unix epoch,
> > 3-byte machine identifier,
> > 2-byte process id, and
> > 3-byte counter, starting with a random value.
> 
> Source: https://github.com/rs/xid

Elements 2 and 3 have been combined into "a 5-byte random number unique to a machine and process," as per https://github.com/mongodb/specifications/blob/master/source/objectid.rst

Neil-Massey:
The BSON dependency ('org.mongodb:bson:4.7.1') is only 491kB - which doesn't add much bloat and to be honest I think I'd propose using that directly (in preference to rolling our own copy or relying on a "some-bloke.com jar).

adacayi:
> > > > fun getAMongoDbId(): String = Hex.encodeHexString(Xid().toByteArray())
> > > 
> > > 
> > > @adacayi @NathanRussellValtech Just trying to understand, how is this code from Xid() library different from the below code that Kotlin provides ?
> > > `Hex.encodeHexString(Random.nextBytes(24))`
> > 
> > 
> > > > fun getAMongoDbId(): String = Hex.encodeHexString(Xid().toByteArray())
> > > 
> > > 
> > > @adacayi @NathanRussellValtech Just trying to understand, how is this code from Xid() library different from the below code that Kotlin provides ?
> > > `Hex.encodeHexString(Random.nextBytes(24))`
> > 
> > 
> > Hi @vishalgupta-valtech , The difference is the way it is generated to ensure the value is globally unique.
> > Here is the explanation of how it is generated:
> > > Xid uses the Mongo Object ID algorithm to generate globally unique ids with a different serialization (base64) to make it shorter when transported as a string: https://docs.mongodb.org/manual/reference/object-id/
> > > 4-byte value representing the seconds since the Unix epoch,
> > > 3-byte machine identifier,
> > > 2-byte process id, and
> > > 3-byte counter, starting with a random value.
> > 
> > 
> > Source: https://github.com/rs/xid
> 
> And I don't know why you'd need to wrap the call in `Hex.encodeHexString` ?? eg:
> 
> ```
> fun getAMongoDbId(): String = Hex.encodeHexString(Xid().toByteArray())
> ```
> 
> Why would it not just be
> 
> ```
> fun getAMongoDbId(): String = Xid().toString()
> ```

The reason is 

> > 
> 
> There is a Java version: https://github.com/0xShamil/java-xid https://mvnrepository.com/artifact/com.github.0xshamil/java-xid

Hi @NathanRussellValtech ,

The reason is, as explained in the [documentation ](https://github.com/0xShamil/java-xid), they store it as Base32 encoded string to save space. To get the hex encoded 24 digit represantation, I converted it to byte array and then encoded in hex format.

>The string representation is using [base32 hex (w/o padding)](https://tools.ietf.org/html/rfc4648#page-10) for better space >efficiency when stored in that form (20 bytes)

Here is an output of the same object in two different forms:

<pre>
        val id = Xid()
        println("To String   = ${id.toString()}")
        println("Hex encoded = ${Hex.encodeHexString(id.toByteArray())}")
</pre>

<pre>
To String   = cd9ac2hlnehnclcn87gg
Hex encoded = 6352a60a35bba376559741e1
<pre>

adacayi:
> The BSON dependency ('org.mongodb:bson:4.7.1') is only 491kB - which doesn't add much bloat and to be honest I think I'd propose using that directly (in preference to rolling our own copy or relying on a "some-bloke.com jar).

I agree with @Neil-Massey. 
If we don't see it as much bloat, then it is better to go with the bson library.

NathanRussellValtech:
> Hi @NathanRussellValtech ,
> 
> The reason is, as explained in the [documentation ](https://github.com/0xShamil/java-xid), they store it as Base32 encoded string to save space. To get the hex encoded 24 digit represantation, I converted it to byte array and then encoded in hex format.
> 
> > The string representation is using [base32 hex (w/o padding)](https://tools.ietf.org/html/rfc4648#page-10) for better space >efficiency when stored in that form (20 bytes)
> 
> Here is an output of the same object in two different forms:
> 
> val id = Xid()
> println("To String   = ${id.toString()}")
> println("Hex encoded = ${Hex.encodeHexString(id.toByteArray())}")
> To String   = cd9ac2hlnehnclcn87gg
> Hex encoded = 6352a60a35bba376559741e1

Oh that blows chunks! If we have to wrap it a hex encoder to get the output format we need .... no!

"blows chunks" - you have to be a certain age 😁 (makes me want to watch the film again though 🤣 )
image

NathanRussellValtech:
> > The BSON dependency ('org.mongodb:bson:4.7.1') is only 491kB - which doesn't add much bloat and to be honest I think I'd propose using that directly (in preference to rolling our own copy or relying on a "some-bloke.com jar).
> 
> I agree with @Neil-Massey. If we don't see it as much bloat, then it is better to go with the bson library.

OK, agreed. The scope of this ticket was:
> Maybe a timeboxed bit of investigation as to whether we can generate a similar ID in a different way (max 1/2 day); otherwise I guess we’ll have to use the mongo libraries :disappointed: 
With that in mind I am well over 1/2 on this, so mongo it is 😢 

@vishalgupta-valtech , if you want to change your use of the hexencoder random thingy back to mongo's ObjectId as you originally had in your PR then go for it 👍 . I think a "globally unique identifier" is better than a random value 👍

NathanRussellValtech:
Closing this PR as we will not be proceeding with this approach

PR #14, opened 2022-10-21
EIP1-1364 - Introduction of IdFactory to build requestIds

NathanRussellValtech:
Following the decision in my earlier to PR to use the mongo library and its `ObjectId` class, this PR introduces a simple "ID Factory" that has methods to build a `requestId` (mongo ObjectId), and also a method ready to be implemented to build a VAC Certificate Reference Number (which is essentially an ID, hence its in the ID Factory)

The factory is a bean because it will be injected into the components that use it. I could have opted for a simple file exposing simple functions, but a bean felt better as it will be easier to mock for the purpose of tests (we would have difficulty with assertions if it were a function that could not be mocked); plus the generation of a certificate reference via `certificateReference` is likely to be fairly complex and may involve other collaborators (not sure yet, but again, the mocking argument wins 😁 )

### DIFF
--- a/src/test/kotlin/uk/gov/dluhc/printapi/service/IdFactoryTest.kt
+++ b/src/test/kotlin/uk/gov/dluhc/printapi/service/IdFactoryTest.kt
@@ -1,0 +1,32 @@
+package uk.gov.dluhc.printapi.service
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.api.Test
+
+internal class IdFactoryTest {
+
+    private val idFactory = IdFactory()
+
+    @Test
+    fun `should generate requestId`() {
+        // Given
+
+        // When
+        val requestId = idFactory.requestId()
+
+        // Then
+        assertThat(requestId).hasSize(24)
+    }
+
+    @Test
+    fun `should generate unique requestId with each call`() {
+        // Given
+        val requestIds = mutableSetOf<String>() // store request IDs in a set to ensure unique entries
+
+        // When
+        repeat(10) { requestIds.add(idFactory.requestId()) }
+
+        // Then
+        assertThat(requestIds).hasSize(10) // 10 elements in the set mean that there were no duplicates
+    }
+}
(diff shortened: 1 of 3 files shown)


Neil-Massey (approved):
LGTM

src/test/kotlin/uk/gov/dluhc/printapi/service/IdFactoryTest.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    @Test
+    fun `should generate requestId`() {
+        // Given
+
+        // When
+        val requestId = idFactory.requestId()
+
+        // Then
+        assertThat(requestId).hasSize(24)
```
adacayi, line 18: One minor suggestion, @NathanRussellValtech . 
We can assert the pattern as well.

<pre>
assertThat(requestId).matches(Regex("[\\da-f]{24}").toPattern())
 </pre>

adacayi (approved):
Added one minor suggestion.

src/test/kotlin/uk/gov/dluhc/printapi/service/IdFactoryTest.kt
NathanRussellValtech: Yeah, I did consider something like that, but in the end I decided that we might be testing mongo's code rather than ours?
I think the thing that's important to us in respect of generating a request ID is that its 24 chars - that is the business rule that we need to fulfil. The format hasn't been defined as a business rule/requirement. We just need it to be 24 chars, and unique. Which I think the 2 tests in this class cover?

adacayi: I think the extra assertion is good to check that we are using mongo's code, rather than a random alphanumeric generator, but still, you can generate random values with the same hex format and they will most probably pass our uniqueness test. So not too worried about the current state of the test. Also, you pointed out the format hasn't been decided yet, so I am happy with it as well.

PR #15, opened 2022-10-21
EIP1-1364: EIP1-2213: Implement ero client to acquire ero data from the ero management api

adacayi:
This PR is to implement an ero client to acquire ero data from the ero management api for a given gssCode.

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClient.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClient.kt
@@ -1,0 +1,63 @@
+package uk.gov.dluhc.printapi.client
+
+import mu.KotlinLogging
+import org.springframework.http.HttpStatus.NOT_FOUND
+import org.springframework.stereotype.Component
+import org.springframework.web.reactive.function.client.WebClient
+import org.springframework.web.reactive.function.client.WebClientResponseException
+import reactor.core.publisher.Mono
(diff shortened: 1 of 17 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClient.kt
```diff
@@ -0,0 +1,63 @@
...
+            .retrieve()
+            .bodyToMono(ElectoralRegistrationOfficesResponse::class.java)
+            .onErrorResume { ex -> handleException(ex, gssCode) }
+            .block()!!
+
+        if (response.eros.isEmpty()) {
+            throw ElectoralRegistrationOfficeNotFoundException(gssCode)
+        }
+
+        return eroMapper.toEroManagementApiEroDto(response.eros[0])
```
Neil-Massey, line 44: Should we raise an exception if size is not `1`?  (Or am I being too defensive?)

Neil-Massey (approved):
LGTM - I guess this is mostly a clone of what's been done elsewhere.  One question

src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClient.kt
adacayi: @Neil-Massey , I expected this would never happen. 
@NathanRussellValtech , do you think we should cover this case?

NathanRussellValtech (approved):
LGTM 👍

PR #16, opened 2022-10-25
EIP1-1364: EIP1-2134: Implement repository and entity classes for print details

adacayi:
This PR is to implement dynamo db configuration, repository and entity classes for persisting print details.

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/PrintDetails.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/PrintDetails.kt
@@ -1,0 +1,52 @@
+package uk.gov.dluhc.printapi.database.entity
+
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbBean
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbPartitionKey
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbSecondaryPartitionKey
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbSecondarySortKey
+import software.amazon.awssdk.enhanced.dynamodb.mapper.annotations.DynamoDbSortKey
+import java.time.LocalDate
+import java.time.OffsetDateTime
+import java.util.UUID
+
+@DynamoDbBean
+data class PrintDetails(
+    @get:DynamoDbPartitionKey var id: UUID? = null,
+    @get:DynamoDbSecondaryPartitionKey(indexNames = [REQUEST_ID_INDEX_NAME]) var requestId: String? = null,
+    var sourceReference: String? = null,
+    var applicationReference: String? = null,
+    @get:DynamoDbSecondaryPartitionKey(indexNames = [SOURCE_TYPE_GSS_CODE_INDEX_NAME]) var sourceType: SourceType? = null,
+    var vacNumber: String? = null,
+    var requestDateTime: OffsetDateTime? = null,
+    var firstName: String? = null,
+    var middleNames: String? = null,
+    var surname: String? = null,
+    var certificateLanguage: CertificateLanguage? = null,
+    var photoLocation: String? = null,
+    var delivery: CertificateDelivery? = null,
+    @get:DynamoDbSortKey @get:DynamoDbSecondarySortKey(indexNames = [SOURCE_TYPE_GSS_CODE_INDEX_NAME]) var gssCode: String? = null,
+    var issuingAuthority: String? = null,
+    var issueDate: LocalDate = LocalDate.now(),
+    var eroEnglish: ElectoralRegistrationOffice? = null,
+    var eroWelsh: ElectoralRegistrationOffice? = null
+) {
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other == null || this::class != other::class) return false
+        other as PrintDetails
+
+        return id == other.id
+    }
+
+    override fun hashCode(): Int = id.hashCode()
+
+    @Override
+    override fun toString(): String {
+        return this::class.simpleName + "(id = $id)"
+    }
+
+    companion object {
+        const val REQUEST_ID_INDEX_NAME = "RequestIdIndex"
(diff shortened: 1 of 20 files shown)


src/test/kotlin/uk/gov/dluhc/printapi/config/LocalStackContainerConfiguration.kt
```diff
@@ -36,10 +50,10 @@ class LocalStackContainerConfiguration {
         fun getInstance(): GenericContainer<*> {
             if (container == null) {
                 container = GenericContainer(
-                    DockerImageName.parse("localstack/localstack:1.1.0")
+                    DockerImageName.parse("localstack/localstack:1.2.0")
```
adacayi, line 53: I needed to update the version to 1.2.0, to be able to see the actual persisted test data by AWS NoSql Workbench. 

There is a known bug for the 1.1.0 version, which prevents NoSql Workbench to show table contents: 
https://github.com/localstack/localstack/issues/5737

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/PrintDetails.kt
```diff
@@ -0,0 +1,52 @@
...
+    var surname: String? = null,
+    var certificateLanguage: CertificateLanguage? = null,
+    var photoLocation: String? = null,
+    var delivery: CertificateDelivery? = null,
+    @get:DynamoDbSortKey @get:DynamoDbSecondarySortKey(indexNames = [SOURCE_TYPE_GSS_CODE_INDEX_NAME]) var gssCode: String? = null,
+    var issuingAuthority: String? = null,
+    var issueDate: LocalDate = LocalDate.now(),
+    var eroEnglish: ElectoralRegistrationOffice? = null,
+    var eroWelsh: ElectoralRegistrationOffice? = null
+) {
```
Neil-Massey, line 32: Is there no status included in the print details?  
Is there no batchId included in the print details?

If there are, will we need an index to allow finding the print details for a particular status and batch id?

adacayi: @Neil-Massey , I have pushed the changes accordingly.

I added a Global Secondary Index for status being the partition key and batchId being the sortKey.

For the index, @Neil-Massey has one question:
> I wonder if that GSI can be a concatenation of say GSS_CODE#STATUS#BATCH_ID

What do you think, @NathanRussellValtech?

By the way, since we only have partition key and sort key for a global secondary index, seems like we need to do a concatenation if we also include gssCode, and I am not sure, if this is something straightforward to implement.

src/test/kotlin/uk/gov/dluhc/printapi/config/LocalStackContainerConfiguration.kt
```diff
@@ -117,4 +132,86 @@ class LocalStackContainerConfiguration {
...
+            .endpointOverride(localStackContainer.getEndpointOverride())
+            .build()
+
+        createPrintDetailsTable(dynamoDbClient, dbConfiguration.printDetailsTableName)
+        return dynamoDbClient
+    }
+
+    private fun createPrintDetailsTable(dynamoDbClient: DynamoDbClient, tableName: String) {
+        if (dynamoDbClient.listTables().tableNames().contains(tableName)) {
+            dynamoDbClient.deleteTable { it.tableName(tableName) }
```
adacayi, line 153: Since we're reusing containers, I added this to make sure we have the latest version of the table.

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/PrintDetails.kt
Neil-Massey: As agreed the Print Batches are not partitioned by GSS Code (ERO) and a batch that is sent to the printer can include a mix of applications across different GSS Codes.  Therefore there is no need to include the GSS Code in the index. 👍

src/test/kotlin/uk/gov/dluhc/printapi/config/LocalStackContainerConfiguration.kt
Neil-Massey: Guess during development this is only ran once and helps development 👍 (I guess with MySQL we don't do anything similar - although Liquibase will complain if an applied changeset is modified)

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/PrintDetailsRepository.kt
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    //
+    // fun getBySourceReference(sourceReference: String, gssCode: String): List<Notification> {
+    //     val queryConditional = QueryConditional.keyEqualTo(key(sourceReference, gssCode))
+    //     val index = table.index(SOURCE_REFERENCE_INDEX_NAME)
+    //     val query = QueryEnhancedRequest.builder().queryConditional(queryConditional).build()
+    //     return index.query(query).flatMap { it.items() }
+    // }
+    //
```
Neil-Massey, line 39: Still needed?

adacayi: Thanks, @Neil-Massey , for spotting it. 
My bad. I removed it.

src/test/kotlin/uk/gov/dluhc/printapi/config/IntegrationTest.kt
```diff
@@ -38,7 +52,32 @@ internal abstract class IntegrationTest {
...
+    @BeforeEach
+    fun clearDatabase() {
+        clearTable(dynamoDbConfiguration.printDetailsTableName)
+    }
+
     companion object {
         val localStackContainer = LocalStackContainerConfiguration.getInstance()
     }
+
+    protected fun clearTable(tableName: String, partitionKey: String = "id", sortKey: String? = null) {
```
Neil-Massey, line 64: Originally I thought this might add a bit too much to the processing, but on reflection the Dynamo tables are going to be largely empty so that shouldn't be a concern for the Integration Tests.  At some point should port this over to the notifications api 👍

Neil-Massey (approved):
LGTM

PR #17, opened 2022-10-26
EIP-1364: EIP2132: Implement PrintDetailsMapper

adacayi:
This PR implements the mapper to map SendApplicationToPrintMessage, ero details and localAuthority to PrintDetails.

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/mapper/PrintDetailsMapper.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/mapper/PrintDetailsMapper.kt
@@ -1,0 +1,30 @@
+package uk.gov.dluhc.printapi.mapper
+
+import org.apache.commons.lang3.RandomStringUtils
+import org.bson.types.ObjectId
+import org.mapstruct.Mapper
+import org.mapstruct.Mapping
+import uk.gov.dluhc.printapi.database.entity.PrintDetails
+import uk.gov.dluhc.printapi.dto.EroManagementApiEroDto
+import uk.gov.dluhc.printapi.messaging.models.CertificateLanguage
+import uk.gov.dluhc.printapi.messaging.models.SendApplicationToPrintMessage
+import java.util.UUID
+
+@Mapper(uses = [SourceTypeMapper::class], imports = [UUID::class, ObjectId::class, RandomStringUtils::class])
+abstract class PrintDetailsMapper {
+    @Mapping(target = "id", expression = "java(UUID.randomUUID())")
+    @Mapping(target = "requestId", expression = "java( new ObjectId().toHexString())")
+    @Mapping(target = "vacNumber", expression = "java( RandomStringUtils.randomAlphanumeric(20))")
+    @Mapping(source = "ero", target = "eroEnglish")
+    @Mapping(source = "ero", target = "eroWelsh", conditionExpression = "java(isWelsh(message))")
+    @Mapping(source = "localAuthority", target = "issuingAuthority")
+    abstract fun toPrintDetails(
+        message: SendApplicationToPrintMessage,
+        ero: EroManagementApiEroDto,
+        localAuthority: String
+    ): PrintDetails
+
+    fun isWelsh(message: SendApplicationToPrintMessage): Boolean {
+        return message.certificateLanguage == CertificateLanguage.CY
+    }
+}
--- a/src/main/kotlin/uk/gov/dluhc/printapi/mapper/PrintDetailsMapper.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/mapper/PrintDetailsMapper.kt
@@ -4,19 +4,27 @@ import org.apache.commons.lang3.RandomStringUtils
 import org.bson.types.ObjectId
 import org.mapstruct.Mapper
 import org.mapstruct.Mapping
+import org.springframework.beans.factory.annotation.Autowired
 import uk.gov.dluhc.printapi.database.entity.PrintDetails
 import uk.gov.dluhc.printapi.dto.EroManagementApiEroDto
 import uk.gov.dluhc.printapi.messaging.models.CertificateLanguage
 import uk.gov.dluhc.printapi.messaging.models.SendApplicationToPrintMessage
+import uk.gov.dluhc.printapi.service.IdFactory
 import java.util.UUID
 
-@Mapper(uses = [SourceTypeMapper::class], imports = [UUID::class, ObjectId::class, RandomStringUtils::class])
+@Mapper(
+    uses = [SourceTypeMapper::class],
+    imports = [UUID::class, ObjectId::class, RandomStringUtils::class]
+)
 abstract class PrintDetailsMapper {
-    @Mapping(target = "id", expression = "java(UUID.randomUUID())")
-    @Mapping(target = "requestId", expression = "java( new ObjectId().toHexString())")
-    @Mapping(target = "vacNumber", expression = "java( RandomStringUtils.randomAlphanumeric(20))")
+    @Autowired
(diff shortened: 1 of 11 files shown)


Neil-Massey (approved):
LGTM

src/main/kotlin/uk/gov/dluhc/printapi/mapper/PrintDetailsMapper.kt
```diff
@@ -0,0 +1,30 @@
...
+    @Mapping(source = "ero", target = "eroEnglish")
+    @Mapping(source = "ero", target = "eroWelsh", conditionExpression = "java(isWelsh(message))")
+    @Mapping(source = "localAuthority", target = "issuingAuthority")
+    abstract fun toPrintDetails(
+        message: SendApplicationToPrintMessage,
+        ero: EroManagementApiEroDto,
+        localAuthority: String
+    ): PrintDetails
+
+    fun isWelsh(message: SendApplicationToPrintMessage): Boolean {
```
vishalgupta-valtech, line 27: Could be protected

vishalgupta-valtech: Line 28 condition could go in line 19 as expression ?not a blocker)

vishalgupta-valtech (approved):

src/main/kotlin/uk/gov/dluhc/printapi/mapper/PrintDetailsMapper.kt
```diff
@@ -0,0 +1,30 @@
...
+import uk.gov.dluhc.printapi.database.entity.PrintDetails
+import uk.gov.dluhc.printapi.dto.EroManagementApiEroDto
+import uk.gov.dluhc.printapi.messaging.models.CertificateLanguage
+import uk.gov.dluhc.printapi.messaging.models.SendApplicationToPrintMessage
+import java.util.UUID
+
+@Mapper(uses = [SourceTypeMapper::class], imports = [UUID::class, ObjectId::class, RandomStringUtils::class])
+abstract class PrintDetailsMapper {
+    @Mapping(target = "id", expression = "java(UUID.randomUUID())")
+    @Mapping(target = "requestId", expression = "java( new ObjectId().toHexString())")
```
NathanRussellValtech, line 16: For these 2 (`requestId` and `vacNumber`), are we not going to use the `IdFactory` we have?
We will 100% need to do that once we know the final format for the VAC Number - it's not going to be a simple algorithm and will likely need the gssCode, date, and (possibly?) previous VAC number if there was one (but ignore that last part for now)

Can you have a look at using the `IdFactory` please?

src/main/kotlin/uk/gov/dluhc/printapi/mapper/PrintDetailsMapper.kt
NathanRussellValtech: I think I prefer it how @adacayi currently has it - I worry that putting expression logic in the `@Mapping` annotation kinda hides it a little. I think a well named method like we have here makes it a lot clearer as to the intent etc ¯\_(ツ)_/¯

src/test/kotlin/uk/gov/dluhc/printapi/mapper/PrintDetailsMapperTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        // Given
+        val ero = buildEroManagementApiEroDto()
+        val localAuthority = ero.localAuthorities[0]
+        val message = buildSendApplicationToPrintMessage(certificateLanguage = certificateLanguageModel)
+        given(sourceTypeMapper.toSourceTypeEntity(any())).willReturn(SourceTypeEntity.VOTER_CARD)
+
+        val expected = with(message) {
+            PrintDetails(
+                id = UUID.randomUUID(),
+                requestId = getAMongoDbId(),
```
NathanRussellValtech, line 54: I'd like to abstract us away from the underlying technology if we could please?
The only reason we are using a mongo ID is that its convenient and suits the business need. As you know I'd rather we weren't using the mongo library at all, because we're not using mongo! ¯\_(ツ)_/¯ 

Presumably the method `getAMongoId` in a test data builder somewhere? Could you rename it please to something like `getARequestId` - because from a business perspective thats what its doing.
Thanks 🙏

src/test/kotlin/uk/gov/dluhc/printapi/mapper/PrintDetailsMapperTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        }
+        // When
+
+        val actual = mapper.toPrintDetails(message, ero, localAuthority.name)
+
+        // Then
+        assertThat(actual).usingRecursiveComparison().ignoringFields("id", "requestId", "vacNumber").isEqualTo(expected)
+        assertThat(actual.id).isNotNull
+        assertThat(actual.requestId).isNotNull.containsPattern(Regex("^[a-f\\d]{24}$").toPattern())
+        assertThat(actual.vacNumber).isNotNull.containsPattern(Regex("^[A-Za-z\\d]{20}$").toPattern())
```
NathanRussellValtech, line 113: This assertion is fine for now, but _may_ have to change in the near future (just FYI)
We have a proposed VAC format (still waiting to get agreement on it) where we will use a 5 bit character set (to give us a total of 32 different chars in the char set), which in practice will mean 0-9 and A-Z, but excluding B, I, O and S (because they look like numbers)

src/test/kotlin/uk/gov/dluhc/printapi/mapper/PrintDetailsMapperTest.kt
```diff
@@ -0,0 +1,135 @@
...
+        verify(sourceTypeMapper).toSourceTypeEntity(SourceTypeModel.VOTER_MINUS_CARD)
+    }
+
+    @Test
+    fun `should map id requestId and vacNumber to random values`() {
```
NathanRussellValtech, line 118: Is this test testing functionality of the mapper of the `IdFactory` ?
As per my comment above I think these values should be coming from the `IdFactory`, and therefore this test is not required?
What do you think?

NathanRussellValtech (commented):
Looking good, but I've made a couple of comments around where we should be getting IDs and VAC numbers from in respect of the mapper.
I wonder if you could have a look at those and see what you think?

src/main/kotlin/uk/gov/dluhc/printapi/mapper/PrintDetailsMapper.kt
adacayi: Hi @NathanRussellValtech ,

Thanks, for the suggestion. 

When I was implementing the mapper, I actually forgot we had the `IdFactory` class. 
I remembered it after pushing the changes and wanted to have your thoughts on the above implementation.
I agree with you. It makes sense to use the factory class, if the generation algorithm will be more complicated in the future.

I will change the code accordingly.

Neil-Massey: Maybe having the expressions in separate referenced functions (and not use expression at all) so that the mapper is kept simple and then there can be specific tests for these which would simplify the test class.

src/test/kotlin/uk/gov/dluhc/printapi/mapper/PrintDetailsMapperTest.kt
adacayi: Thanks, for pointing this @NathanRussellValtech. 
I have removed this test and just assert the `IdFactory` is used generating requestId and vacNumber in the test above it.

PR #18, opened 2022-10-26
EIP1-2214: Implement job scheduler

fayimora:
* Schedule locks is temporary disabled

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,8 @@ The following environment variables must be set in order to run the application:
 * `API_ERO_MANAGEMENT_URL` - the base URL of the ERO Management REST API service.
 * `DYNAMODB_ENDPOINT` - the localstack endpoint
 * `DYNAMODB_PRINT_DETAILS_TABLE_NAME` - table name to persist print details
+* `DYNAMODB_SCHEDULER_LOCKS_TABLE_NAME` - table name to persist scheduler locks
+* `JOBS_PRINT_REQUEST_CRON` - cron expression for print request job
 
 ### Authentication and authorisation
 Requests are authenticated by the presence of a signed cognito JWT as a bearer token in the HTTP request `authorization` header.  
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ The following environment variables must be set in order to run the application:
 * `DYNAMODB_ENDPOINT` - the localstack endpoint
 * `DYNAMODB_PRINT_DETAILS_TABLE_NAME` - table name to persist print details
 * `DYNAMODB_SCHEDULER_LOCKS_TABLE_NAME` - table name to persist scheduler locks
-* `JOBS_PRINT_REQUEST_CRON` - cron expression for print request job
+* `JOB_PRINT_REQUESTS_CRON` - cron expression for print request job
 
 ### Authentication and authorisation
 Requests are authenticated by the presence of a signed cognito JWT as a bearer token in the HTTP request `authorization` header.  
--- a/src/main/kotlin/uk/gov/dluhc/printapi/config/SchedulerLockConfiguration.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/config/SchedulerLockConfiguration.kt
@@ -1,0 +1,22 @@
+package uk.gov.dluhc.printapi.config
+
+import net.javacrumbs.shedlock.core.LockProvider
+import net.javacrumbs.shedlock.provider.dynamodb2.DynamoDBLockProvider
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.scheduling.annotation.EnableScheduling
+import software.amazon.awssdk.services.dynamodb.DynamoDbClient
+
+@Configuration
+@EnableScheduling
+// @EnableSchedulerLock(defaultLockAtMostFor = "PT5M")
+class SchedulerLockConfiguration(
+    private val dynamoDbClient: DynamoDbClient,
+    private val dynamoDbConfiguration: DynamoDbConfiguration
+) {
+
+    @Bean
+    fun schedulerLockProvider(): LockProvider {
+        return DynamoDBLockProvider(dynamoDbClient, dynamoDbConfiguration.schedulerLockTableName)
+    }
+}
--- a/src/main/kotlin/uk/gov/dluhc/printapi/config/SchedulerLockConfiguration.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/config/SchedulerLockConfiguration.kt
@@ -2,6 +2,7 @@ package uk.gov.dluhc.printapi.config
 
 import net.javacrumbs.shedlock.core.LockProvider
 import net.javacrumbs.shedlock.provider.dynamodb2.DynamoDBLockProvider
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty
 import org.springframework.context.annotation.Bean
 import org.springframework.context.annotation.Configuration
 import org.springframework.scheduling.annotation.EnableScheduling
@@ -9,6 +10,7 @@ import software.amazon.awssdk.services.dynamodb.DynamoDbClient
 
 @Configuration
 @EnableScheduling
+@ConditionalOnProperty("jobs.enabled", havingValue = "true")
 // @EnableSchedulerLock(defaultLockAtMostFor = "PT5M")
 class SchedulerLockConfiguration(
     private val dynamoDbClient: DynamoDbClient,
--- a/src/main/kotlin/uk/gov/dluhc/printapi/jobs/BatchPrintRequestsJob.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/jobs/BatchPrintRequestsJob.kt
@@ -1,0 +1,17 @@
+package uk.gov.dluhc.printapi.jobs
+
+import mu.KotlinLogging
+import org.springframework.scheduling.annotation.Scheduled
+import org.springframework.stereotype.Component
+
+private val logger = KotlinLogging.logger {}
+
+@Component
+class BatchPrintRequestsJob {
+
+    @Scheduled(cron = "\${jobs.print-requests.cron}")
+//    @SchedulerLock(name = "\${jobs.print-requests.name}", lockAtMostFor = "\${jobs.print-requests.lockAtMostFor}", lockAtLeastFor = "\${jobs.print-requests.lockAtLeastFor}")
+    fun run() {
+        logger.info { "Hello, I'm your batching agent" }
+    }
+}
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -12,6 +12,6 @@ dynamodb:
 jobs:
   print-requests:
     name: "BatchPrintRequests"
-    cron: ${JOBS_PRINT_REQUEST_CRON}
+    cron: ${JOB_PRINT_REQUESTS_CRON}
 #    lock-at-most-for: ${JOBS_PRINT_REQUEST_LOCK_AT_MOST_FOR}
 #    lock-at-least-for: ${JOBS_PRINT_REQUEST_LOCK_AT_LEAST_FOR}
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -7,3 +7,11 @@ api:
 
 dynamodb:
   print-details-table-name: ${DYNAMODB_PRINT_DETAILS_TABLE_NAME}
+  scheduler-lock-table-name: ${DYNAMODB_SCHEDULER_LOCKS_TABLE_NAME}
+
+jobs:
+  print-requests:
+    name: "BatchPrintRequests"
+    cron: ${JOBS_PRINT_REQUEST_CRON}
+#    lock-at-most-for: ${JOBS_PRINT_REQUEST_LOCK_AT_MOST_FOR}
+#    lock-at-least-for: ${JOBS_PRINT_REQUEST_LOCK_AT_LEAST_FOR}
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -10,8 +10,10 @@ dynamodb:
   scheduler-lock-table-name: ${DYNAMODB_SCHEDULER_LOCKS_TABLE_NAME}
 
 jobs:
+  enabled: true
   print-requests:
     name: "BatchPrintRequests"
-    cron: ${JOB_PRINT_REQUESTS_CRON}
+    cron: "${JOB_PRINT_REQUESTS_CRON:0 0 */1 * * *}" # every hour by default
 #    lock-at-most-for: ${JOBS_PRINT_REQUEST_LOCK_AT_MOST_FOR}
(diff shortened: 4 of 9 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/config/SchedulerLockConfiguration.kt
```diff
@@ -0,0 +1,22 @@
...
+import net.javacrumbs.shedlock.core.LockProvider
+import net.javacrumbs.shedlock.provider.dynamodb2.DynamoDBLockProvider
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.scheduling.annotation.EnableScheduling
+import software.amazon.awssdk.services.dynamodb.DynamoDbClient
+
+@Configuration
+@EnableScheduling
+// @EnableSchedulerLock(defaultLockAtMostFor = "PT5M")
```
NathanRussellValtech, line 12: Did you mean to leave this is?
Do we need this annotation to define lock behaviour, such as hold long the lock is held for etc? If we need (should?) configure things like that, should the values be properties in `application.yml` or similar?

src/main/kotlin/uk/gov/dluhc/printapi/jobs/BatchPrintRequestsJob.kt
```diff
@@ -0,0 +1,17 @@
...
+import org.springframework.scheduling.annotation.Scheduled
+import org.springframework.stereotype.Component
+
+private val logger = KotlinLogging.logger {}
+
+@Component
+class BatchPrintRequestsJob {
+
+    @Scheduled(cron = "\${jobs.print-requests.cron}")
+//    @SchedulerLock(name = "\${jobs.print-requests.name}", lockAtMostFor = "\${jobs.print-requests.lockAtMostFor}", lockAtLeastFor = "\${jobs.print-requests.lockAtLeastFor}")
```
NathanRussellValtech, line 13: Should this line not be commented?

src/main/kotlin/uk/gov/dluhc/printapi/jobs/BatchPrintRequestsJob.kt
```diff
@@ -0,0 +1,17 @@
...
+    fun run() {
+        logger.info { "Hello, I'm your batching agent" }
```
NathanRussellValtech, line 15: 😁

src/main/resources/application.yml
```diff
@@ -6,4 +6,12 @@ api:
...
-  print-details-table-name: ${DYNAMODB_PRINT_DETAILS_TABLE_NAME}
\ No newline at end of file
+  print-details-table-name: ${DYNAMODB_PRINT_DETAILS_TABLE_NAME}
+  scheduler-lock-table-name: ${DYNAMODB_SCHEDULER_LOCKS_TABLE_NAME}
+
+jobs:
+  print-requests:
+    name: "BatchPrintRequests"
+    cron: ${JOBS_PRINT_REQUEST_CRON}
+#    lock-at-most-for: ${JOBS_PRINT_REQUEST_LOCK_AT_MOST_FOR}
```
NathanRussellValtech, line 16: Commented out?

src/test/kotlin/uk/gov/dluhc/printapi/config/LocalStackContainerConfiguration.kt
```diff
@@ -214,4 +216,17 @@ class LocalStackContainerConfiguration {
         val mappedPort = getMappedPort(DEFAULT_PORT)
         return URI("http://$ipAddress:$mappedPort")
     }
+
+    protected fun createSchedulerLockTable(dynamoDbClient: DynamoDbClient, tableName: String) {
```
NathanRussellValtech, line 220: Minor - `protected`? Could it be `private` ?

src/test/resources/application-test.yml
```diff
@@ -15,4 +15,10 @@ api:
...
 dynamodb:
-  print-details-table-name: print-details
\ No newline at end of file
+  print-details-table-name: print-details
+  scheduler-lock-table-name: scheduler-locks
+
+jobs:
+    print-requests:
+        name: BatchPrintRequests
+        cron: 0 0 */1 * * *
```
NathanRussellValtech, line 24: Could you put a comment after/above the cron expression explaining what it evaluates to please?
Also (because I didn't know what the cron evaluates to) I had to look it up - I went to https://crontab.guru (my usual goto 😁 ) and it wasn't happy with the cron because it has 6 elements. Do we need a 5 or 6 element cron expression here? What format is Spring happy with? 

https://www.baeldung.com/cron-expressions (gives examples of 5 elements)
https://www.baeldung.com/cron-syntax-linux-vs-spring (describes how spring support a 6th element for some reason that I didnt read properly 🤣 )
https://spring.io/blog/2020/11/10/new-in-spring-5-3-improved-cron-expressions (suggests 6 elements?)

NathanRussellValtech, line 24: The cron you have put here is "at midnight, every day of the month" (according to crontab.guru)
What does this mean for our integration tests in general?

Your `@Scheduled` class at the moment does nothing, which is fine. I guess all you are doing here is providing a value so that Spring is happy and starts the context.

What happens when the `@Scheduled` class has some behaviour in it? Could this potentially get triggered when we didn't expect it to? (eg: maybe someone is working really late over midnight? (it has been known! 🙄 ). Or more likely, what happens if our CI stack is running the tests over a midnight boundary?) Might this be a problem somehow?

But on the subject of integration tests generally (and I know it wasn't in your spike, so thats fine), have you had any thoughts as to how we might test `@Scheduled` class once we have some behaviour in it?
Clearly we will write the class in such a way that we can unit test it with mocks etc. But will we need to integration test it? Will we need to test it in such a way that the cron is triggered, and it queries the database, and produces a batch, and uploads it to the SFTP server etc etc ?
That might be really really really hard to do 🤔 

I honestly dont know the answer to that, but I wonder if it might be a case that we (somehow) disable scheduled jobs altogether (all of them) in our integration test context, so that they dont 'randomly' trigger etc. And that we just accept that we cant integration test this part of the system in this way?

What do you think? Would definitely be worth getting other views and opinions on this from the rest of the API dev team 🤔

src/main/resources/application.yml
NathanRussellValtech: If/when you add these back in, I wonder if these are more suited to being values in here, rather than referenced as environment variables. EG:
```
jobs:
  print-requests:
    name: "BatchPrintRequests"
    cron: ${JOBS_PRINT_REQUEST_CRON}
    lock-at-most-for: PT5M
```
The thinking is, will this value (the 'lock at most for' for the print-request job) ever be different across environments? Will we have a different lock time for `dev` vs `int` vs `prod` for example?

If the answer is no/unlikely then we can do it without an environment variable, which makes infra etc a bit simpler.

(I would also suggest the same argument _might_ be true of the cron expression itself - will we want this triggering at different cron schedules in different environments? - can we just put the cron expression here and be done with it? Which would mean you dont need it in `application-test.yml` ?)

NathanRussellValtech:
> * Schedule locks is temporary disabled

Why is that? I guess it ties in with my PR comments about commented out code. But why are they temporarily disabled? Were they causing you problems in some way?

README.md
```diff
@@ -33,6 +33,8 @@ The following environment variables must be set in order to run the application:
 * `API_ERO_MANAGEMENT_URL` - the base URL of the ERO Management REST API service.
 * `DYNAMODB_ENDPOINT` - the localstack endpoint
 * `DYNAMODB_PRINT_DETAILS_TABLE_NAME` - table name to persist print details
+* `DYNAMODB_SCHEDULER_LOCKS_TABLE_NAME` - table name to persist scheduler locks
+* `JOBS_PRINT_REQUEST_CRON` - cron expression for print request job
```
Neil-Massey, line 37: Minor - probably `JOB`

src/main/resources/application.yml
Neil-Massey: I guess the lock expiry and the CRON are linked - if the lock expiry is longer than the CRON frequency then the CRON does not have the desired effect.  But I would suggest that is unlikely to happen in practice - I would guess the CRON frequency will not be less than 5 minutes and the expiry will not be more than 5 minutes.

For these I would assume setting sensible defaults in the application.yml will be sufficient - If required environment properties can always override (even if not defined explicitly here)

NathanRussellValtech: Yes, agreed. I would imagine `dev` and `int` would be the only envs where we _might_ want something different - in examples like that perhaps the best approach would be to override the relevant properties in `application-dev.yml` etc
(we already do something similar in VCA to control feature flags in dev where we will never want a feature on)

src/test/resources/application-test.yml
```diff
@@ -15,4 +15,10 @@ api:
     url: http://will-be-replaced-by-wireMockServer-bean
 
 dynamodb:
-  print-details-table-name: print-details
\ No newline at end of file
+  print-details-table-name: print-details
+  scheduler-lock-table-name: scheduler-locks
+
+jobs:
```
Neil-Massey, line 21: Possibly for Integration testing, it may be better to call the method that the scheduler calls directly and to disable scheduled tasks in the test context (it might make testing easier, more predictable and faster than relying on a CRON to fire).

[An approach to disable scheduled tasks](https://stackoverflow.com/questions/61977922/is-there-a-global-flag-in-spring-boot-to-disable-all-the-scheduled-jobs):
```
@Configuration
@EnableScheduling
@ConditionalOnProperty(name="jobs.enabled", havingValue = "true")
public class ConditionalEnableScheduling {
}
```

src/main/kotlin/uk/gov/dluhc/printapi/config/SchedulerLockConfiguration.kt
fayimora: Yes, this is intentionally commented out because the DB needs to be set up first else no one will be able to run the app. I'll bring it back in a follow-up PR

src/main/kotlin/uk/gov/dluhc/printapi/jobs/BatchPrintRequestsJob.kt
fayimora: Same as above

src/test/kotlin/uk/gov/dluhc/printapi/config/LocalStackContainerConfiguration.kt
fayimora: Yeah it can be

src/test/resources/application-test.yml
fayimora: > Do we need a 5 or 6 element cron expression here? What format is Spring happy with?

Spring uses 6 elements, not the UNIX 5 elements

README.md
fayimora: Does `JOB_PRINT_REQUESTS_CRON ` work better?

src/test/resources/application-test.yml
fayimora: Yes, I totally agree with this. Said a similar thing to Nathan above

README.md
Neil-Massey: Yes, sorry, I was only referring to prefix (JOB v JOBS) - definitely use `JOB_PRINT_REQUESTS_CRON` 👍

src/main/resources/application.yml
```diff
@@ -6,4 +6,12 @@ api:
...
 dynamodb:
-  print-details-table-name: ${DYNAMODB_PRINT_DETAILS_TABLE_NAME}
\ No newline at end of file
+  print-details-table-name: ${DYNAMODB_PRINT_DETAILS_TABLE_NAME}
+  scheduler-lock-table-name: ${DYNAMODB_SCHEDULER_LOCKS_TABLE_NAME}
+
+jobs:
+  print-requests:
+    name: "BatchPrintRequests"
+    cron: ${JOB_PRINT_REQUESTS_CRON}
```
NathanRussellValtech, line 15: As per the earlier comments, did we agree that this property (`cron`) would be a value in this yml, rather than a property from the environment? eg:
```
jobs:
  print-requests:
    name: "BatchPrintRequests"
    cron: 0 0 */1 * * *
```
And then you can do away with the environment variable altogether (which now doesnt match the env name in the readme ?)
My preference here would be to simply put the cron here, then get rid of the env var from the readme 👍

src/test/resources/application-test.yml
NathanRussellValtech, line 21: These properties (jobs.print-requests ...) are not required here now if the are not overriding the ones in the main application yml.
But I think I'd like to see the conditional scheduling config and bean that @Neil-Massey has suggested in this PR please?

src/main/resources/application.yml
fayimora: @NathanRussellValtech  I have it like this: 
```
cron: "${JOB_PRINT_REQUESTS_CRON:0 0 */1 * * *}" # every hour by default
```

I think it's still worth having a variable to override in the env but with a default value. Do you want me to get rid of the env var completely?

src/test/resources/application-test.yml
fayimora: Ok I'll look to add that to this PR then

fayimora: Added

src/main/resources/application.yml
```diff
@@ -6,4 +6,14 @@ api:
...
-  print-details-table-name: ${DYNAMODB_PRINT_DETAILS_TABLE_NAME}
\ No newline at end of file
+  print-details-table-name: ${DYNAMODB_PRINT_DETAILS_TABLE_NAME}
+  scheduler-lock-table-name: ${DYNAMODB_SCHEDULER_LOCKS_TABLE_NAME}
+
+jobs:
+  enabled: true
+  print-requests:
+    name: "BatchPrintRequests"
+    cron: "${JOB_PRINT_REQUESTS_CRON:0 0 */1 * * *}" # every hour by default
```
NathanRussellValtech, line 16: Looking good now. The only thing I would do (but you can do it in your next PR if you like) is to rid of this env var `JOB_PRINT_REQUESTS_CRON` altogether, both from here, and also in the readme.

Generally speaking we document env vars in the readme that are must haves - the app wont start/operate correctly without them.

In this case its an override, and we can do it without specifying the env var:
```
    cron: "0 0 */1 * * *" # every hour by default
```
If we wanted to override it we have 2 options - we can either commit an application yml for the environment with just the override. EG:
```application-prod.yml
jobs:
  print-requests:
    cron: "0 0 */2 * * *}"
```
or we can set an environment variable that corresponds to the property path:
```
JOBS_PRINT_REQUESTS_CRON="0 0 */3 * * *"
```
Spring will do magic and map the env var to the correct property. Magic 😁 

So, as mentioned, generally speaking we only use env vars when we need to set the value per environment, and it is only those env vars that we document in the readme (the essential ones)

NathanRussellValtech (approved):
LGTM, thanks for making the suggested changes.
If you could think about removing the env var as per my last comment (do it in your next PR) that would be good as well 👍

PR #20, opened 2022-10-27
EIP1-1364: EIP1-2135: Persist send application to print message

adacayi:
This pr is to implement the message listener functionality to save the received message to print-details table.

### DIFF
--- a/src/test/kotlin/uk/gov/dluhc/printapi/messaging/SendApplicationToPrintMessageListenerIntegrationTest.kt
+++ b/src/test/kotlin/uk/gov/dluhc/printapi/messaging/SendApplicationToPrintMessageListenerIntegrationTest.kt
@@ -4,9 +4,21 @@ import ch.qos.logback.classic.Level
 import org.assertj.core.api.Assertions.assertThat
 import org.awaitility.kotlin.await
 import org.junit.jupiter.api.Test
+import software.amazon.awssdk.services.dynamodb.model.ScanRequest
 import uk.gov.dluhc.printapi.config.IntegrationTest
+import uk.gov.dluhc.printapi.database.entity.Address
+import uk.gov.dluhc.printapi.database.entity.CertificateDelivery
+import uk.gov.dluhc.printapi.database.entity.CertificateLanguage
+import uk.gov.dluhc.printapi.database.entity.ElectoralRegistrationOffice
+import uk.gov.dluhc.printapi.database.entity.PrintDetails
+import uk.gov.dluhc.printapi.database.entity.SourceType
 import uk.gov.dluhc.printapi.testsupport.TestLogAppender
+import uk.gov.dluhc.printapi.testsupport.testdata.aValidRequestId
+import uk.gov.dluhc.printapi.testsupport.testdata.aValidVacNumber
+import uk.gov.dluhc.printapi.testsupport.testdata.model.buildElectoralRegistrationOfficeResponse
 import uk.gov.dluhc.printapi.testsupport.testdata.model.buildSendApplicationToPrintMessage
+import java.time.LocalDate
+import java.util.UUID
 import java.util.concurrent.TimeUnit.SECONDS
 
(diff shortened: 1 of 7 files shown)


src/test/kotlin/uk/gov/dluhc/printapi/messaging/SendApplicationToPrintMessageListenerIntegrationTest.kt
```diff
@@ -4,28 +4,97 @@ import ch.qos.logback.classic.Level
...
 
         // Then
         await.atMost(5, SECONDS).untilAsserted {
-            assertThat(
-                TestLogAppender.hasLog(
-                    "Sending application [${payload.sourceReference}] to print", Level.INFO
-                )
-            ).isTrue()
+            wireMockService.verifyEroManagementGetEro(gssCode)
+            val response = dynamoDbClient.scan(
```
NathanRussellValtech, line 87: Is there any mileage here to use the repository rather than the dynamo client?
(I don't know what the repository class has in it yet wrt query/read methods - if there are no suitable methods in the repository yet then what you've done here is probably right. But if there are suitable methods would that make this test a little tidier?)

src/test/kotlin/uk/gov/dluhc/printapi/messaging/SendApplicationToPrintMessageListenerIntegrationTest.kt
```diff
@@ -41,7 +110,7 @@ internal class SendApplicationToPrintMessageListenerIntegrationTest : Integratio
         await.during(5, SECONDS).until {
             assertThat(
                 TestLogAppender.hasLog(
-                    "Sending application [${payload.sourceReference}] to print", Level.INFO
+                    "Print message with source reference [${payload.sourceReference}] received", Level.INFO
```
NathanRussellValtech, line 113: Any thoughts on changing this assertion to assert that a record has not been written to dynamo, rather than asserting that log messages have not been written?

(I'm not a fan on asserting business logic based on the presence or not of log messages - I know why we did it initially because the service and repo etc where not there, and thats fine. But the implementation now allows us to do a better assertion ?)

src/test/kotlin/uk/gov/dluhc/printapi/messaging/SendApplicationToPrintMessageListenerIntegrationTest.kt
adacayi: Hi, @NathanRussellValtech ,
The repository currently only contains getting items by the primary key, hence needed to use the dynamodb client.

NathanRussellValtech: I thought that might be the case; but thought I'd ask anyway 😁 
Thats cool 👍

NathanRussellValtech (approved):
LGTM, but if we can tidy the assertion to assert no record was written to the database rather than no log message written, that would be even better 👍

adacayi:
> LGTM, but if we can tidy the assertion to assert no record was written to the database rather than no log message written, that would be even better 👍

Thanks, @NathanRussellValtech ,
I have just pushed the changes for the assertion as well.

src/test/kotlin/uk/gov/dluhc/printapi/messaging/SendApplicationToPrintMessageListenerIntegrationTest.kt
adacayi: Thanks, for the comment, @NathanRussellValtech .
I have changed the assertion to assert no data has persisted.

NathanRussellValtech: Yeah, thats much better. Thank you 🙏

PR #21, opened 2022-10-27
Feature/eip1 1878 query print repo

Neil-Massey:
Split out small PR that queries the repo for print requests in suitable status and for particular batch


src/main/kotlin/uk/gov/dluhc/printapi/database/entity/PrintDetails.kt
```diff
@@ -16,20 +16,23 @@ data class PrintDetails(
...
     var middleNames: String? = null,
     var surname: String? = null,
     var certificateLanguage: CertificateLanguage? = null,
+    var certificateFormat: CertificateFormat? = CertificateFormat.STANDARD,
     var photoLocation: String? = null,
     var delivery: CertificateDelivery? = null,
     @get:DynamoDbSecondarySortKey(indexNames = [SOURCE_TYPE_GSS_CODE_INDEX_NAME]) var gssCode: String? = null,
     var issuingAuthority: String? = null,
     var issueDate: LocalDate = LocalDate.now(),
+    var suggestedExpiryDate: LocalDate = issueDate.plusYears(10),
```
NathanRussellValtech, line 31: I've not been close enough to the dynamo stuff (in both print-api and notifications-api), but I'm a bit confused over a couple of things ...

These dynamo entity (item ?) classes appear to be data classes, not regular classes that we used for JPA. Was that a conscious decision?

I'm also confused by all the nullable fields, even those with a default value. I assumed/guessed it worked the same way as JPA when retrieving records from the DB, in that JPA will new up the empty object with its no-args constructor, then will call the setters / use reflection to populate the fields.
So I assumed it worked the same way ....

But then I saw `issueDate`, `suggestedExpiryDate` and `status` are all non-nullable fields but with default values.
Was this intentional? Why are these fields different to the other fields, even those that also have a default value (eg: `vacVersion` and `certificateFormat`)

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/PrintDetailsRepository.kt
```diff
@@ -29,6 +33,15 @@ class PrintDetailsRepository(client: DynamoDbEnhancedClient, tableConfig: Dynamo
         }
     }
 
+    fun getAllByStatusAndBatchId(status: Status, batchId: String): List<PrintDetails> {
```
NathanRussellValtech, line 36: Oh my days! I have so much to learn about dynamo! 😬 
(that says more about me than your code here, which I'm sure is absolutely fine 😁 )

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/testdata/entity/PrintDetailsBuilder.kt
```diff
@@ -38,7 +38,7 @@ fun buildPrintDetails(
     photoLocation: String = "arn:aws:s3:::source-document-storage/$gssCode/$sourceReference/${UUID.randomUUID()}/" +
         faker.file().fileName("", null, "jpg", ""),
     status: Status = Status.PENDING_ASSIGNMENT_TO_BATCH,
-    batchId: UUID? = UUID.randomUUID()
+    batchId: String? = UUID.randomUUID().toString()
```
NathanRussellValtech, line 41: Minor - I know this only a test builder, but is it worth removing the dashes from the batchId so that it better reflects what will actually be set?
```
batchId: String? = UUID.randomUUID().toString().replace("-", "")
```

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/PrintDetails.kt
```diff
@@ -16,20 +16,23 @@ data class PrintDetails(
     var applicationReference: String? = null,
     @get:DynamoDbSecondaryPartitionKey(indexNames = [SOURCE_TYPE_GSS_CODE_INDEX_NAME]) var sourceType: SourceType? = null,
     var vacNumber: String? = null,
+    var vacVersion: String? = "1",
```
NathanRussellValtech, line 19: I'm not sure the logic for these defaults (`vacVersion`, `suggestedExpiry` , and possibly `certificateFormat`) live in the entity class, but they are fine here for now.

I could well imagine a scenario where the version of the certificate and the suggested expiry should be in `application.yml` so they can be "tuned"; and I can also imagine a scenario where we might want to print a certificate using template version that is not the default.
It feels like the service that is creating these entities via the repository should control these field values / know about the defaults .... but this is fine for now. We can (and probably will) refactor this as and when we have a story to set these values

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/PrintDetailsRepository.kt
Neil-Massey: Possibly will need to revisit this once we have a larger data set to verify against.
Working with Dynamo there is much closer coupling between the database indexes and how the query is stated - not like with MySQL where we just are for data, here we also say what index to use :-)

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/PrintDetails.kt
Neil-Massey: I don't really think we can compare DynamoDB with JPA - there are no proxies, no dirty detection, no lazy loading, no recommendation that classes need to be mutable.  [Tutorials I read at the time](https://betterprogramming.pub/aws-java-sdk-v2-dynamodb-enhanced-client-with-kotlin-spring-boot-application-f880c74193a2) all used mutable data classes.

I added the defaults here as we don't currently have clear requirements for these properties and assume later stories will enhance/update/refactor the code appropriately - I saw that `IssueDate` had set it's default in similar manner so went with that approach.

PR #25, opened 2022-10-28
EIP1-1878 Produce zip stream containing print requests

Neil-Massey:
PR to generate a zip stream containing print requests that are streamed to an SFTP destination.

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/config/ThreadPoolConfiguration.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/config/ThreadPoolConfiguration.kt
@@ -1,0 +1,22 @@
+package uk.gov.dluhc.printapi.config
+
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.context.annotation.Bean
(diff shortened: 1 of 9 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/config/ThreadPoolConfiguration.kt
```diff
@@ -0,0 +1,22 @@
...
+import org.springframework.context.annotation.Bean
+import org.springframework.context.annotation.Configuration
+import org.springframework.scheduling.concurrent.ThreadPoolTaskExecutor
+
+@Configuration
+class ThreadPoolConfiguration {
+
+    @Bean
+    fun zipStreamProducerTaskExecutor(
+        @Value("\${thread-pool.zip.core-size:5}") coreSize: Int,
```
NathanRussellValtech, line 13: Minor - can we put these defaults in `application.yml` ? Then at least all the property definitions and their values are in one place (saves hunting all over the code base in 6 months looking where we set a given property! :D )

NathanRussellValtech (approved):

PR #27, opened 2022-10-31
EIP1-2131 - Addition of CertificateNumber

NathanRussellValtech:
As discussed in the tech forum just now 😁 I present my finest nerdiness 👍 
This PR adds in a class called `CertificateNumber` that works in a very similar way to `ObjectId` (in that you new one up when you want a new certificate number). Internally it works the same way of holding the value in 12 bytes. The difference is how the 12 byte ID is serialized - `ObjectId` serializes as 24 hex chars. `CertificateNumber` serializes as 20 characters from its own alphabet

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/domain/CertificateNumber.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/domain/CertificateNumber.kt
@@ -1,6 +1,7 @@
 package uk.gov.dluhc.printapi.domain
 
 import java.security.SecureRandom
+import java.time.Clock
 import java.time.Instant
 import java.util.concurrent.atomic.AtomicInteger
 
@@ -68,7 +69,7 @@ class CertificateNumber {
     }
 
     constructor() : this(
-        timestamp = dateToTimestampSeconds(Instant.now()),
+        timestamp = dateToTimestampSeconds(Instant.now(Clock.systemUTC())),
         randomValue1 = RANDOM_VALUE1,
         randomValue2 = RANDOM_VALUE2,
         counter = NEXT_COUNTER.getAndIncrement(),
--- a/src/main/kotlin/uk/gov/dluhc/printapi/domain/CertificateNumber.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/domain/CertificateNumber.kt
@@ -28,9 +28,11 @@ import java.util.concurrent.atomic.AtomicInteger
  *  `increment` is an [AtomicInteger] and is a static member seeded from a random number. The use of `getAndIncrement()` means
  *  this is a thread safe way of generating an incremental counter.
  *
- * An instance of [CertificateNumber] is a 12 byte identifier which serializes via it's `toString()` method into a 20
- * character string using the character set `0123456789ACDEFGHJKLMNPQRTUVWXYZ` (specifically B, I, O and S are excluded
- * as they can be misread as numbers)
+ * An instance of [CertificateNumber] is a 12 byte identifier which serializes as a 20 character string using the character
+ * set `0123456789ACDEFGHJKLMNPQRTUVWXYZ` (specifically B, I, O and S are excluded as they can be misread as numbers)
+ *
+ * Class constructors are convenience constructors for the purpose of tests. The public API and preferred method of
+ * creating a new Certificate Number is to use the static factory method [CertificateNumber.create]
  */
 class CertificateNumber {
 
@@ -57,6 +59,12 @@ class CertificateNumber {
 
         private fun dateToTimestampSeconds(date: Instant): Int =
             date.epochSecond.toInt()
+
+        /**
+         * Creates a new 20 character string Certificate Number.
+         */
+        fun create(): String =
+            CertificateNumber().toString()
     }
 
     constructor() : this(
@@ -88,10 +96,7 @@ class CertificateNumber {
         this.counter = counter
     }
 
-    /**
-     * Convenience constructor used by tests to create CertificateNumber with known internal values.
-     */
-    internal constructor(timestamp: Int, randomValue1: Int, randomValue2: Short, counter: Int) {
+    constructor(timestamp: Int, randomValue1: Int, randomValue2: Short, counter: Int) {
         validateValues(randomValue1, counter)
         this.timestamp = timestamp
         this.randomValue1 = randomValue1
--- a/src/main/kotlin/uk/gov/dluhc/printapi/domain/CertificateNumber.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/domain/CertificateNumber.kt
@@ -1,0 +1,142 @@
+package uk.gov.dluhc.printapi.domain
+
+import java.security.SecureRandom
+import java.time.Instant
+import java.util.concurrent.atomic.AtomicInteger
+
+/**
+ * A globally unique identifier, heavily based on [mongodb's ObjectId class](https://github.com/mongodb/mongo-java-driver/blob/master/bson/src/main/org/bson/types/ObjectId.java)
+ *
+ * Consists of 12 bytes, divided as follows:
+ *
+ * ```
+ *   | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 |
+ *   |---------------|-----------|-------|-------------|
+ *   | time          |rnd 1      | rnd 2 | increment   |
+ * ```
+ *
+ * Where:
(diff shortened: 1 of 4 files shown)


matt-wills-valtech (approved):

src/main/kotlin/uk/gov/dluhc/printapi/domain/CertificateNumber.kt
```diff
@@ -0,0 +1,142 @@
...
+ *  will be the same.
+ *
+ *  `increment` is an [AtomicInteger] and is a static member seeded from a random number. The use of `getAndIncrement()` means
+ *  this is a thread safe way of generating an incremental counter.
+ *
+ * An instance of [CertificateNumber] is a 12 byte identifier which serializes via it's `toString()` method into a 20
+ * character string using the character set `0123456789ACDEFGHJKLMNPQRTUVWXYZ` (specifically B, I, O and S are excluded
+ * as they can be misread as numbers)
+ */
+class CertificateNumber {
```
matt-wills-valtech, line 35: Great until we have another use case for a 20 character value, but I can't think of an appropriate name! 😉

sentioservices (commented):
I'm concerned that we've opted for this implementation because it is good at reducing the probability of a clash of certificate numbers, 
at the expense of making other certificate numbers relatively easy to guess, especially those generated around the same time.
I'm not sure what the precise vulnerability is but it feels like the good practise of not making IDs easily guessable should be extended to Certificate Numbers.
To put the chances of a clash into perspective, when generating a 20 character digit string from a set of 32 characters the number of combinations is 1.2 x 10^30. This compares favourably with:
- the number of grains of sand in all the beaches on earth: 75 x 10^17
- the number of stars in the universe: 1 x 10^24
I presume that we'll be storing the certificate number in a DB, so to guarantee uniqueness we could additionally wrap the insert statement with a retry when that particular unique constraint is violated.

NathanRussellValtech:
Thanks for your comments @sentioservices. These are all valid comments, but I don't know what you are suggesting as alternatives?

Our primary requirement from the business is a 20 character unique reference.
Do we meet that? Yes ✅ 

Is there a requirement for the certificate number to be "non-guessable"?
No. Have a look at the [jira](https://technologyprogramme.atlassian.net/browse/EIP1-2131) where you will find quite a lot of analysis specifically in this area. There were originally "requirements" that certain data would be encoded in the certificate such as issue date and ERO/gssCode, to "provide authenticity". Digging into this in more detail with the business (Guy Daws) it because clear that these were just words, with no business requirement or process behind them. He agreed that nobody at the polling station would eyeball/check the certificate number in any way - an elector would arrive, present their certificate, the photo matches, they are allowed to vote. He also agreed that if there were such a process the algorithm would have to be simple enough such that we could communicate it in process manuals etc in simple terms, and that it would have to be simple enough that the polling station official could decode it / validate it "in their head" without the need for a calculator or looking up online.

The core challenge here was generating a unique reference, and that it was 20 characters; not that it was "secure" / non-guessable. Even if it were, what would stop a fraudster just guessing any 20 characters? Without a validation process (see above point) who is going to determine that its real or not?

If we had 24 characters available, we would almost certainly have just used mongo's `ObjectId` "as is". We already use that elsewhere for IDs/reference numbers. IER use them as their unique `applicationId` that we then adopt as our unique ID within the system. And we use it in the `print-api` as the `requestId`. OK, I accept these are slightly different use cases as they are not public facing. But my question is, if we could have used `ObjectId` for the main certificate number, would you have raised the same concerns? Are your concerns about the underlying algorithm of 12 bytes of data, or the way we are serializing that data into a 20 character string? 

Given your comments over the apparent sequential nature of the serialization, I'm guessing you are really questioning the underlying algorithm that Mongo use? (12 bytes where 4 bytes for the datetime which clearly increment; 5 bytes whilst randomly intialized are then fixed; and 3 bytes for the counter which clearly increments)
I don't know that we are in a position to question Mongo's algorithm for generating a unique ID? Maybe we are? But it seems pretty well industry accepted and battle tested to me?
The Mongo algorithm for generating a unique ID (via their 12 byte algorithm) is taken as read - I don't think we can question that. Is it non-guessable / non-sequental? No, but I don't think it was ever designed as such. Its a unique ID
Does their uniqueness fit our requirements? Yes. Do we need to be concerned over apparent sequential IDs / guess-ability ? No. We have no requirements for that.

OK, so that takes us onto your comments on the character set itself.
Our requirement was to generate a 20 character ID / reference number. On the basis that I wanted to re-use the greater wisdom of Mongo's internal algorithm, and not re-invent that particular wheel (which I would not be able to do as well ¯\_(ツ)_/¯ ), my starting point was a 12 byte ID. The approach I came up with was splitting those 12 bytes (96 bits) into 5 bit chunks. 96 bits divided by 5 = 19 remainder 1. Which was a perfect fit for 20 characters.
5 bit chunks means I needed a 5 bit alphabet, which would contain 32 characters. 0-9 and A-Z are obvious contenders (especially as there is a requirement from the business that the generated certificate number must not contain symbols/punctuation etc)
0-9 and A-Z gives us a character set of 36 characters, so I needed to lose 4 characters. Dropping B, I, O and S seemed obvious as these 4 characters can be misread as numbers.

I'm not sure the relevance of citing Softwire's implementation. They are generating a different reference for a different purpose; and by their own admission is not guaranteed unique (they have mentioned this several times and is the main reason they also have `applicationId` ... which is a Mongo `ObjectId` ¯\_(ツ)_/¯ )
Their character set is 30 characters, rather than our 32. And their rationale for excluding characters seems questionable (what number or other letter could be misread as an A, E or U ? Why not drop S and 5 ?, 8 and B ?)

Through their own admission their algorithm is not unique, and I would argue not particular thought through? But perhaps they have very different requirements to us, so perhaps is not right or fair to compare the 2 approaches? 

```
/**
 * Generate a user facing ref num.
 * The characters used to make the ref num are chosen uniformly at random using a
 * cryptographically strong method
 *
 * The refNum takes the structure of a prefix as described below, followed by 9 characters, each of which can be anything
 * from 0-9 or A-Z, except for 0, A, E, I, O, U.
 * We don't use these characters as they can be confusing when the refNum needs to be read from letters and emails
 *
 * @param prefixChar a prefix for the ref num.
 *  "A" - Register to Vote digital submissions
 *  "P" - Postal Vote digital submissions
 *  "R" - Proxy Vote digital submissions
 *  "V" - Voter Card digital submissions
 */
export function generateRefNum(prefixChar: string): string {
    const characterSet = "123456789BCDFGHJKLMNPQRSTVWXYZ";
    let refNumber: string = prefixChar;

    for (let i = 0; i < 9; i++) {
        // for randomInt, the minimum is inclusive while the maximum is exclusive, and we have 30 characters in the set
        const randomInt = crypto.randomInt(0, 30);
        refNumber += characterSet[randomInt];
    }
    return refNumber;
}
```

sentioservices:
One further thought. Passports use a 9 digit number. If there is no requirement to make the Certificate Number unguessable then following their example would be good because having fewer characters to type, all of which can be entered from a numeric keypad, will be much more efficient and less error prone when the number needs to be re-entered into the system. Uniqueness can still be achieved using a DB sequence.

sentioservices:
Hi @NathanRussellValtech,
Thanks for explaning your rational above.
I was thinking about the IDs used on the other forms of voter card identification and passports use a 9 digit number. If we were to use DB sequence to genereate such a number then it would also meet the requirement of being unique and being "up to 20 characters".
I like the idea of a number because come the point when users need to re-enter it:
- the fewer digits the user has to enter the less scope for them to get it wrong.
- it can be entered from a numeric keypad, arguably faster and less error prone for many users.
- less error prone when communicating verbally.
Admittedly I don't have evidence for these assertions, just my experience.
Was there some reason why you ruled out using a DB sequence?

NathanRussellValtech:
Hi @sentioservices 
I'll be honest, I didn't consider a DB sequence. 

Could we do it with a sequence? Yes, maybe - dynamo has Atomic Counters, so I'm sure we could get it working with that.

How does it solve your predictability / guess-ability concerns though?
As discussed earlier, even though non-guessable is not a requirement. if it were I'd suggest a simple number might be easier/more temping to guess (oh, its a number, I'll just add 1 to it?) vs. a string that is made up of apparent random alpha numeric characters (remembering also that to really be able to "guess" a cert number from the proposed 20 char string you'd need to "see" 2 or more certificate numbers in order to try to determine a pattern/algorithm)

sentioservices:
Hi @NathanRussellValtech, wrt predictability / guess-ability if it's not specified in the JIRA ticket, and other government documents like passports aren't worried about it then it isn't a concern for us.

PR #29, opened 2022-11-01
EIP1-2131 - Addition of CertificateNumberGenerator

NathanRussellValtech:
This PR is similar to my [previous PR](https://github.com/cabinetoffice/eip-ero-print-api/pull/27) in that it allows for the generation of Certificate Numbers, but this one uses a `CertificateNumberGenerator` bean rather than a `CertificateNumber` value object.

The concept and algorithm is identical though - the earlier PR uses a value object (where we are only interested in the toString() method); this PR uses a generator that returns the string

Which do we prefer?

The version using the value object has better tests as it is easier to test (via its overloaded constructors, which are really only there to support the tests. With the generator version we do not have that luxury

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateNumberGenerator.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateNumberGenerator.kt
@@ -1,0 +1,77 @@
+package uk.gov.dluhc.printapi.service
+
+import org.springframework.stereotype.Component
+import java.security.SecureRandom
+import java.time.Clock
+import java.time.Instant
+import java.util.concurrent.atomic.AtomicInteger
+
+/**
+ * A generator of globally unique Certificate Numbers, heavily based on [mongodb's ObjectId class](https://github.com/mongodb/mongo-java-driver/blob/master/bson/src/main/org/bson/types/ObjectId.java)
+ *
+ * A Certificate Number is a 20 character string using the character set `0123456789ACDEFGHJKLMNPQRTUVWXYZ`
+ * (specifically B, I, O and S are excluded as they can be misread as numbers)
+ *
+ * The Certificate Number is derived from 12 bytes, arranged as follows:
+ *
+ * ```
+ *   | 0 | 1 | 2 | 3 | 4 | 5 | 6 | 7 | 8 | 9 | 10 | 11 |
+ *   |---------------|-----------|-------|-------------|
+ *   | time          |rnd 1      | rnd 2 | increment   |
+ * ```
+ *
+ * Where:
+ *   * `time` is the timestamp since epoch rounded to second precision. The first 4 bytes of the integer are used.
+ *   * `rnd 1` is a random integer. This is a static member. The first 3 bytes of the integer are used.
+ *   * `rnd 2` is a random short. This is a static member. The first 2 bytes of the short are used.
+ *   * `increment` is a sequential increment, starting from a random integer. The first 3 bytes of the integer are used.
+ *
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateNumberGenerator.kt
```diff
@@ -0,0 +1,77 @@
...
+ *  the same `rnd 1` and `rnd 2` values. The net result is that bytes [4, 5, 6, 7, 8] of 2 instances generated in the same JVM
+ *  will be the same.
+ *
+ *  `increment` is an [AtomicInteger] and is a static member seeded from a random number. The use of `getAndIncrement()` means
+ *  this is a thread safe way of generating an incremental counter.
+ *
+ * The generated Certificate Number is a serialization of the 12 bytes.
+ */
+@Component
+class CertificateNumberGenerator(private val clock: Clock) {
```
matt-wills-valtech, line 39: Is it fair to say you've made this a bean (which requires a `Clock` instance), to aid testing?

Either way, to be honest I find it a peculiar case for a bean. I think a static factory method might be more appropriate - for example like `UUID.randomUUID()`.

I'm not going to lose sleep over it. I just think it's more of a "utility" type class, like `Math` etc (although we don't like the term utility or helper etc! 😆 )

NathanRussellValtech: Now thats interesting. The original Mongo `ObjectId` class does indeed have a static `get()` method (as well as the default constructor), though we don't use it (we call the default constructor)
IE. we could do:
```
val requestId = ObjectId.get()
```
When I created the `CertificateNumber` value object in the previous PR I decided not to do the static `get()` method, and instead use the constructors.

What I could do, and perhaps this has some mileage .... 🤔  ... is go back to the original PR with the value object and:
* make all of it's constructors `internal` so they can only be used by the tests
* implement a static (`companion object`) method that generates a `CertificateNumber` and returns it's string value:
```
companion object {
  ....
  fun next(): String {
    return CertificateNumber().toString()
  }
}
```
That would mean the public API of the `CertificateNumber` class is the static `next()` method, which returns a string.
And we would use it like this:
```
val certificateNumber = CertificateNumber.next()
```

Do you prefer that (I think I do 🤔 )

NathanRussellValtech: As discussed above, I've added a static factory method `create` to the original PR with the `CertificateNumber` class (to be able to compare and contrast the 2 approaches)
I named the method `create` rather than `next` as I felt it was better (`next` suggests a sequence, which is kinda is ... but mostly isn't 🤣 )
So, on the original PR the public API is now:
```
val certificateNumber: String = CertificateNumber.create()
```

matt-wills-valtech: That looks good to me. I appreciate some of this comes down to personal preference, but I think it's a common approach used elsewhere 👍

NathanRussellValtech: So to be clear @matt-wills-valtech , you are saying you prefer the original approach with the `CertificateNumber` value object with the static factory method? Is that right? (I think it gets my thumbs up as well FWIW)

matt-wills-valtech: That's right. The original approach with `CertificateNumber.create()`.

NathanRussellValtech:
As per the discussion on this PR, I think we prefer the original approach, so closing this PR without merging
The original PR has been merged

PR #32, opened 2022-11-02
Feature/EIP1-2261 Read printer response file

sentioservices:
This PR adds a listener to the Print-API service to listen for messages containing the names of the print response files. These files contain updates on the progress of batch and individual print requests. 
This request reads and deletes the files. The processing of the file will happen in EIP1-2262

TODO list
- [x] infra [PR ](https://github.com/cabinetoffice/eip-ero-infra/pull/227) merged?


src/test/resources/sftp/local/OutBound/status-20221101171156056.json
```diff
@@ -1 +1,19 @@
-Hello World 1
\ No newline at end of file
+{
```
vishalgupta-valtech, line 1: Addresses @NathanRussellValtech comments in my previous PR around having some valid values 🙏🏼

src/main/kotlin/uk/gov/dluhc/printapi/service/FilenameFactory.kt
```diff
@@ -22,5 +22,7 @@ class FilenameFactory(private val clock: Clock) {
 
     companion object {
         private var TIMESTAMP_FORMATTER: DateTimeFormatter = DateTimeFormatter.ofPattern("yyyyMMddHHmmssSSS")
+
+        fun createFileNamePath(fileDirectory: String, fileName: String) = "$fileDirectory/$fileName"
```
NathanRussellValtech, line 26: Is there any reason why this function is in the companion object (kinda like a java static) rather than an instance (bean) member? Just wondering why the change of approach ¯\_(ツ)_/¯

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseFileService.kt
```diff
@@ -0,0 +1,23 @@
...
+import org.springframework.stereotype.Service
+import uk.gov.dluhc.printapi.printprovider.models.PrintResponses
+
+private val logger = KotlinLogging.logger {}
+
+@Service
+class PrintResponseFileService(
+    val sftpService: SftpService,
+) {
+    fun fetchAndUnmarshallPrintResponses(filePathToProcess: String): PrintResponses =
```
NathanRussellValtech, line 13: Is this function (and also `removeRemoteFile`) adding anything by being in this bean? Could/should the `ProcessPrintResponseFileMessageListener` inject `SftpService` and call the methods directly ?

src/main/kotlin/uk/gov/dluhc/printapi/service/SftpService.kt
```diff
@@ -51,4 +53,19 @@ class SftpService(
...
         return newFileName
     }
+
+    /**
+     * Fetches the file from the path on the remote server and
+     * unmarshalls the content to an object of type `responseObjectType`
+     * @param filePathToProcess the path to the file on the remote server
+     * @param responseObjectType the type of the object to marshall the json content of the file to
+     */
+    fun <T> fetchAndUnmarshallFile(filePathToProcess: String, responseObjectType: Class<T>): T {
```
NathanRussellValtech, line 63: Kinda related to my previous comment, I dont know if its right that the SFTP service is responsible for deserializing into an object .... 🤔 
I know you've used generics, which means it could deserialize to any type, but might it be better if this method simply returned a string, then the thing that called it can worry about turning it into a thing that is relevant to that class/use case
(which would mean we _do_ need the method `fetchAndUnmarshallPrintResponses` in `PrintResponseFileService`)

(Also .... very minor naming question .... but is "unmarshall" the right terminology? To me (and I could be alone in this) "marshalling" and "unmarshalling" makes me think of XML and jaxb ??)

src/test/kotlin/uk/gov/dluhc/printapi/config/SftpContainerConfiguration.kt
```diff
@@ -27,13 +31,13 @@ class SftpContainerConfiguration {
...
             if (container == null) {
-                val makeHomeDirRunCommand = "mkdir -p /home/$USER/$PRINT_REQUEST_UPLOAD_PATH; chmod -R 007 /home/$USER"
+                val sw = StopWatch.createStarted()
+
                 container = GenericContainer(
                     ImageFromDockerfile()
                         .withDockerfileFromBuilder { builder ->
                             builder
                                 .from("atmoz/sftp:latest")
-                                .run(makeHomeDirRunCommand)
```
NathanRussellValtech, line 36: Do we not need to do this anymore ?

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseFileServiceIntegrationTest.kt
```diff
@@ -0,0 +1,137 @@
...
+        @ParameterizedTest
+        @CsvSource(
+            value = [
+                "0, 0",
+                "0, 1",
+                "1, 0",
+                "1, 1"
+            ]
+        )
+        fun `should fetch remote file and unmarshall to a PrintResponses object`(
```
NathanRussellValtech, line 35: Does the parameterized nature of this test add anything?
How is the scenario that has 1 BatchResponse and 1 PrintResponse different to any other combination?

src/test/kotlin/uk/gov/dluhc/printapi/service/SftpServiceTest.kt
```diff
@@ -31,11 +32,14 @@ internal class SftpServiceTest {
...
 
+    @Mock
+    private lateinit var objectMapper: ObjectMapper
+
     private lateinit var sftpService: SftpService
 
     @BeforeEach
     fun setUp() {
-        sftpService = SftpService(sftpInboundTemplate, sftpOutboundTemplate)
+        sftpService = SftpService(sftpInboundTemplate, sftpOutboundTemplate, objectMapper)
```
NathanRussellValtech, line 42: I know you didnt add this test class and BeforeEach hook; but I wonder why we are not using `@InjectMocks` ? 🤔

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/testdata/model/PrintResponseBuilder.kt
```diff
@@ -0,0 +1,41 @@
...
+fun buildPrintResponses(
+    batchResponses: List<BatchResponse> = listOf(buildBatchResponse()),
+    printResponses: List<PrintResponse> = listOf(buildPrintResponse()),
+): PrintResponses =
+    PrintResponses()
+        .withBatchResponses(batchResponses)
+        .withPrintResponses(printResponses)
+
+fun buildBatchResponse(
+    batchId: String = Random.nextInt().toString(),
```
NathanRussellValtech, line 20: You can use `aValidBatchId()` here - it better reflects what a batch ID looks like 👍

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/testdata/model/PrintResponseBuilder.kt
```diff
@@ -0,0 +1,41 @@
...
+    .withMessage(message)
+    .withStatus(status)
+    .withTimestamp(timestamp)
+
+fun buildPrintResponse(
+    requestId: UUID = UUID.randomUUID(),
+    message: String = DataFaker.faker.famousLastWords().lastWords(),
+    statusStep: PrintResponse.StatusStep = PrintResponse.StatusStep.IN_PRODUCTION,
+    status: PrintResponse.Status = PrintResponse.Status.SUCCESS,
+    timestamp: Date = Date(),
```
NathanRussellValtech, line 35: Hmmm, that's interesting .... I know this is a generated class, but AFAIK the date fields should all be ISO 8601, so I'd be expecting this to be a java 8 date class of some kind .... 🤔 
I'll have a look at the yamlschema for this field .... 🤔

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/testdata/model/PrintResponseBuilder.kt
```diff
@@ -0,0 +1,41 @@
...
+    status: BatchResponse.Status = BatchResponse.Status.SUCCESS,
+    timestamp: Date = Date(),
+) = BatchResponse()
+    .withBatchId(batchId)
+    .withMessage(message)
+    .withStatus(status)
+    .withTimestamp(timestamp)
+
+fun buildPrintResponse(
+    requestId: UUID = UUID.randomUUID(),
```
NathanRussellValtech, line 31: I dont think requestId should be a UUID either .... 🤔 
It's meant to be a 24 char hexstring, hence we use mongo's ObjectId

Something's not right (in the yamlschema I suspect)

src/test/resources/sftp/local/OutBound/status-20221101171156056.json
```diff
@@ -1 +1,19 @@
...
+  "BatchResponses": [
+    {
+      "batchId": "1194720300",
```
NathanRussellValtech, line 4: BatchId is a UUID without the dashes (32 chars)

src/test/resources/sftp/local/OutBound/status-20221101171156056.json
```diff
@@ -1 +1,19 @@
...
+      "status": "SUCCESS",
+      "timestamp": "2022-11-03T17:51:51.777+00:00",
+      "message": "This is a Batch response. Hooray!"
+    }
+  ],
+  "PrintResponses": [
+    {
+      "requestId": "c27003dc-5e41-4a74-9032-3af706acaf01",
```
NathanRussellValtech, line 12: requestId is a 24 char mongo ID

NathanRussellValtech (commented):
Hi @sentioservices - can you hold off merging this please until I've sorted out the data types in the yamlschema (and the resultant generated classes) 👍

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseFileService.kt
matt-wills-valtech: Is the intention to abstract away the underlying implementation (in this case sftp) from the Listener?

sentioservices (commented):
Updated following PR comments

src/main/kotlin/uk/gov/dluhc/printapi/service/SftpService.kt
```diff
@@ -1,6 +1,8 @@
 package uk.gov.dluhc.printapi.service
 
+import com.fasterxml.jackson.databind.ObjectMapper
 import com.jcraft.jsch.ChannelSftp
+import com.nimbusds.jose.util.IOUtils
```
NathanRussellValtech, line 5: Is ObjectMapper still used in this class?

Is there a better IOUtils to use here? One from apache commons or springframework perhaps?
This one from nimbusds.jose is part of Spring Security, and specifically the JWT support stuff I think ?

src/test/kotlin/uk/gov/dluhc/printapi/service/SftpServiceTest.kt
NathanRussellValtech: Awesome; thank you 👍

NathanRussellValtech (approved):
LGTM; just a minor comment on whether ObjectMapper is still required in SftService; and also where we get IOUtils from
Other than that, looking good 👍 Thanks for addressing the previous comments 👍

src/main/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintResponseFileMessageListener.kt
```diff
@@ -0,0 +1,41 @@
...
+    @SqsListener("\${sqs.process-print-response-file-queue-name}")
+    override fun handleMessage(@Valid @Payload payload: ProcessPrintResponseFileMessage) {
+        with(payload) {
+            val filePath = "$directory/$fileName"
+            logger.info { "Begin processing PrintResponse file [$filePath]" }
+            val printResponsesString = sftpService.fetchFile(filePath)
+            val printResponses = objectMapper.readValue(printResponsesString, PrintResponses::class.java)
+            processPrintResponses(printResponses)
+            sftpService.removeFileFromOutBoundDirectory(filePath)
+            logger.info { "Completed processing PrintResponse file [$filePath]" }
```
Neil-Massey, line 33: I would have expected a service class to be responsible for managing these details and listener only calls that service method.

Service should handle all exceptions that should not cause the SQS message to be retried.

matt-wills-valtech: I'd vote for that - sounds like a good idea.

src/main/kotlin/uk/gov/dluhc/printapi/service/SftpService.kt
```diff
@@ -51,4 +54,18 @@ class SftpService(
         sftpOutboundTemplate.rename("$fileDirectoryPath/$originalFileName", "$fileDirectoryPath/$newFileName")
         return newFileName
     }
+
+    /**
+     * Fetches the file from the path on the remote server
+     * @param filePathToProcess the path to the file on the remote server
+     * @return the contents of the remote file
+     */
+    fun fetchFile(filePathToProcess: String): String {
```
vishalgupta-valtech, line 63: 1. To be in line with other method you could rename this to `fetchFileFromOutBoundDirectory`?
2. Ideally you could pass the `directory` and `fileName` to this method. Same for `removeMethod`

src/test/resources/sftp/local/OutBound/invalid-json.json
```diff
@@ -0,0 +1 @@
+- Invalid : JSON
```
vishalgupta-valtech, line 1: I think the filename should still be as per fileNaming pattern of  `status-d17.json` . The content seems ok for an invalid json

sentioservices: I've deleted this file because the json processing wasn't required as part of this ticket.

src/main/kotlin/uk/gov/dluhc/printapi/service/SftpService.kt
sentioservices: I've updated the method names as you suggested 👍

src/main/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintResponseFileMessageListener.kt
sentioservices: I've abstracted this functionality into uk.gov.dluhc.printapi.service.PrintResponseFileService#processPrintResponseFile 👍

src/main/kotlin/uk/gov/dluhc/printapi/service/SftpService.kt
sentioservices: Thanks, I hadn't spotted that rogue import.

matt-wills-valtech (approved):

Neil-Massey (approved):

PR #33, opened 2022-11-03
EIP1-2343 - New required fields in send to print SQS message

matt-wills-valtech:
Future stories will expose REST APIs on the print-api that need to return data about the print request (VAC Certificate).

The data that these REST APIs need to return include attributes relating to the Voter Card Application as well as the certificate print request. This PR adds two additional fields to the SQS message which VCA will need to send to this API:

* The application received datetime
* The delivery mechanism (delivery or collection)

These two fields are stored in Dynamo, but not sent to the print provider. Hence the scope of this change goes as far as `PrintDetails` enity.

Example JSON (refer to `applicationReceivedDateTime` and `deliveryMethod`):

```
{
  "sourceReference": "string",
  "applicationReference": "string",
  "sourceType": "voter-card",
  "gssCode": "stringstr",
  "requestDateTime": "2022-06-02T13:20:01.000Z",
  "applicationReceivedDateTime": "2022-06-01T00:00:00.000Z",
  "firstName": "John",
  "middleNames": "Malcolm",
  "surname": "Smith",
  "certificateLanguage": "en",
  "photoLocation": "string",
  "delivery": {
    "deliveryClass": "standard",
    "deliveryMethod": "delivery",
    "addressee": "string",
    "address": {
      "street": "string",
      "property": "string",
      "locality": "string",
      "town": "string",
      "area": "string",
      "postcode": "string",
      "uprn": "1695839383"
    }
  }
}
```

Corresponding PR for VCA: https://github.com/cabinetoffice/eip-ero-voter-card-applications-api/pull/267


NathanRussellValtech (approved):
LGTM, though not sure how the mapping happens between the SQS message and the enum (deliveryMethod comes in as lower case in the SQS message, but the enum values are uppercase)

PR #36, opened 2022-11-03
EIP1-564: EIP1-2261: Schema definition for process print response

vishalgupta-valtech:
image

### DIFF
--- a/src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
+++ b/src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Print API SQS Message Types
-  version: '1.4.0'
+  version: '1.5.0'
   description: |-
     Print API SQS Message Types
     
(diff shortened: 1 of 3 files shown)


src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Print API SQS Message Types
-  version: '1.4.0'
+  version: '1.5.0'
   description: |-
     Print API SQS Message Types
```
NathanRussellValtech, line 6: Can you put a dummy `path` and `post` method in here please (like we do with the others)
I know its not a REST API !! (you're telling the wrong man 😁 ) , but we're using the `path` elements to "document" [sic] the queue which then hooks up to the model as its request payload.
Have a look at what we've done with the others

vishalgupta-valtech: Ofcourse 😆 . Code pushed 👍🏼

NathanRussellValtech (approved):
LGTM 👍

PR #37, opened 2022-11-03
EIP1-2323 - PrintDetails now has a list of PrintRequestStatus

NathanRussellValtech:
This PR introduces a list of print request statuses on our `PrintDetails` dynamo entity, and to change the existing `status` item to return the last/most recent event in the list of `printRequestStatuses`.

The most recent event is determined by a new `eventDateTime` field, which is either the timestamp that we created the entry, or the timestamp from the print provider's response file, depending on who processed the event.


src/main/kotlin/uk/gov/dluhc/printapi/database/entity/PrintDetails.kt
```diff
@@ -32,9 +33,22 @@ data class PrintDetails(
...
     var eroWelsh: ElectoralRegistrationOffice? = null,
-    @get:DynamoDbSecondaryPartitionKey(indexNames = [STATUS_BATCH_ID_INDEX_NAME]) var status: Status = Status.PENDING_ASSIGNMENT_TO_BATCH,
-    @get:DynamoDbSecondarySortKey(indexNames = [STATUS_BATCH_ID_INDEX_NAME])var batchId: String? = null
+    var printRequestStatuses: MutableList<PrintRequestStatus>? = null,
+    @get:DynamoDbSecondarySortKey(indexNames = [STATUS_BATCH_ID_INDEX_NAME]) var batchId: String? = null
 ) {
+
+    var status: Status? = null
+        @DynamoDbSecondaryPartitionKey(indexNames = [STATUS_BATCH_ID_INDEX_NAME])
+        get() = printRequestStatuses?.sortedBy { it.dateTime }?.last()?.status
```
Neil-Massey, line 42: I wouldn't a no-op setter would be required for this to work

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/testdata/entity/PrintDetailsBuilder.kt
```diff
@@ -69,7 +72,17 @@ fun buildPrintDetails(
     eroEnglish = eroEnglish,
     eroWelsh = eroWelsh,
     batchId = batchId,
-    status = status,
+    printRequestStatuses = printRequestStatuses,
+)
+
+fun aValidPrintDetailsWithSingleStatus(
```
Neil-Massey, line 78: I'm not clear why you are being so descriptive about fact it was a single status... Can we overload the `aValidPrintDetails` with one taking a single status and other a list?

NathanRussellValtech: Thats a fair shout, I can overload the original 👍 

I guess I was coming from what we used to do in previous java projects, where we would have several builders all describing the thing to be built, eg:
```
class PizzaTestDataBuilder() {

    public static Pizza aPizza(String... toppings) ....

    public static Pizza aMeatFeastPizza() {
        return aPizza("Ham", "Chicken", "Pepperoni", "Spicy Ground Beef");
    }

    public static Pizza aPepperoniPizza() {
        return aPizza("Pepperoni");
    }
}
```

(Friday night is pizza night in the Russell household! 😁 )

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/PrintDetails.kt
NathanRussellValtech: Now, here's the thing ... the setter is called by dynamo's deserialization code, but from what I can tell it doesn't really make much different / have any real effect because when dynamo serializes it uses the getter. So a no-op, or the default impl that sets the backing field (that is promptly ignored by virtue of the getter) seems to make no difference.

What I did try was making the setter throw an `UnsupportedOperationException` - my thinking was that I wanted to actively prevent developers from calling `printDetails.status = someStatus`. But because dynamo calls the setter, dynamo obviously started failing.

I'll have a tinker and see if I can make this better / clearer in some way ....

NathanRussellValtech: Just had a play, and it looks like dynamo attribute mapping only works with a class property/field (rather than a method pretending to be a property). So this didn't work:
```
    @DynamoDbSecondaryPartitionKey(indexNames = [STATUS_BATCH_ID_INDEX_NAME])
    @DynamoDbAttribute("status")
    fun status(): Status? {
        return printRequestStatuses?.sortedBy { it.dateTime }?.last()?.status
    }
```

Well, more accurately, the tests **did** pass, but thats because we assert the value of the `status` field (method) and it still returns the correct value. What didn't work was that on the actual dynamo table it had not stored the item with a `status` attribute (so other things such as manual queries in the AWS console, or other code, that try to use the `status` attribute wont work because that attribute doesn't exist)

NathanRussellValtech: Boo, hiss!!

> In order for the AWS SDK v2 for Java’s Enhanced Client to detect an attribute on a @DynamoDbBean, the field must have a setter. In other words, just having a getter is not enough, even for derived attributes. This may or may not be obvious, but forgetting the setter for an attribute can result in a misleading error, or the attribute being silently ignored ...

At least that explains why the properties in the entity class must be `var` and not `val` (`val`'s do not have a setter)

https://davidagood.com/dynamodb-enhanced-client-java-missing-setters/

NathanRussellValtech: Added a no-op setter, and marked it as deprecated with a suitable explanation, which hopefully will prevent/deter devs from calling the setter. Its the best I could come up with ¯\_(ツ)_/¯

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/testdata/entity/PrintDetailsBuilder.kt
NathanRussellValtech: Renamed 👍

NathanRussellValtech (commented):
This PR looks like a thing of actual beauty! I'd love to approve it, but Github as other ideas ... saying that I can't approve a PR that I originally raised!
How rude!

matt-wills-valtech (approved):

matt-wills-valtech:
> This PR looks like a thing of actual beauty! I'd love to approve it, but Github has other ideas ... saying that I can't approve a PR that I originally raised! How rude!

Fortunately, I can approve it 😉

PR #46, opened 2022-11-08
EIP1-2413 Add userId to SendApplicationToPrintMessage

matt-wills-valtech:
Future stories will expose REST APIs on the print-api that need to return data about the print request (VAC Certificate). The data that these REST APIs need to return include attributes relating to the Voter Card Application as well as the certificate print request.

A previous PR added `applicationReceivedDateTime` and `deliveryMethod` to the SQS message which VCA will need to send to this API. This PR adds the ID of the user that is requesting the print.

Corresponds to https://github.com/cabinetoffice/eip-ero-voter-card-applications-api/pull/272


NathanRussellValtech (approved):
LGTM 👍

PR #47, opened 2022-11-08
EIP1-2386 Add certificate tables

Neil-Massey:


build.gradle.kts
```diff
@@ -100,6 +107,7 @@ dependencies {
     testImplementation("org.mockito.kotlin:mockito-kotlin:4.0.0")
 
     testImplementation("org.testcontainers:testcontainers:1.17.4")
+    testImplementation("org.testcontainers:mysql:1.17.5")
```
vishalgupta-valtech, line 110: Minor : It will be nice to keep all test containers dependency on same 1.17.5 version

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Address.kt
```diff
@@ -0,0 +1,83 @@
...
+
+    @Size(max = 255)
+    var area: String? = null,
+
+    @Size(max = 12)
+    var uprn: String? = null,
+
+    @NotNull
+    @UpdateTimestamp
+    var dateCreated: Instant? = null,
```
vishalgupta-valtech, line 58: I am aware we are doing something similar in vca with `dateCreated as @UpdateTimestamp ` 
Should we use @CreationTimestamp instead (not a blocker, just a question)

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Address.kt
```diff
@@ -0,0 +1,83 @@
...
+    @NotNull
+    @Size(max = 255)
+    var street: String? = null,
+
+    @NotNull
+    @Size(max = 10)
+    var postcode: String? = null,
+
+    @Size(max = 255)
+    var `property`: String? = null,
```
vishalgupta-valtech, line 42: Why `single quotes` around property? Perhaps a typo?

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Certificate.kt
```diff
@@ -0,0 +1,118 @@
...
+        other as Certificate
+
+        return id != null && id == other.id
+    }
+
+    override fun hashCode(): Int = javaClass.hashCode()
+
+    @Override
+    override fun toString(): String {
+        return this::class.simpleName + "(id = $id , certificateNumber = $vacNumber , gssCode = $gssCode, dateCreated = $dateCreated , createdBy = $createdBy , version = $version )"
```
vishalgupta-valtech, line 116: Q: Should we leak `certificateNumber` in logs?

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/ElectoralRegistrationOffice.kt
```diff
@@ -0,0 +1,81 @@
...
+    @Size(max = 255)
+    var name: String? = null,
+
+    @NotNull
+    @Size(max = 20)
+    var phoneNumber: String? = null,
+
+    @NotNull
+    @Size(max = 255)
+    var emailAddress: String? = null,
```
vishalgupta-valtech, line 45: `emailAddress` could be null in VCA for an applicant. So hopefully all ERO's do have it :)

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/PrintRequestStatus.kt
```diff
@@ -0,0 +1,70 @@
...
+    @GeneratedValue(strategy = GenerationType.IDENTITY, generator = "UUID")
+    @GenericGenerator(name = "UUID", strategy = UseExistingOrGenerateUUID.NAME)
+    var id: UUID? = null,
+
+    @NotNull
+    @Size(max = 255)
+    var status: Status? = null,
+
+    @NotNull
+    var eventDateTime: Instant? = null, // either the "timestamp" from the print provider, or the current time
```
vishalgupta-valtech, line 39: `eventDateTime` is marked as non nullable. However Initially this `eventDateTime` could be null but I remember @matt-wills-valtech mentioned that we will set this to same time as that of dateCreated when initial record is created for print request status

src/main/resources/db/changelog/db.changelog-master.xml
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+         http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd">
```
vishalgupta-valtech, line 7: It will be nice to use latest 4.1 xsd?

```
<databaseChangeLog
    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
```

vishalgupta-valtech (commented):
LGTM some minor queries/comments

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/PrintRequest.kt
```diff
@@ -0,0 +1,133 @@
...
+import javax.persistence.OneToOne
+import javax.persistence.Table
+import javax.persistence.Version
+import javax.validation.constraints.NotNull
+import javax.validation.constraints.Size
+
+@Table
+@Entity
+@EntityListeners(AuditingEntityListener::class)
+class PrintRequest(
```
adacayi, line 36: @Neil-Massey , will the `PrintRequests` be audited? 
If yes, do we need the @Audited annotation in the entity as well?

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Address.kt
Neil-Massey: I assumed it must be a Kotlin reserved word

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Certificate.kt
Neil-Massey: Probably not - I'll remove.. most only have id and audit details

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/PrintRequest.kt
```diff
@@ -0,0 +1,133 @@
...
+    @Id
+    @Type(type = UUIDCharType)
+    @GeneratedValue(strategy = GenerationType.IDENTITY, generator = "UUID")
+    @GenericGenerator(name = "UUID", strategy = UseExistingOrGenerateUUID.NAME)
+    var id: UUID? = null,
+
+    @NotNull
+    @Size(max = 24)
```
adacayi, line 44: @Neil-Massey, I think the current implementation does not validate the entity.
We can make it work like this:
<pre>
@<b>field:</b>Size(min=24, max = 24)
</pre>


Do you think it is worth covering the validation in an integration test as well?

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/ElectoralRegistrationOffice.kt
Neil-Massey: I think this relates to the ERO (not an applicant)

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/PrintRequest.kt
Neil-Massey: Discussed with Nathan and there is to be no audit tables for this project as apart from the `latest status` and the addition of new print requests and events the schema is immutable.

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/PrintRequest.kt
Neil-Massey: TBH I'm not sure about the value of these annotations I included as saw they are used in VCA.

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Address.kt
matt-wills-valtech: I would vote for 2 separate fields - dateCreated and lastUpdated with the relevant annotations 👍

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Certificate.kt
matt-wills-valtech: Paranoid 😉

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/PrintRequestStatus.kt
matt-wills-valtech: It should never be null - as you said it will be set to the same time as dateCreated when an initial record is created.

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/PrintRequest.kt
matt-wills-valtech: I think they're worth having (so long as they're in sync with VCA), but I don't think we need the overheard of integration tests for them.

src/main/kotlin/uk/gov/dluhc/printapi/rds/repository/UserPrincipalAuditorAware.kt
```diff
@@ -0,0 +1,23 @@
...
+        const val SYSTEM: String = "system"
+    }
+
+    override fun getCurrentAuditor(): Optional<String> {
+        return with(SecurityContextHolder.getContext().authentication) {
+            if (this != null && this.isAuthenticated)
+                Optional.of(this.name)
+            else
+                Optional.of(SYSTEM)
+        }
```
sentioservices, line 21: for this to work I think we'd need to propagate the user in the sqs message?

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Address.kt
matt-wills-valtech: Having said that, it's different for immutable entities. The `Address` entity in RCA just has a `dateCreated` field with the `@CreationTimestamp` annotation:

```
    @NotNull
    @CreationTimestamp
    var dateCreated: Instant? = null,
```

Mutable objects also have `updatedAt` and `version`:
```
    @NotNull
    @UpdateTimestamp
    var updatedAt: Instant? = null,

    @Version
    var version: Long? = null,
```

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/PrintRequest.kt
adacayi: Thanks, @Neil-Massey ,
In that case do we need the `@EntityListeners(AuditingEntityListener::class)`?

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/PrintRequest.kt
adacayi: @Neil-Massey, @matt-wills-valtech one thing I wanted to point out is the constraint annotations like `@Size` needs to be for the field, for validation to work. Otherwise, it won't work.

For that we need to add `field` to the annotation as below:

`@field:Size`

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/PrintRequest.kt
Neil-Massey: I think that is required for Spring to process the annotations on created, created_by fields - but will double check and remove if possible.

src/test/kotlin/uk/gov/dluhc/printapi/rds/repository/CertificateRepositoryTest.kt
```diff
@@ -0,0 +1,108 @@
...
+        printRequest.addPrintRequestStatus(printRequestStatus)
+        certificate.addPrintRequest(printRequest)
+        val expected = certificateRepository.save(certificate)
+
+        // When
+        val actual = certificateRepository.findById(expected.id!!)
+
+        // Then
+        assertThat(actual).isPresent
+        assertThat(actual.get()).isEqualTo(expected)
```
adacayi, line 106: @Neil-Massey, this will use the equals method of the Certificate, which only checks the equality of the id. To check if expected and actual are equal for all the fields we need to use recursive comparison.

<pre>
assertThat(actual.get()).<b>usingRecursiveComparison()</b>.isEqualTo(expected)
</pre>

vishalgupta-valtech (approved):
LGTM just a couple of minor comments. 👍🏼

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Certificate.kt
```diff
@@ -0,0 +1,118 @@
...
+    @NotNull
+    var issueDate: LocalDate = LocalDate.now(),
+
+    @NotNull
+    var suggestedExpiryDate: LocalDate = issueDate.plusYears(10),
+
+    @NotNull
+    @Size(max = 50)
+    @Enumerated(EnumType.STRING)
+    var status: Status = Status.PENDING_ASSIGNMENT_TO_BATCH,
```
adacayi, line 76: @Neil-Massey , looking into the current implementation, the status needs to be inferred from the print request statuses by taking the latest status.
 
Do you agree?

For the context, this is the current implementation with dynamodb:

<pre>
    var status: Status?
        @DynamoDbSecondaryPartitionKey(indexNames = [STATUS_BATCH_ID_INDEX_NAME])
        get() = printRequestStatuses?.sortedBy { it.eventDateTime }?.last()?.status
        @Deprecated(
            """
            Programmatically setting the status property is not supported and will have no effect.
            The status property and its setter are provided so that dynamodb persists a status attribute, where the value
            is the last element from the printRequestStatuses list (element with the most recent eventDateTime).
            To set the status add a new PrintRequestStatus to the printRequestStatuses list.
            """
        )
        set(_) { /* no-op setter */ }
</pre>

Neil-Massey: Yes - It's actually a little bit more involved now as we have Certificate -> List<Requests> -> List<Statuses>
This PR was mainly around the Liquibase and Entities... but you're right how the status is to be managed needs updating...

vishalgupta-valtech: My understanding was when we completely migrate to `mysql` then we anyways need to revisit the bits done using `DynamoDB`. So in the scope of current PR this may not be needed.

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/PrintRequest.kt
NathanRussellValtech: Yes. The point @abdullah-sanver-valtech is making is that because this is a class, and the "fields" here are constructor arguments, simply annotating with `@Size` is annotating the constructor argument, not the backing field itself.
Abdullah is right, we need to use `@field` as a prefix if we expect it to work ....

... which is another discussion in itself .... do we even want/care about these annotations on the entity fields? What is going to actually process them? More importantly, what will happen if it fails said validation ?

(re: the annotation, the kotlin terminology we are talking about here is "annotation use site targets" - https://kotlinlang.org/docs/annotations.html#annotation-use-site-targets )

matt-wills-valtech: If the annotations are processed (ignoring the specifics of that for a moment), then I would assume it would be best to fail early (hopefully with a meaningful error message), rather than either fail to insert the data, or insert "invalid" data?

build.gradle.kts
Neil-Massey: Done

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Address.kt
Neil-Massey: Yes, C&P from VAC.. will use created annotation 👍

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Address.kt
Neil-Massey: Removed, it was probably related to a reserved word in DynamoDB - not Kotlin or MySQL.

build.gradle.kts
```diff
@@ -75,6 +75,13 @@ dependencies {
     implementation("org.springframework.boot:spring-boot-starter-security")
     implementation("org.springframework.boot:spring-boot-starter-oauth2-resource-server")
 
+    // mysql
+    implementation("org.springframework.boot:spring-boot-starter-data-jpa")
+    implementation("org.liquibase:liquibase-core")
+    runtimeOnly("mysql:mysql-connector-java")
+    runtimeOnly("software.aws.rds:aws-mysql-jdbc:1.1.1")
```
NathanRussellValtech, line 82: Adding the AWS MySql deps here will trigger a load of owasp dependency scanner false positives! 😢 
Please can you add the relevant suppression to the owasp suppressions file (copy the relevant bit from VCA or RCA's version of the file 👍 )

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Certificate.kt
```diff
@@ -0,0 +1,118 @@
...
+    @NotNull
+    @Size(max = 20)
+    @Enumerated(EnumType.STRING)
+    var sourceType: SourceType? = null,
+
+    @NotNull
+    @Size(max = 255)
+    var sourceReference: String? = null,
+
+    @Size(max = 255)
```
NathanRussellValtech, line 57: Minor - should this be `@NotNull` as well?
(not that I like these constraint validations on our JPA entities. I think they are noise and dont add any function/behaviour/benefit other than perhaps a visual clue to a dev reading the source code? Maybe I'm wrong? Maybe its just me ¯\_(ツ)_/¯ )

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Certificate.kt
```diff
@@ -0,0 +1,118 @@
...
+    @NotNull
+    var suggestedExpiryDate: LocalDate = issueDate.plusYears(10),
+
+    @NotNull
+    @Size(max = 50)
+    @Enumerated(EnumType.STRING)
+    var status: Status = Status.PENDING_ASSIGNMENT_TO_BATCH,
+
+    @NotNull
+    @Size(max = 80)
```
NathanRussellValtech, line 79: And this is an example of where these annotations dont add anything, and in this case is plain misleading ... I think the max size for a gss code is  either 9 or 10 (I forget which - its either [E|S|W|N] and 8 digits, or [E|S|W|N] and 9 digits
Its defo not 80 😁  ¯\_(ツ)_/¯

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Delivery.kt
```diff
@@ -0,0 +1,79 @@
...
+    @NotNull
+    @Size(max = 255)
+    var addressee: String? = null,
+
+    @OneToOne(cascade = [CascadeType.ALL])
+    var address: Address? = null,
+
+    @NotNull
+    @Size(max = 20)
+    var deliveryClass: DeliveryClass? = null,
```
NathanRussellValtech, line 46: Do these 2 enum fields need this ? :
```
@Enumerated(EnumType.STRING)
```
(I can never remember whether you **need** them or not ¯\_(ツ)_/¯

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/PrintRequest.kt
NathanRussellValtech: For my money I would expect any validations to happen long before we go near any persistence layer - request payload validation if it were a REST API or SQS message. How else is data getting into the system? Validate at the point it enters the system?

src/test/kotlin/uk/gov/dluhc/printapi/config/IntegrationTest.kt
```diff
@@ -97,9 +99,6 @@ internal abstract class IntegrationTest {
     @Autowired
     protected lateinit var processPrintResponseFileMessageQueue: MessageQueue<ProcessPrintResponseFileMessage>
 
-    @Autowired
```
NathanRussellValtech, line 100: Might need this back in (to prevent a merge conflict). I know it wasnt used, but I put it there ready to be used by @fayimora and @abdullah-sanver-valtech in their current tickets; and I think they'll be wanting it there ¯\_(ツ)_/¯

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/testdata/RandomUtils.kt
```diff
@@ -19,8 +30,52 @@ fun getAMongoDbId(): String = ObjectId().toHexString()
 
 fun aValidRequestId(): String = getAMongoDbId()
 
+fun aValidRequestDateTime(): OffsetDateTime = OffsetDateTime.now()
```
NathanRussellValtech, line 33: I'm not saying the original file was perfect in this regard, but are these new functions you've added really random?

Either the source filename is wrong (`RandomUtils` [sic]), or we should move some of these functions out into more specialised source files?

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/PrintRequest.kt
Neil-Massey: OK - Have activated the annotation for Hibernate validation, and removed all that prevented saving (cannot have size on enums and the spring managed audited properties cannot be null checked)

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Certificate.kt
Neil-Massey: I agree - copied from VCA (for consistency) but as they are missing the `field:` specification -  I doubt Hibernate is enforcing anything.

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Certificate.kt
Neil-Massey: I just copied definition from VCA:
```
              - column:
                  name: gss_code
                  type: varchar(80)
                  constraints:
                    nullable: false
```

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Delivery.kt
Neil-Massey: I thought it was a JPA thing as it creates as null and then sets values

src/main/resources/db/changelog/ddl/0001_create_certificate_tables.xml
```diff
@@ -0,0 +1,248 @@
...
+    </changeSet>
+
+    <changeSet author="neil.massey@valtech.com" id="0001-EIP1-2386-create-print-request-table" context="ddl">
+        <createTable tableName="print_request">
+            <column name="id" type="uuid">
+                <constraints nullable="false" primaryKey="true" unique="true"
+                             primaryKeyName="print_request_pkey"/>
+            </column>
+            <column name="certificate_id" type="uuid">
+                <constraints nullable="false"
```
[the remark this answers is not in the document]
Neil-Massey: Yes `FK` for the win!

build.gradle.kts
Neil-Massey: Suppressed 👍

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Certificate.kt
Neil-Massey: It doesn't get mapped into what we send the printer and the `sourceReference` and `sourceType` are used for correlation with `VCA`... no idea why it's included in request to print API

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/testdata/RandomUtils.kt
Neil-Massey: Moved fixed values

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Certificate.kt
Neil-Massey: Updated to set the correct status when saving the certificate.  
Presumably most of the task processing will be based on the Certificate status (and not the PrintRequest) - although the events are against the individual request.

PR #48, opened 2022-11-08
EIP1-1878 - Fix mapping of ERO fields in print request entity

NathanRussellValtech:
This PR fixes the mapping of the ERO fields in the `PrintRequest` entity.
At the moment the fields that were missing (ERO phone number, email address, website, address street and address postcode) are mapped to empty string as we don't have the final values yet. This will be updated in a future PR


Neil-Massey (approved):
LGTM

PR #51, opened 2022-11-09
EIP1-2262 - Added queue and basic listener for processing print response message

NathanRussellValtech:
This PR is an enabler for the real work thats being done in EIP1-2262

It:
* adds a new message type `ProcessPrintResponseMessage` to the SQS schema and codegens it
* adds a new queue bean for the new type, so that messages can be sent to it (future work in 2262)
* adds a new listener for the new type, so that messages can be consumed (future work on 2262)
* documents the environment variable required for the queue name

The corresponding infra changes (create the queue, wire the env var into the container) have already been done and merged

Next tasks after this are to:
* Add the updated SQS schema to VCA (even though VCA wont be publishing or consuming messages of this type)
* Add the queue and env var to our docker-compose setup

### DIFF
--- a/src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
+++ b/src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Print API SQS Message Types
-  version: '1.6.0'
+  version: '1.7.0'
   description: |-
     Print API SQS Message Types
     
@@ -49,6 +49,16 @@ paths:
       responses:
         '204':
           description: No response content.
+  '/process-print-response':
+    post:
+      tags:
+        - SQS Queues
+      requestBody:
+        $ref: '#/components/requestBodies/ProcessPrintResponseMessage'
+      responses:
+        '204':
+          description: No response content.
+
 components:
   #
   # Schema and Enum Definitions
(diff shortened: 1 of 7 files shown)


src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Print API SQS Message Types
-  version: '1.6.0'
+  version: '1.7.0'
```
matt-wills-valtech, line 4: Don't forget to update VCA 😄

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -242,6 +252,46 @@ components:
...
+            - PROCESSED
+            - IN-PRODUCTION
+            - DISPATCHED
+            - NOT-DELIVERED
+        status:
+          type: string
+          enum:
+            - SUCCESS
+            - FAILED
+          description: Indicates if the step change was successful or not.
```
matt-wills-valtech, line 284: So this represents what we receive for each PrintResponse:

```
      {
        "requestId": "507f1f77bcf86cd799439011",
        "timestamp": "2022-09-29T11:03:20.123Z",
        "statusStep": "PROCESSED",
        "status": "SUCCESS"
      }
```

matt-wills-valtech (approved):

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
NathanRussellValtech: Yes, exactly. With an optional `message` field in the case of `status` being `FAILED`

PR #53, opened 2022-11-09
EIP1-2262: EIP1-2419: Implement method to process print response

adacayi:
This PR is to implement service method to process individual print responses received by the print provider


src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
```diff
@@ -0,0 +1,92 @@
...
+                }
+
+                printDetailsRepository.updateItems(printDetails)
+            }
+        }
+    }
+
+    fun processPrintResponse(printResponse: ProcessPrintResponseMessage) {
+        val newStatus = with(printResponse) {
+            when {
```
fayimora, line 67: I don't know if kotlin has pattern matching but can we do  something like 

```
when ((statusStep, status)) {
  (ResponseStatusStep.PROCESSED, ResponseStatus.SUCCESS) -> VALIDATED_BY_PRINT_PROVIDER
}
```

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
```diff
@@ -0,0 +1,92 @@
...
+                statusStep == ResponseStatusStep.IN_MINUS_PRODUCTION && status == ResponseStatus.SUCCESS -> IN_PRODUCTION
+                statusStep == ResponseStatusStep.DISPATCHED && status == ResponseStatus.SUCCESS -> DISPATCHED
+                statusStep == ResponseStatusStep.NOT_MINUS_DELIVERED && status == ResponseStatus.FAILED -> NOT_DELIVERED
+                statusStep == ResponseStatusStep.PROCESSED && status == ResponseStatus.FAILED -> PRINT_PROVIDER_VALIDATION_FAILED
+                statusStep == ResponseStatusStep.IN_MINUS_PRODUCTION && status == ResponseStatus.FAILED -> PRINT_PROVIDER_PRODUCTION_FAILED
+                statusStep == ResponseStatusStep.DISPATCHED && status == ResponseStatus.FAILED -> PRINT_PROVIDER_DISPATCH_FAILED
+                else -> throw IllegalArgumentException("Undefined statusStep [$statusStep] and status [$status] combination")
+            }
+        }
+
```
vishalgupta-valtech, line 78: These conditions are quite confusing. I have another plan/implementation. 
My version would be this to evaluate `newStatus`. Feel free to refactor this please.

```
val newStatus = with(printResponse) {
            when (status) {
                ResponseStatus.SUCCESS -> {
                    when (statusStep) {
                        ResponseStatusStep.PROCESSED -> VALIDATED_BY_PRINT_PROVIDER
                        ResponseStatusStep.IN_MINUS_PRODUCTION -> IN_PRODUCTION
                        ResponseStatusStep.DISPATCHED -> DISPATCHED
                        else -> throw IllegalArgumentException("Undefined statusStep [$statusStep] and status [$status] combination")
                    }
                }
                ResponseStatus.FAILED -> {
                    when (statusStep) {
                        ResponseStatusStep.NOT_MINUS_DELIVERED -> NOT_DELIVERED
                        ResponseStatusStep.PROCESSED -> PRINT_PROVIDER_VALIDATION_FAILED
                        ResponseStatusStep.IN_MINUS_PRODUCTION -> PRINT_PROVIDER_PRODUCTION_FAILED
                        ResponseStatusStep.DISPATCHED -> PRINT_PROVIDER_DISPATCH_FAILED
                    }
                }
            }
        }
```

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingServiceTest.kt
```diff
@@ -0,0 +1,185 @@
...
+
+            // Then
+            verify(printDetailsRepository).getByRequestId(requestId)
+            verify(printDetailsRepository).updateItems(capture(captor))
+            assertThat(captor.value).usingRecursiveComparison().isEqualTo(listOf(expectedPrintDetails))
+        }
+    }
+
+    @Test
+    fun `should throw exception given non defined statusStep and status combination`() {
```
vishalgupta-valtech, line 165: This test can be renamed to 
`should throw exception given status is success and statusStep is not-delivered`

I cannot think of any other combination where an exception will be thrown

vishalgupta-valtech (commented):
LGTM, some comments added .

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
```diff
@@ -0,0 +1,92 @@
...
+                    it.addStatus(
+                        status = newStatus,
+                        dateCreated = OffsetDateTime.now(clock),
+                        eventDateTime = timestamp,
+                        message = null
+                    )
+
+                    if (status != SUCCESS) {
+                        it.requestId = idFactory.requestId()
+                    }
```
NathanRussellValtech, line 57: I know it's not in the story, but based on a problem we had yesterday can we also null out the batchId in this if block please (so the PrintDetails instance has a status of PENDING_ASSIGNMENT_TO_BATCH and a null batchId)
```
                    if (status != SUCCESS) {
                        it.requestId = idFactory.requestId()
                        it.batchId = null
                    }
```

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
NathanRussellValtech: Ah jinx 😁 I was just writing the exact same suggestion and the exact same code snippet 😁 Great minds think alike 😁 👍 

The other idea/approach I was thinking was a MapStruct mapper? Would that be a better/cleaner approach? And it has the added bonus that we can write a unit test for the mapping 👍 

What do you think? 🤔

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
```diff
@@ -0,0 +1,92 @@
...
+        with(printResponse) {
+            printDetails.addStatus(
+                status = newStatus,
+                dateCreated = OffsetDateTime.now(clock),
+                eventDateTime = timestamp,
+                message = message
+            )
+        }
+
+        printDetailsRepository.updateItems(listOf(printDetails))
```
NathanRussellValtech, line 90: Is there a better method on the repository (could there be?)
It seems a bit weird to call a method that is about updating many items, when we only have one 🤔

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingServiceTest.kt
```diff
@@ -0,0 +1,185 @@
...
+
+    @BeforeEach
+    fun setup() {
+        service = PrintResponseProcessingService(printDetailsRepository, idFactory, fixedClock)
+    }
+
+    @Nested
+    inner class ProcessBatchResponses {
+        @Test
+        fun `should update print details`() {
```
NathanRussellValtech, line 67: How about `should update all print requests within each batch`? on the basis that we are processing some batches, and the expected side effect is an update to the print requests ?

(Also, I'm not fussed whether you choose to use the term "print request" or "print details" in the method name - the business and the Print Provider refer to it as a "print request", but our modelling in dynamo has used the term "print details" .... not sure what the new name will be when we move over to JPA? @Neil-Massey ?)

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
NathanRussellValtech: Ah, one other thing in the case of a `FAILED` status; we also need to map the `message` field from the `BatchResponse` into the new `PrintResponseStatus` object on line 52 above

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingServiceTest.kt
```diff
@@ -0,0 +1,185 @@
...
+                listOf(printDetails1),
+                listOf(printDetails2)
+            )
+            given(idFactory.requestId()).willReturn(newRequestId)
+
+            val expectedPrintRequestStatus1 = PrintRequestStatus(
+                status = PENDING_ASSIGNMENT_TO_BATCH,
+                dateCreated = now,
+                eventDateTime = batchResponse1.timestamp,
+                message = null
```
NathanRussellValtech, line 87: This instance will have it's `message` field populated (see my comment on the implementation about processing a `FAILED` batch and setting the `message` property)

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingServiceTest.kt
NathanRussellValtech: That is true @vishalgupta-valtech 👍 

Though I think both of these tests (this one, and the main one above that does all the combinations), can be moved to a mapper test if we can implement the logic in a mapper 🤞

NathanRussellValtech (commented):
This looks really good. I think there's a couple of changes around how we process a failed batch, and possibly a new mapper?; but hopefully not too hard
Nice work 👍

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
vishalgupta-valtech: Well tbh i also thought of a mapper here.. 
so if that's easier to mapstruct it then its even better 😊 (btw i m still not there, still catching up with your great mind @NathanRussellValtech)

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
adacayi: Thanks, @fayimora , for the suggestion.
There is actually a way to do it:

<pre>
        val newStatus = with(printResponse) {
            when(statusStep to status) {
                ResponseStatusStep.PROCESSED to ResponseStatus.SUCCESS -> VALIDATED_BY_PRINT_PROVIDER
                ResponseStatusStep.IN_MINUS_PRODUCTION to ResponseStatus.SUCCESS -> IN_PRODUCTION
                ResponseStatusStep.DISPATCHED to ResponseStatus.SUCCESS -> DISPATCHED
                ResponseStatusStep.NOT_MINUS_DELIVERED to ResponseStatus.FAILED -> NOT_DELIVERED
                ResponseStatusStep.PROCESSED to ResponseStatus.FAILED -> PRINT_PROVIDER_VALIDATION_FAILED
                ResponseStatusStep.IN_MINUS_PRODUCTION to ResponseStatus.FAILED -> PRINT_PROVIDER_PRODUCTION_FAILED
                ResponseStatusStep.DISPATCHED to ResponseStatus.FAILED -> PRINT_PROVIDER_DISPATCH_FAILED
                else -> throw IllegalArgumentException("Undefined statusStep [$statusStep] and status [$status] combination")
            }
        }
</pre>

I have read the comments below from @vishalgupta-valtech and @NathanRussellValtech  and seems like it will make it more readable if we deal with success and fail statuses separately, and I agree.

i.e.
<pre>
val newStatus = with(printResponse) {
            when (status) {
                ResponseStatus.SUCCESS -> {
                    when (statusStep) {
                        ResponseStatusStep.PROCESSED -> VALIDATED_BY_PRINT_PROVIDER
                        ResponseStatusStep.IN_MINUS_PRODUCTION -> IN_PRODUCTION
                        ResponseStatusStep.DISPATCHED -> DISPATCHED
                        else -> throw IllegalArgumentException("Undefined statusStep [$statusStep] and status [$status] combination")
                    }
                }
                ResponseStatus.FAILED -> {
                    when (statusStep) {
                        ResponseStatusStep.NOT_MINUS_DELIVERED -> NOT_DELIVERED
                        ResponseStatusStep.PROCESSED -> PRINT_PROVIDER_VALIDATION_FAILED
                        ResponseStatusStep.IN_MINUS_PRODUCTION -> PRINT_PROVIDER_PRODUCTION_FAILED
                        ResponseStatusStep.DISPATCHED -> PRINT_PROVIDER_DISPATCH_FAILED
                    }
                }
            }
        }
</pre>

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
adacayi: Thanks, @vishalgupta-valtech  and @NathanRussellValtech for your suggestions.
I will introduce a mapper accordingly.

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
adacayi: Thanks, @NathanRussellValtech ,
I will address the issues.

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
adacayi: @NathanRussellValtech , this is the only update method in the dynamodb repository and since we will be switching to mysql shortly, I didn't want to introduce a new method to the repository at this point.

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingServiceTest.kt
adacayi: Thanks, for the suggestion @NathanRussellValtech ,
I will rename it accordingly. I think we will be using print requests when we move over to JPA, so happy to rename it to print request.

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingServiceTest.kt
adacayi: Thanks, @NathanRussellValtech ,
I will fix this.

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingServiceTest.kt
NathanRussellValtech: That is a fair comment; completely agree 👍 💯

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
adacayi: @NathanRussellValtech , I tried really hard to use `Mapstruct` for the mapping, but unfortunately couldn't find a way to map to an enum from two source values. Hence, the mapper is just using the suggested when implementation.

src/main/kotlin/uk/gov/dluhc/printapi/mapper/StatusMapper.kt
```diff
@@ -0,0 +1,32 @@
...
+import uk.gov.dluhc.printapi.database.entity.Status.NOT_DELIVERED
+import uk.gov.dluhc.printapi.database.entity.Status.PRINT_PROVIDER_DISPATCH_FAILED
+import uk.gov.dluhc.printapi.database.entity.Status.PRINT_PROVIDER_PRODUCTION_FAILED
+import uk.gov.dluhc.printapi.database.entity.Status.PRINT_PROVIDER_VALIDATION_FAILED
+import uk.gov.dluhc.printapi.database.entity.Status.VALIDATED_BY_PRINT_PROVIDER
+import uk.gov.dluhc.printapi.messaging.models.ProcessPrintResponseMessage.StatusStep
+import uk.gov.dluhc.printapi.database.entity.Status as StatusEntityEnum
+import uk.gov.dluhc.printapi.messaging.models.ProcessPrintResponseMessage.Status as StatusModelEnum
+
+class StatusMapper {
```
vishalgupta-valtech, line 14: I expected using `MapStruct` interface?

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
```diff
@@ -4,15 +4,18 @@ import uk.gov.dluhc.printapi.database.entity.Status.PENDING_ASSIGNMENT_TO_BATCH
...
 import uk.gov.dluhc.printapi.database.repository.PrintDetailsRepository
+import uk.gov.dluhc.printapi.mapper.StatusMapper
+import uk.gov.dluhc.printapi.messaging.models.ProcessPrintResponseMessage
 import uk.gov.dluhc.printapi.printprovider.models.BatchResponse
 import uk.gov.dluhc.printapi.printprovider.models.BatchResponse.Status.SUCCESS
 import java.time.Clock
 import java.time.OffsetDateTime
 
-class PrintResponseBatchService(
+class PrintResponseProcessingService(
```
NathanRussellValtech, line 14: Minor (and I can live with this if you or Fayi are doing it in a subsequent PR), but is it worth annotating this class with `@Service` and the mapper above with `@Component` ?

NathanRussellValtech (approved):
LGTM; thanks for making the changes 👍

src/main/kotlin/uk/gov/dluhc/printapi/mapper/StatusMapper.kt
```diff
@@ -0,0 +1,32 @@
...
+
+    fun toStatusEntityEnum(statusStep: StatusStep, statusModelEnum: StatusModelEnum): StatusEntityEnum {
+        if (statusModelEnum == StatusModelEnum.SUCCESS) {
+            return when (statusStep) {
+                StatusStep.PROCESSED -> VALIDATED_BY_PRINT_PROVIDER
+                StatusStep.IN_MINUS_PRODUCTION -> IN_PRODUCTION
+                StatusStep.DISPATCHED -> DISPATCHED
+                else -> throw IllegalArgumentException("Undefined statusStep [$statusStep] and status [SUCCESS] combination")
```
vishalgupta-valtech, line 22: Could this message be more appropriate ? 
`Print status cannot be in statusStep [`NOT_MINUS_DELIVERED`] when the status is [`SUCCESS`] `

src/main/kotlin/uk/gov/dluhc/printapi/mapper/StatusMapper.kt
NathanRussellValtech: From what @abdullah-sanver-valtech mentioned in his reply on the original comment on this, I think he tried using MapStruct and couldnt get it to work (or something?), so resorted to writing it like this
Personally I prefer this written like this vs MapStruct "magic" - I can read this code really easily and grok what its doing. MapStruct is just voodoo and wizardry to me! 🤣

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
adacayi: Thanks, @NathanRussellValtech, for the suggestions. 

I intentionally didn't include `@Service` annotation here and the `@Component` annotation in the mapper, because the current tests for those classes are just unit tests and they are not asserting the annotation usage. 
Hopefully, we will have integration tests with the follow-up subtasks which will require those annotations to be present and we will add them accordingly at that point.

NathanRussellValtech: That's cool; as long as you have it in hand 👍 😁

src/main/kotlin/uk/gov/dluhc/printapi/mapper/StatusMapper.kt
vishalgupta-valtech:
```
@Mapper
abstract class StatusMapper {

    @ValueMapping(target = "VALIDATED_BY_PRINT_PROVIDER", source = "PROCESSED")
    @ValueMapping(target = "IN_PRODUCTION", source = "IN_MINUS_PRODUCTION")
    @ValueMapping(target = "DISPATCHED", source = "DISPATCHED")
    @ValueMapping(source = MappingConstants.ANY_UNMAPPED, target = MappingConstants.THROW_EXCEPTION)
    abstract fun toStatusEntityEnumOnSuccessStatus(statusStep: StatusStep): StatusEntityEnum

    @ValueMapping(target = "NOT_DELIVERED", source = "NOT_MINUS_DELIVERED")
    @ValueMapping(target = "PRINT_PROVIDER_VALIDATION_FAILED", source = "PROCESSED")
    @ValueMapping(target = "PRINT_PROVIDER_PRODUCTION_FAILED", source = "IN_MINUS_PRODUCTION")
    @ValueMapping(target = "PRINT_PROVIDER_DISPATCH_FAILED", source = "DISPATCHED")
    abstract fun toStatusEntityEnumOnFailedStatus(statusStep: StatusStep): StatusEntityEnum
```

vishalgupta-valtech: @NathanRussellValtech @abdullah-sanver-valtech 

```
@Mapper
interface StatusMapper {

    @ValueMapping(target = "VALIDATED_BY_PRINT_PROVIDER", source = "PROCESSED")
    @ValueMapping(target = "IN_PRODUCTION", source = "IN_MINUS_PRODUCTION")
    @ValueMapping(target = "DISPATCHED", source = "DISPATCHED")
    @ValueMapping(source = MappingConstants.ANY_UNMAPPED, target = MappingConstants.THROW_EXCEPTION)
    fun toStatusEntityEnumOnSuccessStatus(statusStep: StatusStep): StatusEntityEnum

    @ValueMapping(target = "NOT_DELIVERED", source = "NOT_MINUS_DELIVERED")
    @ValueMapping(target = "PRINT_PROVIDER_VALIDATION_FAILED", source = "PROCESSED")
    @ValueMapping(target = "PRINT_PROVIDER_PRODUCTION_FAILED", source = "IN_MINUS_PRODUCTION")
    @ValueMapping(target = "PRINT_PROVIDER_DISPATCH_FAILED", source = "DISPATCHED")
    fun toStatusEntityEnumOnFailedStatus(statusStep: StatusStep): StatusEntityEnum
```

NathanRussellValtech: I dont think that does the same thing does it? We want a single mapper method that takes the `StatusStep` and the `Status`
What are the target enums eg: `PRINT_PROVIDER_VALIDATION_FAILED` - they don't exist and are not part of our design in this respect 🤔 

I think the version Abdullah has is easier to read by far 👍 💯

vishalgupta-valtech (approved):
LGTM. Thanks for addressing review comments. (I will leave the decision to use `mapstruct` or not to you @abdullah-sanver-valtech )

src/main/kotlin/uk/gov/dluhc/printapi/mapper/StatusMapper.kt
NathanRussellValtech: > What are the target enums eg: `PRINT_PROVIDER_VALIDATION_FAILED` - they don't exist and are not part of our design in this respect 🤔 

Actually they are, ignore me 😁 

I still think What Abdullah is proposing is better though 👍

PR #54, opened 2022-11-10
Feature/eip1 2386 map sqs send application to print message

Neil-Massey:


src/main/kotlin/uk/gov/dluhc/printapi/rds/mapper/RdsElectoralRegistrationOfficeMapper.kt
```diff
@@ -0,0 +1,18 @@
+package uk.gov.dluhc.printapi.rds.mapper
+
+import org.mapstruct.Mapper
+import org.mapstruct.Mapping
+import uk.gov.dluhc.printapi.dto.EroManagementApiEroDto
+import uk.gov.dluhc.printapi.rds.entity.ElectoralRegistrationOffice
+
+@Mapper
+interface RdsElectoralRegistrationOfficeMapper {
```
Neil-Massey, line 9: Prefix is so Spring Application context can start up without having duplicate bean names.

src/main/kotlin/uk/gov/dluhc/printapi/rds/mapper/CertificateMapper.kt
```diff
@@ -0,0 +1,38 @@
...
+    @Mapping(target = "vacNumber", expression = "java( idFactory.vacNumber() )")
+    @Mapping(source = "localAuthority", target = "issuingAuthority")
+    @Mapping(target = "printRequests", expression = "java( toPrintRequestList(message, ero) )")
+    abstract fun toCertificate(
+        message: SendApplicationToPrintMessage,
+        ero: EroManagementApiEroDto,
+        localAuthority: String
+    ): Certificate
+
+    protected fun toPrintRequestList(
```
Neil-Massey, line 32: Couldn't get MapStruct to do this automagically..

vishalgupta-valtech: Try this AfterMapping 

```
@Autowired
    protected lateinit var idFactory: IdFactory

    @Autowired
    protected lateinit var printRequestMapper: PrintRequestMapper

    @Mapping(target = "vacNumber", expression = "java( idFactory.vacNumber() )")
    @Mapping(source = "localAuthority", target = "issuingAuthority")
    abstract fun toCertificate(
        message: SendApplicationToPrintMessage,
        ero: EroManagementApiEroDto,
        localAuthority: String
    ): Certificate

    @AfterMapping
    protected fun addPrintRequestToCertificate(
        message: SendApplicationToPrintMessage,
        ero: EroManagementApiEroDto,
        @MappingTarget certificate: Certificate
    ) {
        certificate.addPrintRequest(printRequestMapper.toPrintRequest(message, ero))
    }
```

src/main/kotlin/uk/gov/dluhc/printapi/rds/mapper/InstantMapper.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.dluhc.printapi.rds.mapper
+
+import org.springframework.stereotype.Component
+import java.time.Instant
+import java.time.OffsetDateTime
+
+@Component
+class InstantMapper {
+    fun toInstant(offsetDateTime: OffsetDateTime?): Instant? = offsetDateTime?.toInstant()
+}
```
vishalgupta-valtech, line 10: Do these mapper belong to `rds` package?

vishalgupta-valtech: oh may be these are all rds (mysql) related changes 👍🏼 . Ignore above comment then

Neil-Massey: Just in `rds` package to keep it clear at this stage which are used for RDS... a future task will move to correct package 👍

vishalgupta-valtech (approved):
LGTM, some minor comments/suggestions

src/main/kotlin/uk/gov/dluhc/printapi/rds/mapper/CertificateMapper.kt
Neil-Massey: Thanks @vishalgupta-valtech that works, and yes it should be using `add` 👍

vishalgupta-valtech: The only difference with using aftermapping is that you can actually call `.addPrintRequest` of Certificate. However IMHO even your impl is ok but just creates a new list of print requests (doesn't add to it)

PR #56, opened 2022-11-11
EIP1-2421: Implement listener to process print responses

fayimora:


adacayi (approved):
LGTM, @fayimora. Thanks.

PR #59, opened 2022-11-14
EIP1-2262: EIP1-2422: Implement error handling for processing print responses

adacayi:


src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseFileService.kt
```diff
@@ -17,11 +19,28 @@ class PrintResponseFileService(
...
+        } catch (ex: MessagingException) {
+            if (isCauseNoSuchFile(ex)) {
+                logger.error("File $fileName was not found when trying to remove from the directory $directory")
+            } else {
+                throw ex
+            }
+        }
+    }
+
+    private fun isCauseNoSuchFile(ex: MessagingException): Boolean {
```
adacayi, line 42: Unfortunately, couldn't find a simpler way to determine if the error is caused by a not found file.

NathanRussellValtech: Yeah, you're right to point out the difficulties you had here 👍 
If this is the best way to do it then thats fine 👍 

One minor suggestion though ... checking `innerCause.id == 2` is a bit "magic number". Nobody will know what `2` means when reading this code. Could you use the static `ChannelSftp.SSH_FX_NO_SUCH_FILE` instead please?. That is the same static that is used when `SftpException` is instantiated, so it feels like the right thing to use?

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseFileService.kt
```diff
@@ -17,11 +19,28 @@ class PrintResponseFileService(
...
 
     private fun parsePrintResponseContent(printResponsesString: String): PrintResponses {
         logger.debug { "Parsing print responses $printResponsesString" }
         return objectMapper.readValue(printResponsesString, PrintResponses::class.java)
     }
+
+    private fun removeFile(directory: String, fileName: String) {
+        try {
+            sftpService.removeFileFromOutBoundDirectory(directory, fileName)
+        } catch (ex: MessagingException) {
```
NathanRussellValtech, line 33: TL;DR;
Change the method in SftpService to declare it throws `IOException`, catch `IOException` here rather than `MessagingException`, log a `warn` rather than `error`, don't re-throw the exception back.

---

Vishal had a similar challenge re: exception handling from SFTP with a ticket he was working on in this project recently; and we agreed (I convinced him 😁 ) that the fact that the Spring SFTP Template class used in `SftpService` throws a "messaging" related exception is just horrible!
It feels wrong that our service classes need to know about Spring messaging.

Have a look in `SftpService` at what Vishal did, but basically he did this:
```
    /**
     * Renames the file by suffixing ".processing" to its original name. Returns the newly renamed file name
     * Note: The "/" is appended between fileDirectoryPath and fileName before it's renamed
     * @param directory the location of the status file to be renamed
     * @param originalFileName the name of the file e.g. fileName.json
     * @throws IOException while renaming the file.
     */
    @Throws(IOException::class)
    fun markFileForProcessing(directory: String, originalFileName: String): String {
```
Therefore the client class (the thing that is calling the `SftpService` method) now only needs to catch and process `IOException`
That would be good if you could do a similar thing here.

The next thing is what to do if SftpService throws an exception that is not caused by file not found (ie. your rethrow on line 37). I'm not massively keen on that as we'll have an exception being thrown back up the stack - are we catching it and processing it anywhere up the stack?

I guess we need to consider what the context is. This is being called from `processPrintResponseFile` which is about processing a single print response file (which contains a collection of Batch Responses and a collection of Print Responses); and we are trying to remove the file as the very last step. By this point we have already processed all of the Batch Responses, and processed all of the Print Responses by putting them on a queue.

If the exception is because the file is not found (other than it being very weird because we had already read and processed it .... but I do get that it _could_ happen) ... it doesn't feel like an `error`. There's nothing we can do about it, and it actually doesn't matter because we have processed the file. It's just that something else has removed it from under us. Feels like a `warn` to me?

Hmmmm, now, in the case of it not being based on file not found ... thats something different altogether ....

Rethrowing it is probably not the thing to do though. Because `printResponseFileService.processPrintResponseFile` is called from a listener, if we rethrow the exception, the listener will throw it back, the message will not get ACK'ed, and it will get re-queued or DLQ'ed if it's failed 5 times. Neither of which feel right?
As mentioned above, by this point - irrespective of what the exception type/cause is - we have already successfully processed the file and it's contents.

I think we'd be better off by not re-throwing, and logging something in the `else` block. _Maybe_ an `error` ?, but again, there's nothing we can do about it. It's probably a `warn` ?

Does that make sense?

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
```diff
@@ -72,8 +78,22 @@ class PrintResponseProcessingService(
...
     fun processPrintResponse(printResponse: ProcessPrintResponseMessage) {
-        val newStatus = statusMapper.toStatusEntityEnum(printResponse.statusStep, printResponse.status)
-        val printDetails = printDetailsRepository.getByRequestId(printResponse.requestId)
+        val newStatus: Status
+        val printDetails: PrintDetails
+
+        try {
+            newStatus = statusMapper.toStatusEntityEnum(printResponse.statusStep, printResponse.status)
+            printDetails = printDetailsRepository.getByRequestId(printResponse.requestId)
+        } catch (ex: Exception) {
```
NathanRussellValtech, line 87: Not sure about this exception handling ... 🤔 
I'm not a fan of catching `Exception` if we can help it? It feels too broad? Is there something more specific we can catch here?
Secondly I'm not sure what we need to do in the case of an exception here. I guess its all about the context of this method and how and when it is called.

As I understand it, its called by the listener for a single print response message by its `requestId`. Its about processing 1 and only 1 print response.

If this were to fail, how might it fail, and what do we want to do?

As you have it at the moment, if the mapper fails (likely scenario ??), or it fails to find the record in the database, then we'll log an `error` and then return. Which means the SQS message will successfully ACK and will be considered complete. I guess this is right?? Or, so we want to to go straight to the DLQ do it can be investigated? (as it stands at the moment it wont DLQ so cant be investigated / re-played)

If the exception is something else (not sure what ?), then we re-throw. As per my comments about this will cause the message to be re-queued until and retried up to 5 times, by which time it hits the DLQ.
Maybe thats right? I guess perhaps for transient errors such as database gone down etc?

What do you think? 🤔 
Any thoughts @Neil-Massey ?

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseFileService.kt
adacayi: Thanks, @NathanRussellValtech , for the detailed suggestion. 

I will change the `SftpService` `removeFileFromOutBoundDirectory` method signature to throw `IOException` and will change the logging for file not found to a warning and for the other IO errors don't throw the exception, but just log it as an error.

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
Neil-Massey: From talking with @stalwartsolutions yesterday it seems that logs are not monitored but DLQ queues are.  So if we want to be aware of issues we need messages to end in the DLQ.  Messages that end in the DLQ can be replayed at a later stage - perhaps after a code or infra fix.  

Perhaps at this stage allowing all errors to be retried so either they succeed on the retry or end in the DLQ is the best approach (instead of logging the error, handling the exception and removing the message from the queue) - but as you say this is only if there is required processing still required.

adacayi: @NathanRussellValtech , @Neil-Massey , thanks for your comments. 
The only cases we log and don't throw an exception, hence the message not ending up in the DLQ, are:

- There is no printRequest for the given `requestId` in the DB
- The StatusStep is `UNDELIVERED` but step is `SUCCESS`, which is not a valid defined scenario.

In both cases, no code or infra change will fix the issues, and I don't think there is a logical way to process that `printResponse`. 
So, maybe we log for those cases and carry on as suggested in the ticket? 
Does that make sense?

src/main/kotlin/uk/gov/dluhc/printapi/service/SftpService.kt
```diff
@@ -72,9 +72,17 @@ class SftpService(
         return responseString!!
     }
 
+    @Throws(IOException::class)
     fun removeFileFromOutBoundDirectory(directory: String, fileName: String): Boolean {
         logger.info { "Removing processed file [$fileName] from directory [$directory]" }
-        return sftpOutboundTemplate.remove(createFileNamePath(directory, fileName))
+        try {
```
NathanRussellValtech, line 78: Do you need to do this? (catching the exception and either throwing as an IOException, or wrapping it in one)
Vishal didnt need to do it in his method in this class?

AFAIK the SFTP Template methods only throw MessagingException, and AFAIK MessagingException is already an IOException ??

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseFileServiceTest.kt
```diff
@@ -48,4 +53,70 @@ internal class PrintResponseFileServiceTest {
...
+        // When
+        printResponseFileService.processPrintResponseFile(directory, fileName)
+
+        // Then
+        val inOrder = inOrder(sftpService, objectMapper, printResponseProcessingService)
+        inOrder.verify(sftpService).fetchFileFromOutBoundDirectory(directory, fileName)
+        inOrder.verify(objectMapper).readValue(fileContent, PrintResponses::class.java)
+        inOrder.verify(printResponseProcessingService).processBatchAndPrintResponses(expectedPrintResponses)
+        inOrder.verify(sftpService).removeFileFromOutBoundDirectory(directory, fileName)
+        assertThat(
```
NathanRussellValtech, line 82: I think we've been round this before haven't we? Rather than writing tests that depend on log messages (feels very fragile?) wouldn't it be better to assert that `printDetailsRepository.updateItems` was not called?

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseFileServiceTest.kt
```diff
@@ -48,4 +53,70 @@ internal class PrintResponseFileServiceTest {
...
+        val directory = "EROP/Dev/OutBound"
+        val fileName = "status-20221101171156056.json"
+        val fileContent = "{\"BatchResponses\": [], \"PrintResponses\": []}"
+
+        given(sftpService.fetchFileFromOutBoundDirectory(any(), any())).willReturn(fileContent)
+        val expectedPrintResponses = PrintResponses().withBatchResponses(emptyList()).withPrintResponses(emptyList())
+        given(objectMapper.readValue(fileContent, PrintResponses::class.java))
+            .willReturn(expectedPrintResponses)
+        val exception =
+            IOException("Failed to remove file", SftpException(ChannelSftp.SSH_FX_NO_SUCH_FILE, "No such file"))
```
NathanRussellValtech, line 69: I'd be happy for this exception to be the relevant MessagingException if there is an easy/convenient constructor. It would be nice if the exception thrown in the test matched the real thing?

src/main/kotlin/uk/gov/dluhc/printapi/service/SftpService.kt
adacayi: @NathanRussellValtech, I have tried it the way Vishal did it by just adding the annotation, but it didn't work. 
Also checked if `MessagingException` is an `IOException`, but it doesn't look like it is. Hence, I went with the current implementation.
<pre>
package org.springframework.messaging;
public class MessagingException extends NestedRuntimeException 

public abstract class NestedRuntimeException extends RuntimeException
</pre>

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseFileServiceTest.kt
adacayi: @NathanRussellValtech , this is the one thrown from the `SftpService`, which we agreed to change to throw `IOException`. 

It was `MessagingException` with my initial PR, but since we made the mentioned change to the `SftpService`, I made the change here to match the corresponding behavior.

src/main/kotlin/uk/gov/dluhc/printapi/service/SftpService.kt
NathanRussellValtech: You are 100% right! and Vishal's implementation is wrong!
The SFTP Template definitely only throws MessagingException, so can you change your impl here to:
```
catch (ex: MessagingException) {
    throw IOException(ex)
}
```
The comments I left on your tests would then still stand, in that we'd need the exception to be thrown to be a MessagingException

Dont worry about Vishal's impl here (which is wrong!) - I'll fix it later and do a separate PR for it 👍

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseFileServiceTest.kt
adacayi: @NathanRussellValtech , I believe this test is testing the failure in the final step when the responses are processed (sent to queue for processing and eventually will be processed), but the response file cannot be deleted, because it is not found.
 
In this case, the data should be updated with the appropriate statuses and since we don't throw any exceptions the only difference from the happy path will be the warning log. We assert there are no exceptions thrown and I also added the assertion for the warning log.

src/main/kotlin/uk/gov/dluhc/printapi/service/SftpService.kt
adacayi: Just to confirm your suggestion @NathanRussellValtech , do you want me to change this:

<pre>
    fun removeFileFromOutBoundDirectory(directory: String, fileName: String): Boolean {
        logger.info { "Removing processed file [$fileName] from directory [$directory]" }
        try {
            return sftpOutboundTemplate.remove(createFileNamePath(directory, fileName))
        } catch (ex: Exception) {
            when (ex.cause) {
                is IOException -> throw ex.cause as IOException
                else -> throw IOException(ex)
            }
        }
    }
</pre>

to this:
<pre>
    fun removeFileFromOutBoundDirectory(directory: String, fileName: String): Boolean {
        logger.info { "Removing processed file [$fileName] from directory [$directory]" }
        try {
            return sftpOutboundTemplate.remove(createFileNamePath(directory, fileName))
        } catch (ex: MessagingException) {
            throw IOException(ex)
        }
    }
</pre>

If the answer is yes, then `SftpService` will still be throwing `IOException` and in the `PrintResponseProcessingService` we will still need to mock it to throw `IOException` instead of `MessagingException`.

Maybe I am missing something with your comment.

NathanRussellValtech: The answer is Yes; and yes, you are right about the tests! 😁 👍

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseFileServiceTest.kt
NathanRussellValtech: Ah, yes, I see
It's fine as you have it, but I would argue that in this case do we really care? If we didn't log the warning in the impl would it affect the behaviour of the class?
Let's leave it as you have it, because we need to get things merged ASAP so we can move on 👍 (but personally I don't like these tests that assert log messages ¯\_(ツ)_/¯ )

NathanRussellValtech (approved):
LGTM; thanks for making the changes 👍

PR #61, opened 2022-11-14
EIP1-2386 Integrate MySQL in process to send printer zip file or print requests

Neil-Massey:
Introduced MySQL alongside the existing DynamoDB for saving a print request to the database and then using that when processing to write printer's request files.  Later the DynamoDB code will be removed once the response processing is updated to use the MySQL database.

Note that the MySQL processing is being added alongside the existing DynamoDB at this stage (response file processing still needs updating - in next PR).  At this stage:

- 1 application is saved in both DynamoDB and MySQL
- Create separate batches for Dynamo and MySQL and raise SQS for all batches
- Processing has 1 listener that skips creating zips for MySQL but updates its status

Next steps:
Update response file processing 
Switch off DynamoDB

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/database/repository/PrintDetailsRepository.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/database/repository/PrintDetailsRepository.kt
@@ -60,12 +60,8 @@ class PrintDetailsRepository(client: DynamoDbEnhancedClient, tableConfig: Dynamo
     fun getAllByStatus(status: Status): List<PrintDetails> {
         val expression = Expression.builder()
             .expression("#print_status = :status")
-            .expressionNames(mapOf(Pair("#print_status", "status")))
-            .expressionValues(
-                mapOf(
-                    Pair(":status", AttributeValue.builder().s(status.toString()).build())
-                )
-            )
+            .expressionNames(mapOf("#print_status" to "status"))
+            .expressionValues(mapOf(":status" to AttributeValue.builder().s(status.toString()).build()))
             .build()
         return table.scan(ScanEnhancedRequest.builder().filterExpression(expression).build()).items().toList()
     }
--- a/src/main/kotlin/uk/gov/dluhc/printapi/database/repository/PrintDetailsRepository.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/database/repository/PrintDetailsRepository.kt
@@ -63,7 +63,7 @@ class PrintDetailsRepository(client: DynamoDbEnhancedClient, tableConfig: Dynamo
             .expressionNames(mapOf(Pair("#print_status", "status")))
             .expressionValues(
                 mapOf(
-                    Pair(":status", AttributeValue.builder().s(Status.PENDING_ASSIGNMENT_TO_BATCH.toString()).build())
+                    Pair(":status", AttributeValue.builder().s(status.toString()).build())
                 )
             )
             .build()
--- a/src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Certificate.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Certificate.kt
@@ -24,7 +24,6 @@ import javax.persistence.GenerationType
 import javax.persistence.Id
 import javax.persistence.JoinColumn
 import javax.persistence.OneToMany
-import javax.persistence.PrePersist
 import javax.persistence.Table
 import javax.persistence.Version
 import javax.validation.constraints.NotNull
@@ -94,28 +93,39 @@ class Certificate(
 
     fun addPrintRequest(newPrintRequest: PrintRequest): Certificate {
         printRequests += newPrintRequest
+        assignStatus()
         return this
     }
 
-    @PrePersist
-    fun prePersist() {
+    fun getCurrentPrintRequest(): PrintRequest {
+        printRequests.sortByDescending { it.requestDateTime }
+        return printRequests.first()
+    }
+
+    /**
+     * Adds the new status to the current PrintRequest and updates the current Certificate status.
+     */
+    fun addStatus(
+        status: Status,
+        eventDateTime: Instant = Instant.now(),
+        message: String? = null
+    ) {
+        val currentPrintRequest = getCurrentPrintRequest()
+        currentPrintRequest.addPrintRequestStatus(
+            PrintRequestStatus(
+                status = status,
+                eventDateTime = eventDateTime,
+                message = message
+            )
+        )
         assignStatus()
     }
 
-    fun getCurrentPrintRequest(): PrintRequest? {
-        printRequests.sortByDescending { it.requestDateTime }
-        return printRequests.firstOrNull()
-    }
-
     private fun assignStatus() {
-        status = if (printRequests.isEmpty()) {
-            Status.PENDING_ASSIGNMENT_TO_BATCH
-        } else {
-            val currentPrintRequest = getCurrentPrintRequest()!!
-            currentPrintRequest.statusHistory.sortByDescending { it.eventDateTime }
-            val currentStatus = currentPrintRequest.statusHistory.first()
-            currentStatus.status
-        }
+        val currentPrintRequest = getCurrentPrintRequest()
+        currentPrintRequest.statusHistory.sortByDescending { it.eventDateTime }
+        val currentStatus = currentPrintRequest.statusHistory.first()
+        status = currentStatus.status
     }
 
     override fun equals(other: Any?): Boolean {
--- a/src/main/kotlin/uk/gov/dluhc/printapi/rds/repository/CertificateRepository.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/rds/repository/CertificateRepository.kt
@@ -9,9 +9,9 @@ import java.util.UUID
 @Repository
 interface CertificateRepository : JpaRepository<Certificate, UUID> {
 
-    fun getByPrintRequestsRequestIdIs(requestId: String): Certificate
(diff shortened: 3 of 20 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Certificate.kt
```diff
@@ -94,28 +93,39 @@ class Certificate(
 
     fun addPrintRequest(newPrintRequest: PrintRequest): Certificate {
         printRequests += newPrintRequest
+        assignStatus()
         return this
     }
 
-    @PrePersist
```
Neil-Massey, line 100: Refactored as the `@PreUpdate` was not invoked reliably as the changes to the associated print request and print request statuses does not cause any changes in the Certificate for the hook to be called that set the correct status.  Status is now managed in directly when print requests are added to the certificate or a new status is added to the current print request through the Certificate addStatus method.

src/test/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintRequestBatchMessageListenerIntegrationTest.kt
```diff
@@ -54,28 +62,118 @@ internal class ProcessPrintRequestBatchMessageListenerIntegrationTest : Integrat
...
+                PrintRequestStatus(ASSIGNED_TO_BATCH, OffsetDateTime.now(clock), OffsetDateTime.now(clock))
             ),
             requestId = requestId,
             eroWelsh = buildElectoralRegistrationOffice(),
             photoLocation = "arn:aws:s3:::$s3Bucket/$s3Path"
         )
         printDetailsRepository.save(details)
+
+        // save certificates in MySQL
+        var certificate = Certificate(
```
Neil-Massey, line 74: Verified manually that the only difference in the generated File Contents is difference in timestamp precision.
This test will be simplified after DynamoDB code is removed.

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintRequestsServiceTest.kt
```diff
@@ -39,31 +48,44 @@ class PrintRequestsServiceTest {
...
         // Then
-        verify(printDetailsRepository).getAllByStatus(Status.PENDING_ASSIGNMENT_TO_BATCH)
-        verify(idFactory, times(3)).batchId()
+        verify(printDetailsRepository).getAllByStatus(PENDING_ASSIGNMENT_TO_BATCH)
+        verify(certificateRepository).findByStatusIs(PENDING_ASSIGNMENT_TO_BATCH)
+        verify(idFactory, times(3 * 2)).batchId()
         verify(printDetailsRepository, times(12)).save(any())
-        verify(processPrintRequestQueue, times(3)).submit(any())
+        verify(certificateRepository, times(12)).save(any())
+        verify(processPrintRequestQueue, times(3 * 2)).submit(any())
```
Neil-Massey, line 82: Done as `*2` just to highlight which will need the `*2` removed when Dynamo is removed

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/PrintDetailsRepository.kt
```diff
@@ -63,7 +63,7 @@ class PrintDetailsRepository(client: DynamoDbEnhancedClient, tableConfig: Dynamo
             .expressionNames(mapOf(Pair("#print_status", "status")))
             .expressionValues(
                 mapOf(
-                    Pair(":status", AttributeValue.builder().s(Status.PENDING_ASSIGNMENT_TO_BATCH.toString()).build())
+                    Pair(":status", AttributeValue.builder().s(status.toString()).build())
                 )
```
NathanRussellValtech, line 67: Minor (and it doesn't matter because this class will be deleted soon, but you can do this neater without a `Pair` like this:
```
                mapOf(
                    ":status" to AttributeValue.builder().s(status.toString()).build()
                )
```

src/main/kotlin/uk/gov/dluhc/printapi/rds/entity/Certificate.kt
```diff
@@ -94,28 +93,39 @@ class Certificate(
...
-    fun prePersist() {
-        assignStatus()
+    fun getCurrentPrintRequest(): PrintRequest {
+        printRequests.sortByDescending { it.requestDateTime }
```
NathanRussellValtech, line 101: Does `sortByDescending` affect (mutate) the collection, or does it return a new collection?
I think you might have to do this instead:
```
fun getCurrentPrintRequest(): PrintRequest =
        printRequests.sortByDescending { it.requestDateTime }.first()
```

src/main/kotlin/uk/gov/dluhc/printapi/rds/repository/CertificateRepository.kt
```diff
@@ -2,8 +2,16 @@ package uk.gov.dluhc.printapi.rds.repository
...
 
 @Repository
-interface CertificateRepository : JpaRepository<Certificate, UUID>
+interface CertificateRepository : JpaRepository<Certificate, UUID> {
+
+    fun getByPrintRequestsRequestIdIs(requestId: String): Certificate
+
+    fun findByStatusAndPrintRequestsBatchIdIs(certificateStatus: Status, batchId: String): List<Certificate>
+
+    fun findByStatusIs(certificateStatus: Status): List<Certificate>
```
NathanRussellValtech, line 16: Minor naming .... method names suffixed with `Is` ? Do we do that elsewhere? 🤔 
Would `findByStatus` etc be more in keeping with what we do elsewhere?

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintRequestsService.kt
```diff
@@ -7,17 +7,47 @@ import uk.gov.dluhc.printapi.database.entity.Status
...
     private val processPrintRequestQueue: MessageQueue<ProcessPrintRequestBatchMessage>,
 ) {
 
     fun processPrintRequests(batchSize: Int) {
+        processDynamoBatch(batchSize)
+        processBatch(batchSize)
+    }
+
+    private fun processBatch(batchSize: Int) {
+        batchCertificates(batchSize).forEach { (batchId, batchOfCertificates) ->
```
NathanRussellValtech, line 29: I have 2 questions here, though I know they are contradictory to each other! 🤣 

1. Should processBatch and processDynamoBatch work from the same call to batchCertificates? 
At the moment they both call it independently , so possibility for there to be 2 different result sets?
2. Should processBatch (which is the one that uses JPA repositories) call a JPA version of batchCertificates?

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintRequestsServiceTest.kt
```diff
@@ -86,15 +117,22 @@ class PrintRequestsServiceTest {
...
-        verify(processPrintRequestQueue).submit(expectedMessage)
+        verify(printDetailsRepository).getAllByStatus(PENDING_ASSIGNMENT_TO_BATCH)
+        verify(certificateRepository).findByStatusIs(PENDING_ASSIGNMENT_TO_BATCH)
+        verify(idFactory, times(2)).batchId()
+        verify(processPrintRequestQueue, times(2)).submit(expectedMessage)
 
         verify(printDetailsRepository).save(capture(savedRecordArgumentCaptor))
         val savedPrintDetails = savedRecordArgumentCaptor.value!!
         assertThat(savedPrintDetails.printRequestStatuses!!.sortedBy { it.eventDateTime }.map { it.status })
-            .containsExactly(Status.PENDING_ASSIGNMENT_TO_BATCH, Status.ASSIGNED_TO_BATCH)
```
NathanRussellValtech, line 96: Did you only mean to do the static import on `PENDING_ASSIGNMENT_TO_BATCH` or is it worth doing `ASSIGNED_TO_BATCH` as well?

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintServiceTest.kt
```diff
@@ -47,4 +56,23 @@ class PrintServiceTest {
         verify(printDetailsMapper).toPrintDetails(message, ero, localAuthority.name)
         verify(printDetailsRepository).save(printDetails)
     }
+
+    @Test
```
NathanRussellValtech, line 60: Separate tests - I like it. Will make it easier to pull out the Dynamo stuff shortly 👍

NathanRussellValtech (approved):
LGTM; a couple of minor comments; plus the one on the repository class where the method names are suffixed `Is` ?

src/main/kotlin/uk/gov/dluhc/printapi/rds/repository/CertificateRepository.kt
Neil-Massey: I copied from `VoterCardApplicationRepository` but agree it reads better without the `Is` so have removed.

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintRequestsServiceTest.kt
Neil-Massey: Imported

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintServiceTest.kt
Neil-Massey: Yeah, where possible done this - but in many cases the existing tests failed with additional code so those tests were updated and have a mix of MySQL and DynamoDB mocks and verifies.

PR #63, opened 2022-11-14
EIP1-2533 - Map static english and welsh ERO contact details

NathanRussellValtech:
This PR maps contact details for the ERO (english) and optionally Welsh into the `PrintDetails` dynamo entity and `PrintDetails` RDS entity

We need this so that when the Print Request is mapped from the database to the PSV in the zipped batch, then we have fields to send in the PSV (they are mandatory fields, and empty string (`""`) which we were sending before is classed by the Print Provider as a non/null value :( )

As and when ERO Contact and Address details come back from the REST API from `ero-management` then we will have the real data, and the area of change will be in `EroManagementApiEroDtoMapper` and a few tests 🤞


src/main/kotlin/uk/gov/dluhc/printapi/dto/EroManagementApiEroDto.kt
```diff
@@ -3,5 +3,8 @@ package uk.gov.dluhc.printapi.dto
 data class EroManagementApiEroDto(
     val id: String,
     val name: String,
-    val localAuthorities: List<EroManagementApiLocalAuthorityDto>
+    val localAuthorities: List<EroManagementApiLocalAuthorityDto>,
+
+    val englishContactDetails: EroContactDetailsDto,
```
NathanRussellValtech, line 8: This class _may_ change as and when we have changed `ero-management-api` to return ERO Contact and Address data. Until we have designed the response data for that API this will do for now

src/main/kotlin/uk/gov/dluhc/printapi/mapper/EroManagementApiEroDtoMapper.kt
```diff
@@ -1,10 +1,48 @@
...
+                postcode = "LL55 1SH",
+            )
+        )
+
+    // TODO - to be replaced with mapping from ElectoralRegistrationOfficeResponse when ERO Management returns contact data
+    protected fun welshContactDetails(): EroContactDetailsDto =
+        EroContactDetailsDto(
+            name = "Etholiadau Cyngor Gwynedd",
+            phoneNumber = "01766 771000",
+            website = "https://www.gwynedd.llyw.cymru/cy/Cyngor/Cysylltu-%c3%a2-ni/Cysylltu-%c3%a2-ni.aspx",
```
NathanRussellValtech, line 38: These are the Welsh details for Gwynedd Council, including accented characters in their website URL and street 👍

src/main/kotlin/uk/gov/dluhc/printapi/mapper/ElectoralRegistrationOfficeMapper.kt
```diff
@@ -1,18 +1,11 @@
...
 interface ElectoralRegistrationOfficeMapper {
 
-    @Mapping(source = "name", target = "name")
-    @Mapping(target = "phoneNumber", constant = "")
-    @Mapping(target = "emailAddress", constant = "")
-    @Mapping(target = "website", constant = "")
-    @Mapping(target = "address.street", constant = "")
-    @Mapping(target = "address.postcode", constant = "")
-    fun map(dto: EroManagementApiEroDto): ElectoralRegistrationOffice
+    fun map(dto: EroContactDetailsDto): ElectoralRegistrationOffice
```
adacayi, line 10: @NathanRussellValtech , seems like we can get rid of this mapper class. 
The `PrintDetailsMapper` will be able map the ero contact details without the help of this class, since there are no exception cases, just mapping fields by name convention.

adacayi (approved):
LGTM, @NathanRussellValtech . 
Added one suggestion to simply code.

src/main/kotlin/uk/gov/dluhc/printapi/dto/AddressDto.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.dluhc.printapi.dto
+
+data class AddressDto(
+    var street: String,
+    var postcode: String,
+    var `property`: String? = null,
```
Neil-Massey, line 6: Probably don't need the backticks on property (I think it's related to DynamoDB elsewhere)

src/main/kotlin/uk/gov/dluhc/printapi/mapper/EroManagementApiEroDtoMapper.kt
Neil-Massey: Interesting, so we'd expect the URL to be stored and provided with URL encoding - not really a question for this PR - just pondering aloud.

Neil-Massey (approved):
LGTM - even though will conflict with my PR changes

src/main/kotlin/uk/gov/dluhc/printapi/mapper/EroManagementApiEroDtoMapper.kt
NathanRussellValtech: Yeah, I don't know TBH. It might be that they only provide the link to the "home page" , which in this case would be `https://www.gwynedd.llyw.cymru/cy/Cyngor` ?
I've used a link here however to a page on their site (their contact us page) as it has "special" characters in the path.
I'm assuming here that they copy and paste the URL encoded version (which is what I did), but the version that appears in the browser bar has the "special" characters in it:
Screenshot 2022-11-15 at 09 31 49

src/main/kotlin/uk/gov/dluhc/printapi/mapper/ElectoralRegistrationOfficeMapper.kt
NathanRussellValtech: Good idea; done (it also removed a couple of tests with it 👍 )

PR #64, opened 2022-11-15
EIP1-2386 Process batch responses file with MySQL

Neil-Massey:
Adds processing of response file using the MySQL database records.

Subsequent PR will remove the Dynamo interactions.

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
@@ -1,6 +1,7 @@
 package uk.gov.dluhc.printapi.service
 
 import org.springframework.stereotype.Service
+import uk.gov.dluhc.printapi.database.entity.Status
 import uk.gov.dluhc.printapi.database.entity.Status.PENDING_ASSIGNMENT_TO_BATCH
 import uk.gov.dluhc.printapi.database.entity.Status.RECEIVED_BY_PRINT_PROVIDER
 import uk.gov.dluhc.printapi.database.entity.Status.SENT_TO_PRINT_PROVIDER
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingServiceTest.kt
```diff
@@ -166,6 +174,80 @@ class PrintResponseProcessingServiceTest {
             assertThat(captor.allValues[1]).usingRecursiveComparison().isEqualTo(listOf(expectedPrintDetails2))
             verify(idFactory).requestId()
         }
+
+        @Test
+        fun `should update all certificates within each batch`() {
```
Neil-Massey, line 179: Mirrored existing test, but would probably prefer this as separate tests for `failed batch` and `success batch` and possibly `multiple batches`

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingServiceTest.kt
```diff
@@ -166,6 +174,80 @@ class PrintResponseProcessingServiceTest {
...
+            service.processBatchResponses(listOf(batchResponse1, batchResponse2))
+
+            // Then
+            verify(certificateRepository).findByStatusAndPrintRequestsBatchId(SENT_TO_PRINT_PROVIDER, batchResponse1.batchId)
+            verify(certificateRepository).findByStatusAndPrintRequestsBatchId(SENT_TO_PRINT_PROVIDER, batchResponse2.batchId)
+            verify(certificateRepository).saveAll(listOf(certificate1))
+            verify(certificateRepository).saveAll(listOf(certificate2))
+            assertThat(certificate1.getCurrentPrintRequest().requestId).isEqualTo(newRequestId)
+            assertThat(certificate1.getCurrentPrintRequest().batchId).isNull()
+            assertThat(
```
Neil-Massey, line 227: As this is not a `data` class no option to use the `copy` approach used in the previous DynamoDb tests.

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
```diff
@@ -71,8 +83,66 @@ class PrintResponseProcessingService(
...
+                if (status != SUCCESS) {
+                    val currentPrintRequest = it.getCurrentPrintRequest()
+                    currentPrintRequest.requestId = idFactory.requestId()
+                    currentPrintRequest.batchId = null
+                }
+            }
+        }
+    }
+
     fun processPrintResponse(printResponse: ProcessPrintResponseMessage) {
```
NathanRussellValtech, line 119: Does this method need to be `@Transactional`?

NathanRussellValtech (approved):
LGTM; just one minor comment about whether the method should be `@Transactional` ?
Other than that 👍

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
Neil-Massey: Yes. definitely - thanks for spotting 👍

PR #65, opened 2022-11-15
EIP1-2386 Set vacVersion and certificateFormat constants

Neil-Massey:


src/main/kotlin/uk/gov/dluhc/printapi/rds/mapper/PrintRequestMapper.kt
```diff
@@ -23,6 +23,8 @@ abstract class PrintRequestMapper {
     protected lateinit var clock: Clock
 
     @Mapping(target = "id", ignore = true)
+    @Mapping(target = "vacVersion", constant = "1")
```
Neil-Massey, line 26: @NathanRussellValtech - Are these constants as expected?

NathanRussellValtech: Perfect 👍

NathanRussellValtech (approved):
LGTM 👍

PR #70, opened 2022-11-15
EIP1-2386 Switch flows to use MySQL

Neil-Massey:


NathanRussellValtech (approved):
The diff is too hard to read, but I'm happy to assume it's good to go (we can always change it moving forwards). Lets get this merged and hope that it fixes the `dev` env in respect of batching, which will unblock Vijay & co. 🤞

PR #72, opened 2022-11-16
EIP1-2386 Refactor to flush database updates before SQS messages are sent.

Neil-Massey:
DynamoDB updates were being done inside the method and then SQS messages sent, with MySQL and the `@Transaction` annotation the semantics had changed so SQS messages were sent before the database updates were made.  Refactored for database updates to happen before SQS messages are sent.


NathanRussellValtech (approved):
I dont fully grok it, but if you're happy I'm happy 😁 
We just need to get this done and working!

PR #73, opened 2022-11-16
EIP1-1878 - Logging in scheduled services; adjusted cron times in dev

NathanRussellValtech:
This PR changes the cron times in `dev` for both cron jobs to every 10 minutes
Plus removes the logging from the `ProcessPrintResponsesBatchJob` (because the service method that it calls also logs, so this was pure noise)
I've also added logging to the service class that the batch cron job calls so that we can see when it is invoked etc.


Neil-Massey (approved):
LGTM - minor comment

PR #74, opened 2022-11-16
EIP1-2386 Remove DynamoDb from project

Neil-Massey:
Removes dynamo and updates all affected tests


src/test/kotlin/uk/gov/dluhc/printapi/messaging/SendApplicationToPrintMessageListenerIntegrationTest.kt
```diff
@@ -167,12 +155,8 @@ internal class SendApplicationToPrintMessageListenerIntegrationTest : Integratio
         sqsMessagingTemplate.convertAndSend(sendApplicationToPrintQueueName, payload)
 
         // Then
-        await.during(5, SECONDS).until {
-            val response = dynamoDbClient.scan(
-                ScanRequest.builder().tableName(dynamoDbConfiguration.printDetailsTableName).build()
-            )
-            assertThat(response.items()).isEmpty()
-            true
+        await.atMost(5, SECONDS).untilAsserted {
```
Neil-Massey, line 158: Found that checking the repo would cause test always to pass even if the validation was skipped as the test doesn't set up all the other dependencies for it to succeed.  Also there is a race condition as the check is same as the initial state.

Found we can positively test for an error message that is directly related to the validation failure.

src/test/kotlin/uk/gov/dluhc/printapi/messaging/SendApplicationToPrintMessageListenerIntegrationTest.kt
```diff
@@ -115,49 +117,35 @@ internal class SendApplicationToPrintMessageListenerIntegrationTest : Integratio
...
-                    "printRequests.delivery.address.version",
-                    "printRequests.delivery.address.dateCreated",
-                    "printRequests.delivery.address.createdBy",
-                    "printRequests.statusHistory.id",
-                    "printRequests.statusHistory.version",
-                    "printRequests.statusHistory.dateCreated",
-                    "printRequests.statusHistory.createdBy",
                     "printRequests.statusHistory.eventDateTime"
                 )
+                .ignoringFieldsMatchingRegexes(
```
Neil-Massey, line 129: Much more satisfying than listing all the names

src/test/kotlin/uk/gov/dluhc/printapi/messaging/SendApplicationToPrintMessageListenerIntegrationTest.kt
```diff
@@ -115,49 +117,35 @@ internal class SendApplicationToPrintMessageListenerIntegrationTest : Integratio
             val response = certificateRepository.findAll()
             assertThat(response).hasSize(1)
             val saved = response[0]
+            val instantEqualToRoundedSeconds: BiPredicate<Instant, Instant> =
```
Neil-Massey, line 120: Value returned from the database is only to second precision - not the millis that is in the expected object.

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/testdata/rds/CertificateBuilder.kt
```diff
@@ -101,26 +101,35 @@ fun printRequestStatusBuilder(
...
 }
 
-fun rdsAddressBuilder(): Address {
-    return Address(
-        street = aValidAddressStreet(),
-        postcode = aValidAddressPostcode()
-    )
-}
+fun rdsAddressBuilder(
+    street: String = DataFaker.faker.address().streetName(),
```
Neil-Massey, line 117: Now using this one in other tests it needs to be fully populated.

RDS prefix will be removed and packaging restructured in separate PR.

PR #75, opened 2022-11-16
EIP1-2386 Persist print request batches before sending SQS messages and reject empty batches

Neil-Massey:
Batches need to be saved to database before SQS messages related to the batches are queued.  Attempting to process a print batch containing no certificates should result in an error so SQS message is retried in a race-condition.


NathanRussellValtech (approved):
LGTM 👍

PR #77, opened 2022-11-16
EIP1-2386 Remove DynamoDB related env props

Neil-Massey:


adacayi (approved):
LGTM, @Neil-Massey. 
Thanks.

PR #78, opened 2022-11-17
EIP1-1990: EIP1-2592: Add open api specifications for get certificate by application id endpoint

adacayi:


src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -44,7 +44,164 @@ paths:
...
+  '/eros/{eroId}/applications/{applicationId}':
+    parameters:
+      - name: eroId
+        description: The ID of the Electoral Registration Office responsible for the application.
+        schema:
+          type: string
+        in: path
+        required: true
+      - name: applicationId
+        description: The identifier of the Voter Authority Certificate to retrieve.
```
Neil-Massey, line 59: I think this should be same as in the VCA spec:
```
      - name: applicationId
        description: The identifier of the Voter Card Application to retrieve.
```

The Certificate has a VAC number that is more like the identifier of the certificate.

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -44,7 +44,164 @@ paths:
...
+                {}
+    get:
+      summary: Returns a single certificate
+      description: Returns a single certificate
+      tags:
+        - Single Certificate
+      responses:
+        '200':
+          $ref: '#/components/responses/Certificate'
+        '404':
```
Neil-Massey, line 109: Do we both with `404`?  (We don't include the `401` and `403` responses)

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -44,7 +44,164 @@ paths:
...
+        connectionType: VPC_LINK
+        connectionId: '${vpc_connection_id}'
+        httpMethod: GET
+
 components:
+  #
+  # Schema and Enum Definitions
+  # --------------------------------------------------------------------------------
+  schemas:
+    CertificateResponse:
```
Neil-Massey, line 133: I was wondering if this was just a summary - the print request that is included seems to be a brief summary - so don't know if we will have in the future objects that include full details.

src/main/resources/openapi/PrintAPIs.yaml
adacayi: @Neil-Massey , makes sense. I will change accordingly.

src/main/resources/openapi/PrintAPIs.yaml
adacayi: I looked in to the VCA specs and saw 404 was included.
I can remove it, if that's something we agreed and not reflected in VCA specs.

src/main/resources/openapi/PrintAPIs.yaml
adacayi: Makes sense to change it to summary, if there will be an endpoint in the future with full details. @NathanRussellValtech , what do you think?

src/main/resources/openapi/PrintAPIs.yaml
NathanRussellValtech: TL;DR; I don't care much ¯\_(ツ)_/¯ 😁 

Personally I think we should leave it there. I think it's more helpful to a consumer looking at the spec in a visual editor like swagger editor, than it not being there.
I take the point about 401 and 403, but 404 seems like a more "normal" response code you might want to know about as a consumer of the API? ¯\_(ツ)_/¯

src/main/resources/openapi/PrintAPIs.yaml
NathanRussellValtech: Agreed. I'm not 100% what our future API requirements will be. We know we will have a different API soon that allows to search for certs and their print requests. Whether this is the same response object type (a collection of them), I don't know at this point.

What I do think though is that what we are returning here is not the full Certificate with all of it's print responses etc. Whether we'll ever have a need to? I honestly don't know, but I think @Neil-Massey is right in that this is more like a summary that in turn contains a list of summaries of the print requests (that for the most part is the last status of each print request)

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -44,7 +44,164 @@ paths:
...
+          example: DEV1W0XDH1368LWE1J40
+        printRequests:
+          type: array
+          description: List of print requests for the certificate.
+          items:
+            $ref: '#/components/schemas/PrintRequest'
+      required:
+        - vacNumber
+        - printRequests
+    PrintRequest:
```
NathanRussellValtech, line 151: As above, is this a full Print Request object, or some kind of summary?
If we choose to rename the object here, then perhaps we also need to rename the property in `Certificate` (or whatever its new name is 😁 ) in a similar way ?

Also, I wonder if the description of the fields in this object need to be in the past tense? `UserId of the user who made the print request`, `The timestamp in ISO-8601 UTC that the print request status update was received` etc
Otherwise someone might look at the schemas in this spec, find this, and assume we can create a new PrintRequest somehow with some API endpoint that uses this object?

What do you think?

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -44,7 +44,164 @@ paths:
...
+          description: The timestamp in ISO-8601 UTC that the print request status update is received
+          example: '2022-11-17T10:53:25Z'
+        message:
+          type: string
+          description: Message received for the print request status update
+      required:
+        - status
+        - userId
+        - dateTime
+    Status:
```
NathanRussellValtech, line 175: I'm thinking about the generated classes here .... I wonder if this enum should be called `PrintRequestStatus` or similar? Finding a model class simply called `Status` might be a bit vague?
(happy for the property in `PrintRequest` to still be called `status` - its just the class itself that I think might benefit from a more specific name?)

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -44,7 +44,164 @@ paths:
...
+        - print-processing
+        - print-failed
+        - dispatched
+        - not-delivered
+
+  #
+  # Response Body Definitions
+  # --------------------------------------------------------------------------------
+  responses:
+    Certificate:
```
NathanRussellValtech, line 188: As above re: `Certificate` .... whatever you choose to name the main `Certificate` schema object, we should probably follow suit here too 👍

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -44,7 +44,164 @@ paths:
         connectionId: '${vpc_connection_id}'
         httpMethod: GET
 
+  #
+  # Single Certificate
+  # --------------------------------------------------------------------------------
+  '/eros/{eroId}/applications/{applicationId}':
```
NathanRussellValtech, line 50: I'm not sure about this path ... it looks a lot like getApplicationById from VCA 😜 
Are we getting an application? or are we getting a certificate? If we are getting a certificate is it a full certificate or a certificate summary in some way? Perhaps that differentiator (cert vs. cert summary) doesnt matter so much in the path?, but I think the path should be different from getApplicationById .... 🤔 

What do you think?

src/main/resources/openapi/PrintAPIs.yaml
adacayi: @NathanRussellValtech , @Neil-Massey , thanks for the suggestion. 
I have renamed Certificate to CertificateSummary accordingly.

src/main/resources/openapi/PrintAPIs.yaml
adacayi: Thanks, @NathanRussellValtech, for the suggestion.
 
I agree that `Status` might be a bit vague. 
I renamed it to `PrintRequestStatus` as you suggested.

src/main/resources/openapi/PrintAPIs.yaml
adacayi: @NathanRussellValtech , I wasn't sure actually for the url and just followed the way we did it in the VCA where the path does not refer to the voter card application itself and we assume it would return a voter card application. In here I thought we might also infer this would return a certificate (or certificate summary), hence the implementation.

After your suggestion I changed it to `certificates/eros/{eroId}/applications/{applicationId}`. 

Does this make sense?

src/main/resources/openapi/PrintAPIs.yaml
adacayi: Thanks, for the suggestions, @NathanRussellValtech ,
I agree for the summary renaming and we can make the descriptions past sense to avoid any confusions.
I pushed the changes accordingly.

src/main/resources/openapi/PrintAPIs.yaml
Neil-Massey: I raised as I only saw one `404` documented in the `VoterCardApplicationAPIs.yaml` but would have expected all the ones with path parameters could return a not found so thought we didn't generally bother.

I personally would detail all expected response codes (appreciate that almost all our endpoints return the 401/403 so would be repetitious and not necessary).

Neil-Massey (approved):
LGTM - The Path could be `certificates/eros/{eroId}/vca/{applicationId}`

src/main/resources/openapi/PrintAPIs.yaml
adacayi: After discussions, this is the final agreed endpoint: `/eros/{eroId}/certificates/applications/{applicationId}`

adacayi:
> certificates/eros/{eroId}/vca/{applicationId}



> LGTM - The Path could be `certificates/eros/{eroId}/vca/{applicationId}`



> LGTM - The Path could be `certificates/eros/{eroId}/vca/{applicationId}`

Thanks, @Neil-Massey. 
As agreed I have changed the endpoint to: ` /eros/{eroId}/certificates/applications/{applicationId}`

PR #79, opened 2022-11-17
EIP1-1878 - fix cardVersion length; set to `A`

NathanRussellValtech:
This PR fixes a couple of bugs that have been identified today as part of the Print Provider integration testing:

1. The field length for `cardVersion` in the `PrintRequest` spec was wrong. It should not have field length constraints
2. The value for the field `cardVersion` in the PSV was `1`. The Print Providers have asked that it is `A`

### DIFF
--- a/src/main/resources/yamlschema/PrintRequest.yaml
+++ b/src/main/resources/yamlschema/PrintRequest.yaml
@@ -51,7 +51,7 @@ properties:
     example: Smith
   cardVersion:
     type: string
-    example: 1
+    example: A
     description: Supplier version of certificate
   cardNumber:
     type: string


adacayi (approved):

src/main/kotlin/uk/gov/dluhc/printapi/mapper/PrintRequestMapper.kt
```diff
@@ -23,7 +23,7 @@ abstract class PrintRequestMapper {
     protected lateinit var clock: Clock
 
     @Mapping(target = "id", ignore = true)
-    @Mapping(target = "vacVersion", constant = "1")
+    @Mapping(target = "vacVersion", constant = "A")
```
Neil-Massey, line 26: Wow - never saw that coming!

src/main/resources/yamlschema/PrintRequest.yaml
```diff
@@ -84,7 +82,6 @@ properties:
     description: Future proof in case of fast track
   photo:
     type: string
-    format: binary
```
Neil-Massey, line 87: Doh

PR #81, opened 2022-11-18
[DRAFT] POC for adding correlation id to logging

NathanRussellValtech:
Draft PR for discussion

This is a POC that aims to add a correlation ID to all log messages using an MCD variable, Interceptors and some AOP

I'm trying to think about this in a generic way so that it could be added to all of our apps, and it would "just work" .... 🤔 

So, my thinking is that there are 3 "starting" points for our apps:
* an incoming HTTP request ie. a REST API
* a consumed (received) SQS message
* a scheduled task (`@Scheduler`)

for each of these 3 "starting" points we can use an Interceptor (http requests) or AOP (SQS and scheduled tasks) to generate a correlation ID and add it to MDC. We can then include the MDC variable in the log pattern. Simple stuff

But, we also want to be able to pass around / maintain a correlation ID across threads, and even across apps.

So, the Interceptor (http requests) and AOP (SQS) need to look at a header on the inbound request (HTTP request or SQS Message), and if there is a header `x-correlation-id` then set that value in the MDC, else set a new correlation ID in the MDC
We don't need to do this on the AOP aspect for scheduled tasks because nothing "calls" it - there is nothing upstream of it that would have set a correlation ID that we need to maintain

Next we need some way of passing the existing correlation ID from MDC into a header `x-correlation-id` when making outbound HTTP requests or producing (sending) SQS messages.
For SQS we can do that with more AOP (though I haven't actually got that working yet!)
For outbound HTTP requests I imagine we'd use another interceptor (though I've not written that yet)

Thats the rough idea anyway;
Thoughts?

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
@@ -1,10 +1,13 @@
 package uk.gov.dluhc.printapi.config
 
 import org.aspectj.lang.JoinPoint
+import org.aspectj.lang.ProceedingJoinPoint
+import org.aspectj.lang.annotation.Around
 import org.aspectj.lang.annotation.Aspect
 import org.aspectj.lang.annotation.Before
 import org.slf4j.MDC
 import org.springframework.messaging.Message
+import org.springframework.messaging.support.GenericMessage
 import org.springframework.stereotype.Component
 import org.springframework.web.servlet.HandlerInterceptor
 import java.lang.Exception
@@ -51,7 +54,7 @@ class CorrelationIdMdcInterceptor : HandlerInterceptor {
 class CorrelationIdMdcMessageListenerAspect {
 
     /**
-     * Pointcut for inbound [Message]s (ie. SQS Message's being directed to a listener class) that sets the correlation ID
+     * Before Advice for inbound [Message]s (ie. SQS Message's being directed to a listener class) that sets the correlation ID
      * MDC variable to the value found in the Message header `x-correlation-id` if set, or a new value.
      * This allows for passing and logging a consistent correlation ID between disparate systems or processes.
      */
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
```diff
@@ -0,0 +1,96 @@
...
+ */
+@Component
+class CorrelationIdMdcInterceptor : HandlerInterceptor {
+
+    override fun preHandle(request: HttpServletRequest, response: HttpServletResponse, handler: Any): Boolean {
+        MDC.put(CORRELATION_ID, request.getHeader(CORRELATION_ID_HEADER) ?: generateCorrelationId())
+        return true
+    }
+
+    override fun afterCompletion(
```
NathanRussellValtech, line 35: Not sure if we need this in an interceptor (the afterCompletion to remove from MDC)
I've done similar things in other projects in the past and I vaguely remember us having to do this, but I can't quite remember why? MDC is ThreadLocal I think ? Not sure why you'd need to do this?

src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
```diff
@@ -0,0 +1,96 @@
...
+@Aspect
+@Component
+class CorrelationIdMdcMessageListenerAspect {
+
+    /**
+     * Pointcut for inbound [Message]s (ie. SQS Message's being directed to a listener class) that sets the correlation ID
+     * MDC variable to the value found in the Message header `x-correlation-id` if set, or a new value.
+     * This allows for passing and logging a consistent correlation ID between disparate systems or processes.
+     */
+    @Before("execution(* org.springframework.messaging.handler.invocation.AbstractMethodMessageHandler.handleMessage(..))")
```
NathanRussellValtech, line 58: This aspect works 👍

src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
```diff
@@ -0,0 +1,96 @@
...
+        val message = joinPoint.args[0] as Message<*>?
+        MDC.put(CORRELATION_ID, message?.headers?.get(CORRELATION_ID_HEADER)?.toString() ?: generateCorrelationId())
+    }
+
+    /**
+     * Pointcut for outbound [Message]s (ie. SQS Message's being sent) that sets the correlation ID
+     * header on the [Message] to either the existing MDC variable or a new value if not set in MDC.
+     * This allows for passing and logging a consistent correlation ID between disparate systems or processes.
+     */
+    @Before("execution(* org.springframework.messaging.support.AbstractMessageChannel.send(..))")
```
NathanRussellValtech, line 69: For the life of me I cannot get this aspect to work 😢 
I've asked [this question on SO](https://stackoverflow.com/questions/74486889/spring-aop-pointcut-execution-not-working) about it

src/main/resources/application.yml
```diff
@@ -43,6 +43,8 @@ thread-pool:
     max-size: 5
 
 logging:
+  pattern:
+    level: "%X{correlationId}%5p"
```
NathanRussellValtech, line 47: Not sure why the property is `logging.pattern.level` ... but is is!
This results in log statements like this, where the correlation ID comes straight after the date:
```
2022-11-18 13:58:40.726 1b9e3569caad4a4fab8538f61c14c6a7 INFO 50873 --- [enerContainer-5] .m.SendApplicationToPrintMessageListener : Print message with source reference [63778f8fe83c0a36ba274284] saved
```

src/test/resources/logback-test.xml
```diff
@@ -1,13 +1,7 @@
 <configuration>
     <include resource="org/springframework/boot/logging/logback/base.xml" />
-    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
```
NathanRussellValtech, line 3: This just tidies our test logback - we were duplicating STDOUT with a slightly different appender pattern for some reason? Anyway, works better without it 👍

NathanRussellValtech:
Closing this PR as its been implemented now

PR #82, opened 2022-11-18
Feature/eip1 1990 service to return certificate summary

Neil-Massey:
Service that retrieves a Certificate based on the ERO, Source Type and Source Reference and maps to the internal DTO.


src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClient.kt
```diff
@@ -21,43 +22,56 @@ class ElectoralRegistrationOfficeManagementApiClient(
...
+
     /**
      * Calls the `ero-management-api` to return a [EroManagementApiEroDto] for the specified gssCode.
      *
      * @param gssCode the gssCode of the localAuthority for the ERO to be returned
      * @return an [EroManagementApiEroDto] for the ERO
      * @throws [ElectoralRegistrationOfficeManagementApiException] concrete implementation if the API returns an error
      */
-    fun getElectoralRegistrationOffice(gssCode: String): EroManagementApiEroDto {
+    fun getElectoralRegistrationOffices(gssCode: String): EroManagementApiEroDto {
```
adacayi, line 48: @Neil-Massey , since this will always return a single ero if successful, the naming confused me actually. 

Maybe something like this `getElectoralRegistrationOfficeByGssCode` is more clear?

src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClient.kt
```diff
@@ -21,43 +22,56 @@ class ElectoralRegistrationOfficeManagementApiClient(
...
             .uri("/eros?gssCode=$gssCode")
             .retrieve()
             .bodyToMono(ElectoralRegistrationOfficesResponse::class.java)
-            .onErrorResume { ex -> handleException(ex, gssCode) }
+            .onErrorResume { ex -> handleException(ex, "gssCode = $gssCode") }
             .block()!!
 
         if (response.eros.isEmpty()) {
-            throw ElectoralRegistrationOfficeNotFoundException(gssCode)
+            throw ElectoralRegistrationOfficeNotFoundException("ERO not found for gssCode = $gssCode")
```
adacayi, line 58: @Neil-Massey . how about just passing in the search criteria only to the NotFoundException and let it deal with the prefix?

<pre>
throw ElectoralRegistrationOfficeNotFoundException(<b>"gssCode = $gssCode"</b>)

class ElectoralRegistrationOfficeNotFoundException(<b>searchCriteria</b>: String) :
    ElectoralRegistrationOfficeManagementApiException(<b>"ERO not found for $searchCriteria"</b>)
</pre>

src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiException.kt
```diff
@@ -9,8 +9,8 @@ package uk.gov.dluhc.printapi.client
 
 abstract class ElectoralRegistrationOfficeManagementApiException(message: String) : RuntimeException(message)
 
-class ElectoralRegistrationOfficeNotFoundException(gssCode: String) :
-    ElectoralRegistrationOfficeManagementApiException("ERO for gssCode $gssCode not found")
+class ElectoralRegistrationOfficeNotFoundException(message: String) :
```
adacayi, line 12: @Neil-Massey , do you think it makes sense to change it to the below, so we guarantee all not found exception messages have the same format:

<pre>
class ElectoralRegistrationOfficeNotFoundException(searchCriteria: String) :
    ElectoralRegistrationOfficeManagementApiException("ERO not found for $searchCriteria")
    </pre>

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
```diff
@@ -121,8 +121,7 @@ class Certificate(
 
     private fun assignStatus() {
         val currentPrintRequest = getCurrentPrintRequest()
-        currentPrintRequest.statusHistory.sortByDescending { it.eventDateTime }
-        val currentStatus = currentPrintRequest.statusHistory.first()
+        val currentStatus = currentPrintRequest.getCurrentStatus()
```
adacayi, line 124: Nice change 👍

src/main/kotlin/uk/gov/dluhc/printapi/database/mapper/CertificateSummaryDtoMapper.kt
```diff
@@ -0,0 +1,31 @@
...
+    fun certificateToCertificatePrintRequestSummaryDto(certificate: Certificate): CertificateSummaryDto {
+        return CertificateSummaryDto(
+            vacNumber = certificate.vacNumber!!,
+            printRequests = toPrintRequests(certificate.printRequests)
+        )
+    }
+
+    private fun toPrintRequests(printRequests: MutableList<PrintRequest>): List<PrintRequestSummaryDto> =
+        printRequests.map { printRequest -> toPrintRequestSummaryDto(printRequest) }.toList()
+
```
adacayi, line 21: @Neil-Massey , I think we can safely omit .toList() at the end.

<pre>
printRequests.map { printRequest -> toPrintRequestSummaryDto(printRequest) } </pre>

src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClient.kt
```diff
@@ -21,43 +22,56 @@ class ElectoralRegistrationOfficeManagementApiClient(
...
 
+    /**
+     * Calls the `ero-management-api` to return a [ElectoralRegistrationOfficeResponse] for the specified eroId.
+     *
+     * @param eroId the ID of the ERO to return
+     * @return a [ElectoralRegistrationOfficeResponse] for the ERO
+     * @throws [ElectoralRegistrationOfficeManagementApiException] concrete implementation if the API returns an error
+     */
+    fun getElectoralRegistrationOffice(eroId: String): ElectoralRegistrationOfficeResponse =
+        eroManagementWebClient
```
adacayi, line 33: @Neil-Massey , I think it will be better if we don't expose the external `ElectoralRegistrationOfficeResponse` to the service layer and return a dto instead here like we did with the method below.

Neil-Massey: This is a copy of approach from VCA:
https://github.com/cabinetoffice/eip-ero-voter-card-applications-api/blob/main/src/main/kotlin/uk/gov/dluhc/votercardapplicationsapi/client/ElectoralRegistrationOfficeManagementApiClient.kt

src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClient.kt
Neil-Massey: I think this is how the diff is displayed, this method name has not been changed

src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClient.kt
Neil-Massey: I was in two minds - the `ElectoralRegistrationOfficeGeneralException` was taking full message so went with same approach:
```
abstract class ElectoralRegistrationOfficeManagementApiException(message: String) : RuntimeException(message)

class ElectoralRegistrationOfficeNotFoundException(message: String) :
    ElectoralRegistrationOfficeManagementApiException(message)

class ElectoralRegistrationOfficeGeneralException(message: String) :
    ElectoralRegistrationOfficeManagementApiException(message)
```

Splitting the message and doing string manipulation as you suggested seemed a bit strange (I started with that approach).  But repeating the message at each call site seems wrong too.  Will review on Monday.

src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClient.kt
Neil-Massey: But will review suggestion on Monday as your approach may be better 😄 .

src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiException.kt
Neil-Massey: As above

src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClient.kt
Neil-Massey: Done - better to map to internal model 👍

src/test/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClientTest.kt
```diff
@@ -53,144 +57,218 @@ internal class ElectoralRegistrationOfficeManagementApiClientTest {
...
+            // Then
+            assertThat(ex.message).isEqualTo(expectedException.message)
+            assertThat(clientRequest.value.url()).hasHost("ero-management-api").hasPath("/eros/$eroId")
         }
-        given(eroMapper.toEroManagementApiEroDto(any())).willReturn(expected)
 
-        // When
-        val ero = apiClient.getElectoralRegistrationOffice(gssCode)
+        @Test
+        fun `should not get Electoral Registration Office given API returns a 500 error`() {
```
adacayi, line 107: @Neil-Massey , do you think it makes sense to cover the non-WebClientResponseException case as well, similar to this test `should throw exception given API returns a non WebClientResponseException` below?

src/main/kotlin/uk/gov/dluhc/printapi/dto/PrintRequestSummaryDto.kt
```diff
@@ -0,0 +1,11 @@
+package uk.gov.dluhc.printapi.dto
+
+import uk.gov.dluhc.printapi.database.entity.Status
+import java.time.Instant
+
+data class PrintRequestSummaryDto(
+    private val status: Status,
```
adacayi, line 7: @Neil-Massey , I think it will be good to have a separate `Status` enum for the dto to decouple it from the entity enum.

src/test/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepositoryIntegrationTest.kt
```diff
@@ -179,6 +181,44 @@ internal class CertificateRepositoryIntegrationTest : IntegrationTest() {
         assertCertificateRecursiveEqual(actual[0], expected)
     }
 
+    @Nested
+    inner class FindByGssCodeInAndSourceTypeAndSourceReference {
+        @Test
+        fun `should find certificate given one exists for provided details`() {
```
adacayi, line 187: Not part of this ticket, I assume, but is there a unique index in the DB to prevent multiple certificates with the same source type and source reference and gssCode? (I guess we cannot have two certificates with the same source type and sourceReference alone). I think it will be good to prevent that in the db layer.

src/test/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClientTest.kt
Neil-Massey: Yes, this is missing from source reference: https://github.com/cabinetoffice/eip-ero-voter-card-applications-api/blob/main/src/test/kotlin/uk/gov/dluhc/votercardapplicationsapi/client/ElectoralRegistrationOfficeManagementApiClientTest.kt

src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClient.kt
adacayi: Thanks, @Neil-Massey , for addressing it.

src/main/kotlin/uk/gov/dluhc/printapi/client/ElectoralRegistrationOfficeManagementApiClient.kt
adacayi: Thanks, @Neil-Massey.

adacayi (approved):
LGTM, @Neil-Massey . 
Thanks for addressing the suggestions.

There are a few suggestions which I added recently. 
If you can have a look at them, that will great.
Thanks, again.

src/main/kotlin/uk/gov/dluhc/printapi/dto/PrintRequestSummaryDto.kt
Neil-Massey: Good spot - I'm not convinced with need so many representations of the same enum and mappings, but that's the design 👍

Neil-Massey: Adds so much complexity for what benefit?

PR #83, opened 2022-11-18
EIP1-1878 - Improve error handling when renaming print response files for processing

NathanRussellValtech:


adacayi (approved):
LGTM, @NathanRussellValtech.
Thanks for addressing the issue quickly.

I think it will be good to cover wrapping `MessagingException` into `IOException` with a unit test as well.

NathanRussellValtech:
Thanks @adacayi - Yes, I did look to add some extra tests, but couldnt find anything similar I could copy and paste! 🤣 
I'll merge it as it is so that it gets deployed ready for testing to resume on Monday. Perhaps we can beef up the tests next week in some way 🤔

PR #84, opened 2022-11-21
EIP1-2515 - set cron and batch size

NathanRussellValtech:
This PR sets the cron and batch size for the `print-api` cron jobs, as per the recommendations in EIP1-2515.

It also renames the `test` context used by integration tests to be consistent with our other apps (and allows for us to use an `application-test.yml` for our `test` environment as and when we build it)


Neil-Massey (approved):
LGTM

src/test/kotlin/uk/gov/dluhc/printapi/config/IntegrationTest.kt
```diff
@@ -50,7 +50,7 @@ private val logger = KotlinLogging.logger {}
     webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
     classes = [IntegrationTest.IntegrationTestConfiguration::class],
 )
-@ActiveProfiles("test")
+@ActiveProfiles("integration-test")
```
Neil-Massey, line 53: Good to disambiguate this from the AWS `test` environment 👍

adacayi, line 53: One question around this, @NathanRussellValtech. I think this chane will prevent getting properties from application-test.yml which might be necessary for the existing tests to be setup properly. 

Can you please explain why this change has been done? Thanks.

NathanRussellValtech: Hi @adacayi - we did this in the other apps as well - it might have been whilst you were off the other week

Basically the problem we are fixing is that we have an AWS environment called `test`. Its the env that live testing is being done in. There was a change in infra where we set the env name as the spring profile, which means we can have files in `main/resources` such as `application.yml`, but also `application-dev.yml`, `application-int.yml`, `application-test.yml` and `application-prod.yml` - We dont use it much, but it gives us another way to set environment specific properties other than via `infra`. It means we can override properties in `application.yml`
Like I said, we don't use it much, but the option is there. At the moment we are using it for some feature flag properties (to enable/disable in certain envs), and also to change the cron intervals in `dev`

The problem of course is that we already had an `application-test.yml` in the `test/resources` folder, which is used by the Integration tests. So what we have done here (and in the other apps) is rename this file to `application-integration-test.yml` and to change the active profile name to match.

Thats it 😁 👍

PR #86, opened 2022-11-21
EIP1-1990 Remove temporary health endpoint from API Gateway

Neil-Massey:


adacayi (approved):
LGTM, @Neil-Massey , thanks.
We will need to update infra accordingly as well.

PR #89, opened 2022-11-23
EIP1-2197: EIP1-2738: Implement repository method to query for daily sent certificates to print provider

adacayi:
This PR is to implement the repository method to query for daily sent certificates to print provider

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepository.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepository.kt
@@ -5,6 +5,7 @@ import org.springframework.stereotype.Repository
 import uk.gov.dluhc.printapi.database.entity.Certificate
 import uk.gov.dluhc.printapi.database.entity.SourceType
 import uk.gov.dluhc.printapi.database.entity.Status
+import java.time.Instant
 import java.util.UUID
 
 @Repository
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepository.kt
```diff
@@ -17,4 +18,10 @@ interface CertificateRepository : JpaRepository<Certificate, UUID> {
     fun findByStatus(certificateStatus: Status): List<Certificate>
 
     fun findByGssCodeInAndSourceTypeAndSourceReference(gssCodes: List<String>, sourceType: SourceType, sourceReference: String): Certificate?
+
+    fun findByPrintRequestsStatusHistoryEventDateTimeBetweenAndPrintRequestsStatusHistoryStatus(
```
Neil-Massey, line 22: Do we need to find the certificates or just get a `count`?  (I think for the rate limit just need a count)

adacayi: Thanks, @Neil-Massey, for the suggestions. 
I definitely agree with you. 
I will make the changes accordingly.

Neil-Massey (approved):
LGTM - thanks Abdullah

PR #90, opened 2022-11-24
EIP1-1878 - logging with correlationId

NathanRussellValtech:
This PR adds a correlation ID to all log messages using an MCD variable, Interceptors and some AOP

I've tried to think about this in a generic way so that it could be added to all of our apps, and it would "just work" .... 🤔

So, my thinking is that there are 3 "starting" points for our apps:

* an incoming HTTP request ie. a REST API
* a consumed (received) SQS message
* a scheduled task (@Scheduler)

for each of these 3 "starting" points we can use an Interceptor (http requests) or AOP (SQS and scheduled tasks) to generate a correlation ID and add it to MDC. We can then include the MDC variable in the log pattern. Simple stuff

But, we also want to be able to pass around / maintain a correlation ID across threads, and even across apps.

So, the Interceptor (http requests) and AOP (SQS) need to look at a header on the inbound request (HTTP request or SQS Message), and if there is a header x-correlation-id then set that value in the MDC, else set a new correlation ID in the MDC
We don't need to do this on the AOP aspect for scheduled tasks because nothing "calls" it - there is nothing upstream of it that would have set a correlation ID that we need to maintain

Next we need some way of passing the existing correlation ID from MDC into a header x-correlation-id when making outbound HTTP requests or producing (sending) SQS messages.
For SQS we can do that with more AOP 👍 
For outbound HTTP requests I imagine we'd use another interceptor (though I've not written that yet)

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
@@ -1,10 +1,13 @@
 package uk.gov.dluhc.printapi.config
 
 import org.aspectj.lang.JoinPoint
+import org.aspectj.lang.ProceedingJoinPoint
+import org.aspectj.lang.annotation.Around
 import org.aspectj.lang.annotation.Aspect
 import org.aspectj.lang.annotation.Before
 import org.slf4j.MDC
 import org.springframework.messaging.Message
+import org.springframework.messaging.support.GenericMessage
 import org.springframework.stereotype.Component
 import org.springframework.web.servlet.HandlerInterceptor
 import java.lang.Exception
@@ -51,7 +54,7 @@ class CorrelationIdMdcInterceptor : HandlerInterceptor {
 class CorrelationIdMdcMessageListenerAspect {
 
     /**
-     * Pointcut for inbound [Message]s (ie. SQS Message's being directed to a listener class) that sets the correlation ID
+     * Before Advice for inbound [Message]s (ie. SQS Message's being directed to a listener class) that sets the correlation ID
      * MDC variable to the value found in the Message header `x-correlation-id` if set, or a new value.
      * This allows for passing and logging a consistent correlation ID between disparate systems or processes.
      */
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/messaging/MessageQueue.kt
```diff
@@ -1,12 +1,24 @@
...
 
 class MessageQueue<T : Any>(
     private val queueName: String,
-    private val queueMessagingTemplate: QueueMessagingTemplate
+    private val queueMessagingTemplate: QueueMessagingTemplate,
 ) {
 
-    fun submit(message: T) =
-        queueMessagingTemplate.convertAndSend(queueName, message)
+    fun submit(payload: T) =
```
NathanRussellValtech, line 12: This is the only bit that is a perhaps a little .... eyebrow raising ....

Previously we were using `convertAndSend` on `QueueMessagingTemplate`. This calls a handful of methods in `QueueMessageTemplate` to convert the payload into a `Message`, then send the `Message` via the `send` method.

We need our aspect to get hold of the `Message`, so we have a pointcut over the `send` method (in the abstract class, but for all intents and purposes its a pointcut over `QueueMessageTemplate#send`

This didn't work :( It turns out that it doesn't work because of the proxy that AOP creates, in that you can only weave over methods that are called directly. You cant weave over methods that are called "internally" by the proxy! 

Kinda makes sense, but has forced me to change our class here so that it calls `QueueMessageTemplate.send`

Which takes me onto the next eyebrow raising part ...

In order to call `send` we need a `Message`. To create a message with all the correct headers, and to use the correct message converters etc we want to be able to call `QueueMessageTemplate.doConvert` .... but it's `protected`, so we can't. A cheeky extension function with some reflection hum-dingery 😁 to the rescue!
(or some might call it a hack? 😁 👍 )

src/test/resources/logback-test.xml
```diff
@@ -1,13 +1,7 @@
 <configuration>
     <include resource="org/springframework/boot/logging/logback/base.xml" />
-    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
```
NathanRussellValtech, line 3: This just tidies our test logback - we were duplicating STDOUT with a slightly different appender pattern for some reason? Anyway, works better without it 👍

src/main/resources/application.yml
```diff
@@ -43,6 +43,8 @@ thread-pool:
     max-size: 5
 
 logging:
+  pattern:
+    level: "%X{correlationId}%5p"
```
NathanRussellValtech, line 47: Not sure why the property is logging.pattern.level ... but is is!
This results in log statements like this, where the correlation ID comes straight after the date:
```
2022-11-18 13:58:40.726 1b9e3569caad4a4fab8538f61c14c6a7 INFO 50873 --- [enerContainer-5] .m.SendApplicationToPrintMessageListener : Print message with source reference [63778f8fe83c0a36ba274284] saved
```

src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
```diff
@@ -0,0 +1,110 @@
...
+
+    /**
+     * Before Advice for inbound [Message]s (ie. SQS Message's being directed to a listener class) that sets the correlation ID
+     * MDC variable to the value found in the Message header `x-correlation-id` if set, or a new value.
+     * This allows for passing and logging a consistent correlation ID between disparate systems or processes.
+     */
+    @Before("execution(* org.springframework.messaging.handler.invocation.AbstractMethodMessageHandler.handleMessage(..))")
+    fun beforeHandleMessage(joinPoint: JoinPoint) {
+        val message = joinPoint.args[0] as Message<*>?
+        MDC.put(CORRELATION_ID, message?.headers?.get(CORRELATION_ID_HEADER)?.toString() ?: generateCorrelationId())
```
Neil-Massey, line 64: Where is this removed?

src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
```diff
@@ -0,0 +1,110 @@
...
+class CorrelationIdMdcScheduledAspect {
+
+    /**
+     * Before Advice for Scheduled tasks (ie. cron tasks) that sets the correlation ID MDC variable to a new value.
+     * Due to the invocation semantics of a Scheduled task it does not make sense to pass a correlation ID from another
+     * system or process into it.
+     */
+    @Before("@annotation(org.springframework.scheduling.annotation.Scheduled)")
+    fun before(joinPoint: JoinPoint) {
+        MDC.put(CORRELATION_ID, generateCorrelationId())
```
Neil-Massey, line 102: Where is this removed?

NathanRussellValtech: Hmmm, they're not currently .... I was thinking this might be OK as I was assuming each cron invocation and SQS handling (the other one you commented on) were on a thread in their own right, and that MDC was thread local ... and that it might be OK?
But perhaps not? 🤔 

Do you think these `@Before` advices might be better off being `@Around`, and then we set the MDC, invoke the method, then clear the MDC ?
Would that be a better approach ?

Neil-Massey: They are thread local, but if that thread is reused and the correlation id is not reset it can become misleading.  Always best to unset the thread local once done with it.  I think your `@Around` suggestion makes sense.

NathanRussellValtech: OK, I've changed these advices to `@Around` in order to clear the MDC 👍

Neil-Massey (approved):
LGTM - As all existing tests are passing should indicate no bugs are introduced by the change 👍  (And no specific ones for the correlation are worthwhile)

PR #91, opened 2022-11-24
EIP1-2197: EIP1-2739: Limit daily certificates sent to print provider

adacayi:
This PR is to limit the daily certificates sent to print provider.

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
@@ -51,8 +51,8 @@ class CertificateBatchingService(
         val countOfRequestsSentToPrintProvider =
             certificateRepository.getPrintRequestStatusCount(startOfDay, endOfDay, Status.ASSIGNED_TO_BATCH)
 
-        if (certificatesPendingAssignment.size + countOfRequestsSentToPrintProvider > dailyLimit) {
-            logDailyLimit(certificatesPendingAssignment, countOfRequestsSentToPrintProvider)
+        if ((certificatesPendingAssignment.size + countOfRequestsSentToPrintProvider) > dailyLimit) {
+            logDailyLimit(certificatesPendingAssignment, countOfRequestsSentToPrintProvider, endOfDay)
             return certificatesPendingAssignment.subList(0, dailyLimit - countOfRequestsSentToPrintProvider)
         }
 
@@ -61,14 +61,16 @@ class CertificateBatchingService(
 
     private fun logDailyLimit(
         certificatesPendingAssignment: List<Certificate>,
-        countOfRequestsSentToPrintProvider: Int
+        countOfRequestsSentToPrintProvider: Int,
+        endOfDay: Instant
     ) {
+        val nextDay = endOfDay.plusSeconds(1)
         logger.warn {
             """Identified ${certificatesPendingAssignment.size} certificates to assign to a batch. 
             Daily print limit is $dailyLimit. 
             $countOfRequestsSentToPrintProvider certificates already sent to print provider today.
             Remaining capacity is ${dailyLimit - countOfRequestsSentToPrintProvider}.
-            ${certificatesPendingAssignment.size + countOfRequestsSentToPrintProvider - dailyLimit} certificates won't be assigned to a batch."""
+            ${certificatesPendingAssignment.size + countOfRequestsSentToPrintProvider - dailyLimit} certificates won't be assigned to a batch until $nextDay at the earliest."""
         }
     }
 }
--- a/src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
@@ -1,10 +1,14 @@
 package uk.gov.dluhc.printapi.service
 
 import mu.KotlinLogging
+import org.springframework.beans.factory.annotation.Value
 import org.springframework.stereotype.Service
 import uk.gov.dluhc.printapi.database.entity.Certificate
 import uk.gov.dluhc.printapi.database.entity.Status
 import uk.gov.dluhc.printapi.database.repository.CertificateRepository
+import java.time.Clock
+import java.time.Instant
+import java.time.temporal.ChronoUnit.DAYS
 import javax.transaction.Transactional
 
 private val logger = KotlinLogging.logger { }
@@ -12,9 +16,11 @@ private val logger = KotlinLogging.logger { }
 @Service
 class CertificateBatchingService(
     private val idFactory: IdFactory,
-    private val certificateRepository: CertificateRepository
+    private val certificateRepository: CertificateRepository,
+    @Value("\${jobs.batch-print-requests.daily-limit}")
+    private val dailyLimit: Int,
+    private val clock: Clock
 ) {
-
     @Transactional
     fun batchPendingCertificates(batchSize: Int): Set<String> {
         val batches = batchCertificates(batchSize)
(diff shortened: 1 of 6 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
```diff
@@ -29,12 +36,41 @@ class CertificateBatchingService(
...
+    private fun limitCertificates(certificatesPendingAssignment: List<Certificate>): List<Certificate> {
+        val currentDate = LocalDate.ofInstant(Instant.now(clock), ZoneOffset.UTC)
+        val startOfDay = currentDate.atStartOfDay().toInstant(ZoneOffset.UTC)
+        val endOfDay = currentDate.plusDays(1).atStartOfDay().toInstant(ZoneOffset.UTC).minusSeconds(1)
+        val countOfRequestsSentToPrintProvider =
+            certificateRepository.getPrintRequestStatusCount(startOfDay, endOfDay, Status.ASSIGNED_TO_BATCH)
+
+        if (certificatesPendingAssignment.size + countOfRequestsSentToPrintProvider> dailyLimit) {
+            logDailyLimit(certificatesPendingAssignment, countOfRequestsSentToPrintProvider)
+            return certificatesPendingAssignment.subList(0, dailyLimit - countOfRequestsSentToPrintProvider)
```
adacayi, line 58: Since it is easier to calculate how many certificates can be sent to print provider, I returned the sub-list accordingly. 

It would be more complicated to return the number corresponding only full-sized batches and the log will be more complicated to understand as well.

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
```diff
@@ -29,12 +35,40 @@ class CertificateBatchingService(
...
+    private fun logDailyLimit(
+        certificatesPendingAssignment: List<Certificate>,
+        countOfRequestsSentToPrintProvider: Int
+    ) {
+        logger.warn {
+            """Identified ${certificatesPendingAssignment.size} certificates to assign to a batch. 
+            Daily print limit is $dailyLimit. 
+            $countOfRequestsSentToPrintProvider certificates already sent to print provider today.
+            Remaining capacity is ${dailyLimit - countOfRequestsSentToPrintProvider}.
+            ${certificatesPendingAssignment.size + countOfRequestsSentToPrintProvider - dailyLimit} certificates won't be assigned to a batch."""
```
NathanRussellValtech, line 71: Minor, might be worth changing the final sentence very slightly to make it clear the certs that wont be batched will be batched tomorrow - its not that they wont be batched ever.
Something like:
```
... won't be assigned to a batch today.
```
or
```
... won't be assigned to a batch until tomorrow.
```
or
```
... won't be assigned to a batch until <tomorrow date> at the earliest.
```
(where <tomorrow date> is an ISO8601 of tomorrow 00:0:00 - eg: `2022-11-25T00:00:00.000Z` - might be more complex than its worth though, so happy if you decide not to do this)

src/main/resources/application-int.yml
```diff
@@ -6,3 +6,7 @@ logging:
     com:
       jcraft:
         jsch: DEBUG
+jobs:
+  batch-print-requests:
+    daily-limit: 20
```
NathanRussellValtech, line 11: I don't think I'd do this in `int` - Vijay is still integration testing with the Print Provider in `int`, and they are sending more that 20 print requests per day.
Doing it in `dev` is fine, but make sure you let the QA community know it's set that way so that a) they understand how to test this story and b) the understand why print requests might not be batched as expected if they are not testing this story! 😁

src/main/resources/application.yml
```diff
@@ -33,6 +33,7 @@ jobs:
     name: "BatchPrintRequests"
     cron: "0 0/15 * * * *" # every 15 minutes, starting on the hour - see analysis and recommendations in EIP1-2515
     batch-size: 50
+    daily-limit: 150000
```
NathanRussellValtech, line 36: Minor (and I can live without this if its not possible ... I'm not sure if it is even possible ....)  ... do Spring properties allow to you express numbers with comma place delimiters? IE. can you do this?
```
daily-limit: 150,000
```
The only reason I mention it is that I'm currently viewing this on a smaller screen. Where I'm in my more "senior years" 😁 my eyesight is not what it used to be (even with my glasses on!), and I had to zoom the screen to determine that it was 150K rather than 15K !! 🤣 

(God, I'm so blind! 🤣 )

NathanRussellValtech (approved):
A couple of minor comments, but LGTM

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
```diff
@@ -29,12 +35,40 @@ class CertificateBatchingService(
...
         }
     }
+
+    private fun limitCertificates(certificatesPendingAssignment: List<Certificate>): List<Certificate> {
+        val startOfDay = Instant.now(clock).truncatedTo(DAYS)
+        val endOfDay = startOfDay.plus(1, DAYS).minusSeconds(1)
+        val countOfRequestsSentToPrintProvider =
+            certificateRepository.getPrintRequestStatusCount(startOfDay, endOfDay, Status.ASSIGNED_TO_BATCH)
+
+        if (certificatesPendingAssignment.size + countOfRequestsSentToPrintProvider > dailyLimit) {
```
Neil-Massey, line 54: please can brackets be added so don't have to worry about the precedence order of the different operators

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
Neil-Massey: I guess in theory we will write this message every time the batching task runs after the daily limit is reached.
Technically we are counting certificates that have been added to a batch - if there has been problems writing to SFTP or more generally the zipping task fails... we won't have "sent to print provider today".

Remaining capacity seems a bit misleading as after the current batch is processed there is no remaining capacity.
Maybe could simplify to something like:
```
  Daily print limit of X reached.  
  X certificates pending assignment will be processed tomorrow.
````

Neil-Massey (approved):
LGTM - Few minor suggestions

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
adacayi: Thanks, @NathanRussellValtech ,
I agree, the current log message might be misleading. I will implement the second option you proposed.

Thanks, @Neil-Massey , for your suggestion as well. 
I might have missed your point, but the remaining capacity in the message refers to the number of requests that can be added to a batch as part of the current batching task. When it runs the next time, and the daily limit is reached, the remaining capacity will be presented as 0.

src/main/resources/application-int.yml
adacayi: Thanks, @NathanRussellValtech ,
I will get in touch with Vijay, to set a sensible limit for the int environment.

src/main/resources/application.yml
adacayi: Thanks, for the suggestion @NathanRussellValtech. 
I will definitely look into it to make it more readable if possible.

adacayi: @NathanRussellValtech , seems like we can do this:

<pre>
daily-limit: 150_000
</pre>

NathanRussellValtech: Awesome! Hurrah for Spring being able to cater for my old man eyes! 🤣 🎉

src/main/resources/application-int.yml
adacayi: @NathanRussellValtech , after talking to Vijay, we agreed on the daily limits to be 10 for dev and 1000 for int respectively. 
He said he will first look into it in dev, and 10 will be easier to test. 
I will make the changes accordingly.

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
Neil-Massey: When I read the message the remaining capacity is always zero - so didn't think including it was useful.

PR #92, opened 2022-11-25
EIP1-1878 - Fix correlation ID; added integration tests

NathanRussellValtech:
This PR fixes a problem where the correlation ID was not being passed correctly ; and introduces some integration tests to test and assert this cross-cutting concern

(At last, I've found a valid reason and use case for basing assertions around log messages! 🤣 Thank goodness for our `TestLogAppender` 👍 😁 )


src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
```diff
@@ -111,4 +111,4 @@ private fun generateCorrelationId(): String =
     UUID.randomUUID().toString().replace("-", "")
 
 private fun getCurrentCorrelationId(): String =
-    MDC.get(CORRELATION_ID_HEADER) ?: generateCorrelationId()
+    MDC.get(CORRELATION_ID) ?: generateCorrelationId()
```
NathanRussellValtech, line 114: This is the fix! Tiny stupid typo! 😁 
(If I'd have written tests in the first place it would have caught it! 🙄 )

src/main/kotlin/uk/gov/dluhc/printapi/rest/SecurityController.kt
```diff
@@ -1,23 +0,0 @@
-package uk.gov.dluhc.printapi.rest
```
NathanRussellValtech, line 1: I've deleted this class and it's test because they are no longer needed now we have other REST APIs that use and test the security. This class should be been deleted a while back. Hey ho 👍

src/test/kotlin/uk/gov/dluhc/printapi/logging/CorrelationIdMdcIntegrationTest.kt
```diff
@@ -0,0 +1,302 @@
...
+                ).hasCorrelationId(expectedCorrelationId)
+                assertThat(
+                    TestLogAppender.getLogEventMatchingRegex(
+                        "Begin processing PrintResponse with requestId $expectedRequestId",
+                        Level.INFO
+                    )
+                ).hasCorrelationId(expectedCorrelationId)
+            }
+        }
+    }
```
Neil-Massey, line 270: Looks good 👍 

Only thing I would question is whether we are testing different things in above two tests - are they both variations of running batch job that submits to SQS listener and then processes the SQS message.  Interesting that the second flow includes log about sending SQS message so is probably most useful.

Neil-Massey (approved):
Looks great!

src/test/kotlin/uk/gov/dluhc/printapi/logging/CorrelationIdMdcIntegrationTest.kt
NathanRussellValtech: Yeah, they are kinda testing the same thing in that the correlation ID is being passed between processes.
I agree, I think the 2nd test (ProcessPrintResponsesBatch) is possibly more useful because its going through more async processes:
ProcessPrintResponsesBatch is a scheduled job which sends a message to a queue. That message is picked up by it's listener, which in turn will send another message to another queue. That 2nd message is picket up by its listener to complete the processing.
So in this case its nice to see/assert the correlation ID has gone through 3 separate processes 👍 

The other test for BatchPrintRequests is basically the same, but there's only 1 queue involved. So its:
BatchPrintRequests is a scheduled job which sends a message to a queue. That message is picked up by it's listener to complete the processing.

I quite like having them both, if nothing else via their comments and assertions, to demonstrate and document what the correlation ID is supposed to do and what value it will have for us when trawling logs

PR #96, opened 2022-11-29
EIP1-2471 - Update Print Provider schema `certificateFormat` enum

NathanRussellValtech:
This PR updates the `certificateFormat` enum as used in `PrintRequest` in the Print Provider schema; by renaming `largeprint` to `large-print` and `audio` to `easy-read`. This change is as per agreement with the Print Provider

Things to note:
* The field and enum is still called `certificateFormat` even though it is **not** the certificate format, and is in fact the format of the supporting information that is sent with the certificate.
  * I was not able to negotiate changing the field name itself as the Print Provider have already developed to the existing spec and are not open to that degree of refactoring.
  * I have updated the `description` attribute to make this as clear as possible
* I have up'versioned the spec; but also changed to semantic versioning with `1.0.0` being this version.
  * I have added a Change History table to the `README.md`
* I have removed the `default` values from the yaml schema for `certificateFormat` and `certificateLanguage` and moved these to our mapper class
  * This will make the next PR of mapping the value from the PrintRequest database entity into a PrintRequest PSV easier (`PrintRequest.supportingInformationFormat` (entity) into `PrintRequest.certificateFormat` (PSV model))

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/mapper/CertificateToPrintRequestMapper.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/mapper/CertificateToPrintRequestMapper.kt
@@ -23,6 +23,8 @@ abstract class CertificateToPrintRequestMapper {
     @Mapping(source = "printRequest.delivery.address.locality", target = "deliveryLocality")
     @Mapping(source = "printRequest.delivery.address.area", target = "deliveryArea")
     @Mapping(source = "printRequest.delivery.address.postcode", target = "deliveryPostcode")
+    @Mapping(expression = "java( uk.gov.dluhc.printapi.printprovider.models.PrintRequest.CertificateLanguage.EN )", target = "certificateLanguage")
+    @Mapping(expression = "java( uk.gov.dluhc.printapi.printprovider.models.PrintRequest.CertificateFormat.STANDARD )", target = "certificateFormat")
     @Mapping(source = "printRequest.delivery.deliveryClass", target = "deliveryOption")
     @Mapping(source = "printRequest.eroEnglish.name", target = "eroNameEn")
     @Mapping(source = "printRequest.eroEnglish.phoneNumber", target = "eroPhoneNumberEn")
--- a/src/main/resources/yamlschema/README.md
+++ b/src/main/resources/yamlschema/README.md
@@ -1,8 +1,18 @@
 # Print Provider Schema
-# v0.0.5
+# v1.0.0
 This folder contains the Print Provider schema files.
 
 The schema is defined in `yamlschema`. Each type is defined in its own file.
+
+## Change History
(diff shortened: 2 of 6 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/mapper/CertificateToPrintRequestMapper.kt
```diff
@@ -23,6 +23,8 @@ abstract class CertificateToPrintRequestMapper {
     @Mapping(source = "printRequest.delivery.address.locality", target = "deliveryLocality")
     @Mapping(source = "printRequest.delivery.address.area", target = "deliveryArea")
     @Mapping(source = "printRequest.delivery.address.postcode", target = "deliveryPostcode")
+    @Mapping(expression = "java( uk.gov.dluhc.printapi.printprovider.models.PrintRequest.CertificateLanguage.EN )", target = "certificateLanguage")
+    @Mapping(expression = "java( uk.gov.dluhc.printapi.printprovider.models.PrintRequest.CertificateFormat.STANDARD )", target = "certificateFormat")
```
NathanRussellValtech, line 27: This mapping line will change in the next PR, to map from the db entity field into this field

adacayi (approved):
LGTM, @NathanRussellValtech.
Thanks.

src/main/resources/yamlschema/README.md
```diff
@@ -1,9 +1,19 @@
 # Print Provider Schema
-# v0.0.5
+# v1.0.0
 This folder contains the Print Provider schema files.
 
 The schema is defined in `yamlschema`. Each type is defined in its own file.
 
+## Change History
```
Neil-Massey, line 7: Helpful summary 👍 
Is this only for specs that are shared externally with third parties?
Or something we may start doing with internal specs to?

Neil-Massey: Should this be `Print Request Schema Change History`?  (Do we need section for the `Print Response` and `Print Responses` specs?

Neil-Massey (approved):

src/main/resources/yamlschema/README.md
NathanRussellValtech: I don't mind us putting a change history on our internal spec - totally happy for that if we feel it has value?

This is an external spec, and given that changes are negotiated and agreed and the spec versioned, I felt it was an additional layer of documentation/professionalism to document the change history. Sure, we have the git history, but I felt this was easier and quicker to reference.

Re: should there be 2 change histories? One for `Print Request` and one for `Print Response`? I don't think so. The schema should be considered one schema for the EROP -> Print Provider integration. Even though it's defined in separate files and contains 2 logical things (requests and responses), overall it's one schema. This is why all the yaml files in the schema were up'versioned, even though individually they had nothing to do with the change being made here.

PR #97, opened 2022-11-29
EIP1-2471: EIP1-2780: Add supportingInformationFormat to SendApplicationToPrintMessage

adacayi:
This PR is to add the supporting information format to `SendApplicationToPrintMessage`, so the value can be received from VCA and later persisted and sent to print provider in the agreed format.

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/config/JacksonConfiguration.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/config/JacksonConfiguration.kt
@@ -21,6 +21,7 @@ class JacksonConfiguration {
             .addModule(KotlinModule.Builder().build())
             .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
             .disable(DeserializationFeature.READ_DATE_TIMESTAMPS_AS_NANOSECONDS)
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
             .serializationInclusion(Include.NON_NULL)
             .build()
 
--- a/src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
+++ b/src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Print API SQS Message Types
-  version: '1.8.0'
+  version: '1.7.0'
   description: |-
     Print API SQS Message Types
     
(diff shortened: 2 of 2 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/config/JacksonConfiguration.kt
```diff
@@ -21,6 +21,7 @@ class JacksonConfiguration {
             .addModule(KotlinModule.Builder().build())
             .disable(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS)
             .disable(DeserializationFeature.READ_DATE_TIMESTAMPS_AS_NANOSECONDS)
+            .configure(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES, false)
```
adacayi, line 24: I thought this was already added as part of the development process for extending the ero response to include new contact fields so that the application won't fail. 

Since we are happy to configure the deserializer to not fail for unknown fields in other apis, I think it will ease our jobs, if we have it here as well.

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -109,6 +109,8 @@ components:
           example: Smith
         certificateLanguage:
           $ref: '#/components/schemas/CertificateLanguage'
+        supportingInformationFormat:
```
NathanRussellValtech, line 112: Can we add this new property to the `required` array please?
We need it to be a mandatory property in the message (though see my other comment about chicken and egg 😁 )

src/main/kotlin/uk/gov/dluhc/printapi/config/JacksonConfiguration.kt
Neil-Massey: So did I... https://github.com/cabinetoffice/eip-ero-print-api/pull/93/files

Neil-Massey (approved):
LGTM

NathanRussellValtech (approved):
LGTM; (assuming you can add `supportingInformationFormat` as a required property 👍 )

But .... you won't be able to merge this yet - we have a chicken and egg situation! 🙄 
What you'll need to do it take this spec (1.8.0) including the required `supportingInformationFormat` property, and bring it into VCA, and then update the VCA code that creates this message to map in the `supportingInformationFormat` field
Once that change is merged and deployed we can then merge and deploy this change.

Or .... you merge it as you have developed (where `supportingInformationFormat` is not required); then do the VCA change, then come back to this and make the field mandatory and up-version it again :( 

Either way is a bit messy :(

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
adacayi: @NathanRussellValtech , maybe for this change I only change the Jackson Configuration to allow unknown fields and then update VCA message and the spec to include the new field and then update `print-api` with the new spec as well.

PR #99, opened 2022-11-29
EIP1-2529 Use contact details from the ERO Management API

Neil-Massey:
Print API uses the ERO term to refer to data related to either the ERO Management API ERO or Local Authority.


src/main/kotlin/uk/gov/dluhc/printapi/mapper/IssuerDtoMapper.kt
```diff
@@ -0,0 +1,28 @@
...
+import uk.gov.dluhc.printapi.dto.IssuerDto
+
+@Mapper
+abstract class IssuerDtoMapper {
+
+    @Mapping(target = "englishContactDetails", expression = "java(toIssuerContactDetailsDto( localAuthority.getName(), localAuthority.getContactDetailsEnglish() ))")
+    @Mapping(target = "welshContactDetails", expression = "java(toNullIssuerContactDetailsDto( localAuthority.getName(), localAuthority.getContactDetailsWelsh() ))")
+    abstract fun toIssuerDto(localAuthority: LocalAuthorityResponse): IssuerDto
+
+    fun toNullIssuerContactDetailsDto(name: String, contactDetails: ContactDetails?): IssuerContactDetailsDto? {
```
Neil-Massey, line 17: We have a bit of a mismatch between what data is available and what data we need - this block only returns a complete contact details. (Otherwise a partial one is attempted using only the name). A future story is likely to return the welsh name and that will resolve this issue.

src/main/kotlin/uk/gov/dluhc/printapi/dto/IssuerDto.kt
```diff
@@ -0,0 +1,6 @@
+package uk.gov.dluhc.printapi.dto
+
+data class IssuerDto(
```
Neil-Massey, line 3: Possibly this could be `LocalAuthorityDto` but not 100% sure that it will always only contain details related to the LA and never an ERO.

src/main/kotlin/uk/gov/dluhc/printapi/mapper/CertificateMapper.kt
```diff
@@ -21,19 +21,19 @@ abstract class CertificateMapper {
 
     @Mapping(target = "id", ignore = true)
     @Mapping(target = "vacNumber", expression = "java( idFactory.vacNumber() )")
-    @Mapping(source = "localAuthority", target = "issuingAuthority")
+    @Mapping(source = "message.gssCode", target = "gssCode")
```
adacayi, line 24: A minor suggestion @Neil-Massey .

Seems like mapping of the gssCode will be done without explicitly specifying it. Also,if there is no source inferred for a target value, seems like it is ignored as well for line 22.

So we can simplify the definition of `toCertificate` method as follows:

<pre>
    @Mapping(target = "vacNumber", expression = "java( idFactory.vacNumber() )")
    @Mapping(source = "issuer.englishContactDetails.name", target = "issuingAuthority")
    abstract fun toCertificate(message: SendApplicationToPrintMessage, issuer: IssuerDto): Certificate
</pre>

src/test/kotlin/uk/gov/dluhc/printapi/service/PrintServiceTest.kt
```diff
@@ -32,18 +32,17 @@ class PrintServiceTest {
...
-        given(certificateMapper.toCertificate(any(), any(), any())).willReturn(certificate)
+        given(eroClient.getIssuer(any())).willReturn(issuer)
+        given(certificateMapper.toCertificate(any(), any())).willReturn(certificate)
 
         // When
         printService.savePrintMessage(message)
 
         // Then
-        verify(eroClient).getElectoralRegistrationOffices(message.gssCode!!)
+        verify(eroClient).getIssuer(message.gssCode!!)
```
adacayi, line 45: Minor, @Neil-Massey . 
It will be good to verify the certificate mapper call as well.

<pre>
verify(certificateMapper).toCertificate(message, issuer)
</pre>

adacayi (approved):
Looks really good, @Neil-Massey. 
Thanks.

PR #100, opened 2022-11-30
EIP1-2471: EIP1-2780: Add supporting information format to send application to print message

adacayi:
This PR is to add supporting information format to send application to print sqs message.


Neil-Massey (approved):
LGTM

PR #102, opened 2022-11-30
EIP1-2471: EIP1-2782: Map supporting information format to db entity

adacayi:
This PR is to map the supporting information format in the `SendApplicationToPrintMessage` model to db entity to persist that value.

### DIFF
--- a/build.gradle.kts
+++ b/build.gradle.kts
@@ -133,6 +133,7 @@ tasks.withType<KotlinCompile> {
 tasks.withType<Test> {
     dependsOn(tasks.withType<GenerateTask>())
     useJUnitPlatform()
+    jvmArgs("--add-opens", "java.base/java.time=ALL-UNNAMED")
 }
 
 tasks.withType<GenerateTask> {
--- a/src/test/kotlin/uk/gov/dluhc/printapi/mapper/PrintRequestMapperTest.kt
+++ b/src/test/kotlin/uk/gov/dluhc/printapi/mapper/PrintRequestMapperTest.kt
@@ -15,16 +15,15 @@ import uk.gov.dluhc.printapi.database.entity.Address
 import uk.gov.dluhc.printapi.database.entity.Delivery
 import uk.gov.dluhc.printapi.database.entity.DeliveryClass
 import uk.gov.dluhc.printapi.database.entity.DeliveryMethod
-import uk.gov.dluhc.printapi.database.entity.ElectoralRegistrationOffice
 import uk.gov.dluhc.printapi.database.entity.PrintRequest
 import uk.gov.dluhc.printapi.database.entity.PrintRequestStatus
 import uk.gov.dluhc.printapi.database.entity.Status
 import uk.gov.dluhc.printapi.database.entity.SupportingInformationFormat.EASY_READ
-import uk.gov.dluhc.printapi.dto.EroContactDetailsDto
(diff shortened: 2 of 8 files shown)


Neil-Massey (approved):
LGTM

build.gradle.kts
```diff
@@ -133,6 +133,7 @@ tasks.withType<KotlinCompile> {
 tasks.withType<Test> {
     dependsOn(tasks.withType<GenerateTask>())
     useJUnitPlatform()
+    jvmArgs("--add-opens", "java.base/java.time=ALL-UNNAMED")
```
Neil-Massey, line 136: Why was this necessary?

adacayi: Sorry, @Neil-Massey, I forgot to add the explanation for this. 
Without this, we get an error when we want to spy `java.time` objects:

image

I needed to spy a `Clock` instance in the `PrintRequestMapperTest` to use `@InjectMocks` instead of using reflection to set the dependencies in the `PrintRequestMapper` hence added the above change.

src/test/kotlin/uk/gov/dluhc/printapi/mapper/PrintRequestMapperTest.kt
```diff
@@ -46,51 +49,32 @@ class PrintRequestMapperTest {
...
-                property = "Pencadlys Cyngor Gwynedd",
-                street = "Stryd y Jêl",
-                town = "Caernarfon",
-                area = "Gwynedd",
-                postcode = "LL55 1SH",
-            )
-        )
+        given(supportingInformationFormatMapper.toPrintRequestEntityEnum(any()))
+            .willReturn(supportingInformationFormatEntityEnum)
+        val expectedEnglishEroContactDetails = toEroContactDetails(ero.englishContactDetails)
```
adacayi, line 76: I changed these, so it is more explicit that the expected values are derived from the `eroDto` and is easier to read and understand.

build.gradle.kts
Neil-Massey: Thanks @abdullah-sanver-valtech - I wondered if that may have bee the reason but wasn't sure 👍

src/test/kotlin/uk/gov/dluhc/printapi/mapper/PrintRequestMapperTest.kt
adacayi: Thanks, @Neil-Massey , for your suggestion.
Added it as an extension method for tests.

<pre>
val expectedEnglishEroContactDetails = ero.englishContactDetails.toElectoralRegistrationOffice()
val expectedWelshEroContactDetails = ero.welshContactDetails!!.toElectoralRegistrationOffice()
</pre>

adacayi: <pre>
package uk.gov.dluhc.printapi.testsupport.testdata.dto

fun EroContactDetailsDto.toElectoralRegistrationOffice() =
    ElectoralRegistrationOffice(
        name = name,
        phoneNumber = phoneNumber,
        website = website,
        emailAddress = emailAddress,
        address = with(address) {
            Address(
                property = property,
                street = street,
                town = town,
                area = area,
                postcode = postcode,
            )
        }
    )
</pre>

PR #103, opened 2022-12-05
EIP1-1878 - Turn down SFTP logging on int

NathanRussellValtech:
This PR removes the additional logging for the SFTP components that we put into `int` recently.


Neil-Massey (approved):
LGTM

PR #106, opened 2022-12-09
EIP1-2885 - Set batch job limit

NathanRussellValtech:
This PR sets a limit to the number of un-batched Print Requests that the batch job will pull from the database at a time (to prevent very very long running processes that span several cron invocations as described in EIP1-2885)

The limit is set at 5,000 in `application.yml` as proposed in EIP1-2885

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepository.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepository.kt
@@ -1,5 +1,6 @@
 package uk.gov.dluhc.printapi.database.repository
 
+import org.springframework.data.domain.Pageable
 import org.springframework.data.jpa.repository.JpaRepository
 import org.springframework.data.jpa.repository.Query
 import org.springframework.stereotype.Repository
@@ -16,10 +17,17 @@ interface CertificateRepository : JpaRepository<Certificate, UUID> {
 
     fun findByStatusAndPrintRequestsBatchId(certificateStatus: Status, batchId: String): List<Certificate>
 
-    fun findByStatus(certificateStatus: Status): List<Certificate>
+    fun findByStatusOrderByApplicationReceivedDateTimeAsc(certificateStatus: Status, pageable: Pageable): List<Certificate>
 
     fun findByGssCodeInAndSourceTypeAndSourceReference(gssCodes: List<String>, sourceType: SourceType, sourceReference: String): Certificate?
 
-    @Query("SELECT COUNT(*) FROM PrintRequestStatus s WHERE s.eventDateTime >= :startInstant AND s.eventDateTime <= :endInstant AND s.status = :status")
+    @Query(
+        value = """
+            SELECT COUNT(s) FROM PrintRequestStatus s
+            WHERE s.eventDateTime >= :startInstant 
+            AND s.eventDateTime <= :endInstant 
+            AND s.status = :status
+        """
+    )
     fun getPrintRequestStatusCount(startInstant: Instant, endInstant: Instant, status: Status): Int
 }
--- a/src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
@@ -2,9 +2,11 @@ package uk.gov.dluhc.printapi.service
 
 import mu.KotlinLogging
 import org.springframework.beans.factory.annotation.Value
+import org.springframework.data.domain.Pageable
 import org.springframework.stereotype.Service
 import uk.gov.dluhc.printapi.database.entity.Certificate
-import uk.gov.dluhc.printapi.database.entity.Status
+import uk.gov.dluhc.printapi.database.entity.Status.ASSIGNED_TO_BATCH
+import uk.gov.dluhc.printapi.database.entity.Status.PENDING_ASSIGNMENT_TO_BATCH
 import uk.gov.dluhc.printapi.database.repository.CertificateRepository
 import java.time.Clock
 import java.time.Instant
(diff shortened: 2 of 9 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepository.kt
```diff
@@ -16,10 +17,17 @@ interface CertificateRepository : JpaRepository<Certificate, UUID> {
 
     fun findByStatusAndPrintRequestsBatchId(certificateStatus: Status, batchId: String): List<Certificate>
 
-    fun findByStatus(certificateStatus: Status): List<Certificate>
+    fun findByStatusOrderByApplicationReceivedDateTimeAsc(certificateStatus: Status, pageable: Pageable): List<Certificate>
 
     fun findByGssCodeInAndSourceTypeAndSourceReference(gssCodes: List<String>, sourceType: SourceType, sourceReference: String): Certificate?
 
-    @Query("SELECT COUNT(*) FROM PrintRequestStatus s WHERE s.eventDateTime >= :startInstant AND s.eventDateTime <= :endInstant AND s.status = :status")
```
NathanRussellValtech, line 23: My IDE was complaining that `COUNT(*)` was not value JPQL (even though it was working fine). I changed it to `COUNT(s)` to keep my IDE quiet! 😁

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepository.kt
```diff
@@ -16,10 +17,17 @@ interface CertificateRepository : JpaRepository<Certificate, UUID> {
 
     fun findByStatusAndPrintRequestsBatchId(certificateStatus: Status, batchId: String): List<Certificate>
 
-    fun findByStatus(certificateStatus: Status): List<Certificate>
+    fun findByStatusOrderByApplicationReceivedDateTimeAsc(certificateStatus: Status, pageable: Pageable): List<Certificate>
```
NathanRussellValtech, line 20: The only way I could create a repository method that implements a limit with the `LIMIT` keyword in the SQL query is by passing in a `Pageable`. This means the caller (our service class) needs to know about and how to create a suitable `Pageable` which I'm not 100% happy about, but its the best I could do

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
```diff
@@ -17,28 +19,32 @@ private val logger = KotlinLogging.logger { }
 class CertificateBatchingService(
     private val idFactory: IdFactory,
     private val certificateRepository: CertificateRepository,
-    @Value("\${jobs.batch-print-requests.daily-limit}")
-    private val dailyLimit: Int,
+    @Value("\${jobs.batch-print-requests.daily-limit}") private val dailyLimit: Int,
+    @Value("\${jobs.batch-print-requests.batch-size}") private val batchSize: Int,
```
NathanRussellValtech, line 23: `batchSize` which originally came from `BatchPrintRequestsJob`
I've added `maxUnBatchedRecords`

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
```diff
@@ -17,28 +19,32 @@ private val logger = KotlinLogging.logger { }
...
         }
         return batches.keys
     }
 
-    fun batchCertificates(batchSize: Int): Map<String, List<Certificate>> {
-        val certificatesPendingAssignment = certificateRepository.findByStatus(Status.PENDING_ASSIGNMENT_TO_BATCH)
+    fun batchCertificates(): Map<String, List<Certificate>> {
+        val certificatesPendingAssignment = certificateRepository.findByStatusOrderByApplicationReceivedDateTimeAsc(
+            PENDING_ASSIGNMENT_TO_BATCH,
+            Pageable.ofSize(maxUnBatchedRecords)
```
NathanRussellValtech, line 40: This is where the service class needs to know about a `Pageable` 😢

src/main/resources/application.yml
```diff
@@ -40,6 +40,7 @@ jobs:
     name: "BatchPrintRequests"
     cron: "0 0/15 * * * *" # every 15 minutes, starting on the hour - see analysis and recommendations in EIP1-2515
     batch-size: 50
+    max-un-batched-records: 5000
```
NathanRussellValtech, line 43: 5000, as per the recommendation in EIP1-2885 .... perhaps I should put a comment after it referencing the ticket number so someone can understand the rationale for this number at some later date 🤔

src/test/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepositoryIntegrationTest.kt
```diff
@@ -329,23 +351,4 @@ internal class CertificateRepositoryIntegrationTest : IntegrationTest() {
             assertThat(actual).isEqualTo(2)
         }
     }
-
-    private fun assertCertificateRecursiveEqual(actual: Certificate, expected: Certificate) {
```
NathanRussellValtech, line 333: This was really complicated and I felt could be much more succinctly replaced with
```
            assertThat(actual)
                .usingRecursiveComparison()
                .ignoringFieldsMatchingRegexes(*IGNORED_FIELDS)
                .isEqualTo(expected)
```

src/main/resources/application.yml
NathanRussellValtech: Done

src/test/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepositoryIntegrationTest.kt
Neil-Massey: I remember being frustrated with the date comparisons not working due to rounding - glad you found ways to simplify 👍 

Nice trick:
```
Instant.now().minusSeconds(it.toLong()).truncatedTo(SECONDS)
```

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepository.kt
NathanRussellValtech: Unfortunately you cant put `LIMIT` in a JPQL query.
It (kinda) works if you make the query a native query, but then I think you need to make the method return a list of `Object` (or `Any` in kotlin 😁 ), and you have to mess around converting/casting/mapping (row mapper ?) the object yourself. It was messy and horrible

As mentioned, the only way I could get it working was with a `Pageable` which seems to be the way t'interwebs suggest you do it 😢

PR #107, opened 2022-12-12
EIP1-2833 - Add addressFormat and deliveryAddressType

NathanRussellValtech:
This PR updates the `print-api` `SendApplicationToPrintMessage` to:
* add `addressFormat`
* rename `deliveryMethod` to `deliveryAddressType`

And associated database and entity changes.

Neither field are sent to the Print Provider in the PSV file yet. This change allows for this in the future.

The `addressFormat` and `deliveryAddressType` enum values are consistent with the related changes in VCA


src/main/resources/db/changelog/ddl/0008_EIP1-2833_alter_delivery.xml
```diff
@@ -0,0 +1,48 @@
...
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+
+    <changeSet author="nathan.russell@valtech.com" id="0008_EIP1-2833_alter_delivery - add addressFormat column" context="ddl">
+
+        <addColumn tableName="delivery">
+            <column name="address_format" type="varchar(20)" value="UK">
```
NathanRussellValtech, line 11: Adds the new column, and sets all existing records to have the value `UK` (thats the only address format we support at the moment)

src/main/resources/db/changelog/ddl/0008_EIP1-2833_alter_delivery.xml
```diff
@@ -0,0 +1,48 @@
...
+
+        <rollback>
+            <dropColumn
+                tableName="delivery"
+                columnName="address_format"
+            />
+        </rollback>
+    </changeSet>
+
+    <changeSet author="nathan.russell@valtech.com" id="0008_EIP1-2833_alter_delivery - rename addressFormat column" context="ddl">
```
NathanRussellValtech, line 24: Rename `delivery_method` to `address_type`; and set all values to `REGISTERED` as thats the only value we support at the moment (previous column value was `DELIVERY`)

src/main/resources/db/changelog/ddl/0008_EIP1-2833_alter_delivery.xml
```diff
@@ -0,0 +1,48 @@
...
+
+        <rollback>
+            <dropColumn
+                tableName="delivery"
+                columnName="address_format"
+            />
+        </rollback>
+    </changeSet>
+
+    <changeSet author="nathan.russell@valtech.com" id="0008_EIP1-2833_alter_delivery - rename delivery_method column" context="ddl">
```
NathanRussellValtech, line 24: Rename delivery_method to address_type; and set all values to REGISTERED as thats the only value we support at the moment (previous column value was DELIVERY)

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -169,19 +169,22 @@ components:
...
         addressee:
           type: string
           maxLength: 255
           description: The addressee to be printed in addition to the delivery address
         address:
           $ref: '#/components/schemas/Address'
+        addressFormat:
+          $ref: '#/components/schemas/AddressFormat'
       required:
-        - deliveryOption
```
NathanRussellValtech, line 181: I think `deliveryOption` was a very early version of this! Should have been `deliveryClass`

Neil-Massey: Without versioning the SQS messages and having different listeners for the different models we are in a position where messages will fail to be processed.

Deployment Issue
Release Print first to support version `1.9.0` of spec - will fail to process any messages from spec `1.8.0`
Release VCA first to send  version `1.9.0` of spec - print will fail to process as missing required properties from spec `1.8.0`

At least in this case messages end up on DLQ and can be Redrived.

Neil-Massey (approved):
LGTM - Nice tidy up of the entities.  Commented on the deployment issue due to not supporting multiple model versions - can discuss offline.

src/main/resources/db/changelog/ddl/0008_EIP1-2833_alter_delivery.xml
adacayi: Minor comment, @NathanRussellValtech. 
We have set this column size to 50 in VCA, so it might be good to set it to 50 here to be consistent. 
```

  <changeSet author="abdullah.sanver@valtech.com" id="0072_EIP1-2819_add_certificate_delivery" context="ddl">
        <createTable tableName="certificate_delivery">
            <column name="id" type="uuid">
                <constraints primaryKey="true" nullable="false" primaryKeyName="certificate_delivery_pkey"/>
            </column>
            <column name="voter_card_application_id" type="uuid">
                <constraints references="voter_card_application(id)" foreignKeyName="fk_certificate_delivery_voter_card_application" nullable="false" />
            </column>
            <column name="line1" type="varchar(255)">
                <constraints nullable="false"/>
            </column>
            <column name="line2" type="varchar(255)" />
...
            <column name="address_format" type="varchar(50)">
                <constraints nullable="false"/>
            </column>
            <column name="delivery_address_type" type="varchar(50)">
                <constraints nullable="false"/>
            </column>
```

src/main/resources/db/changelog/ddl/0008_EIP1-2833_alter_delivery.xml
```diff
@@ -0,0 +1,48 @@
...
+        <update tableName="delivery">
+            <column name="delivery_method" value="REGISTERED" />
+        </update>
+
+        <renameColumn
+            tableName="delivery"
+            oldColumnName="delivery_method"
+            newColumnName="address_type"
+            columnDataType="varchar(20)"
```
adacayi, line 33: Similar comment as above, @NathanRussellValtech. 
We have set this column size to 50 in VCA, so it might be good to set it to 50 here as well for consistency.

NathanRussellValtech: Yeah, OK, good shout. Will do 👍

src/main/resources/db/changelog/ddl/0008_EIP1-2833_alter_delivery.xml
NathanRussellValtech: Done; thanks 👍

src/main/resources/db/changelog/ddl/0008_EIP1-2833_alter_delivery.xml
adacayi: On a further note, I think it will also be better if we have the same column name as VCA here (`delivery_address_type`). We have delivery_address_type in the VCA table and I saw the sqs message in this PR also has the name `DeliveryAddressType`, so it will be easier to understand the purpose of the column and be more consistent if we have the same naming here as well.

NathanRussellValtech: I've done as suggested, but I have left the rollback at `varchar(20)` because thats what it is before this change (ie. rolling back to the same state that it is beforehand)

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -169,19 +169,22 @@ components:
       properties:
         deliveryClass:
           $ref: '#/components/schemas/DeliveryClass'
-        deliveryMethod:
-          $ref: '#/components/schemas/DeliveryMethod'
+        addressType:
```
adacayi, line 172: Minor comment here as well, @NathanRussellValtech . 
It was easier for me to understand what the `addressType` here refers to when I saw the ref pointing to `DeliveryAddressType`. I think it will be easier to understand if we have this field named as `deliveryAddressType` as well.

adacayi (approved):
LGTM, @NathanRussellValtech .
Added some minor suggestions around data size and column/field names.

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
NathanRussellValtech: Done 👍

PR #113, opened 2023-01-04
EIP1-3144: Map English Ero value if certificate language is Welsh and no Welsh details exist

adacayi:
This PR is about sending English ERO details as Welsh ERO details in the PSV file where certificate language is Welsh, but we don't have ERO details for Welsh.

**UPDATE**

The implementation has changed to map the English ERO details to Welsh ERO details in the `PrintRequestMapper`, as agreed with @Neil-Massey.

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/mapper/ElectoralRegistrationOfficeMapper.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/mapper/ElectoralRegistrationOfficeMapper.kt
@@ -1,0 +1,39 @@
+package uk.gov.dluhc.printapi.mapper
+
+import org.mapstruct.Mapper
+import org.mapstruct.Mapping
+import uk.gov.dluhc.printapi.database.entity.ElectoralRegistrationOffice
+import uk.gov.dluhc.printapi.dto.EroContactDetailsDto
+import uk.gov.dluhc.printapi.messaging.models.CertificateLanguage
+import uk.gov.dluhc.printapi.messaging.models.CertificateLanguage.CY
+import uk.gov.dluhc.printapi.messaging.models.CertificateLanguage.EN
+
+private const val ELECTORAL_REGISTRATION_OFFICER_EN = "Electoral Registration Officer"
+private const val ELECTORAL_REGISTRATION_OFFICER_CY = "Swyddog Cofrestru Etholiadol"
+
+@Mapper
+abstract class ElectoralRegistrationOfficeMapper {
+    @Mapping(target = "name", expression = "java( getName(language) )")
+    protected abstract fun toElectoralRegistrationOfficeFromNotNullContactDetails(
+        eroContactDetails: EroContactDetailsDto,
+        language: CertificateLanguage
+    ): ElectoralRegistrationOffice
+
+    fun toElectoralRegistrationOffice(
+        eroContactDetails: EroContactDetailsDto?,
+        language: CertificateLanguage
+    ): ElectoralRegistrationOffice? {
+        if (eroContactDetails == null) {
+            return null
+        }
+
+        return toElectoralRegistrationOfficeFromNotNullContactDetails(eroContactDetails, language)
+    }
+
+    protected fun getName(language: CertificateLanguage): String {
+        return when (language) {
+            CY -> ELECTORAL_REGISTRATION_OFFICER_CY
+            EN -> ELECTORAL_REGISTRATION_OFFICER_EN
+        }
+    }
+}
--- a/src/main/kotlin/uk/gov/dluhc/printapi/service/PrintRequestsFileProducer.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/service/PrintRequestsFileProducer.kt
@@ -81,7 +81,7 @@ class PrintRequestsFileProducer {
             arrayOf(
                 requestId,
                 issuingAuthorityEn,
-                getWelshValue(issuingAuthorityCy, issuingAuthorityEn),
+                issuingAuthorityCy ?: "",
                 issueDate.format(DATE_FORMATTER),
                 suggestedExpiryDate.format(DATE_FORMATTER),
                 requestDateTime.format(DATE_TIMESTAMP_FORMATTER),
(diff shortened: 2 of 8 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/service/PrintRequestsFileProducer.kt
```diff
@@ -111,19 +111,30 @@ class PrintRequestsFileProducer {
...
+                getWelshValue(eroDeliveryLocalityCy, eroDeliveryLocalityEn),
+                getWelshValue(eroDeliveryTownCy, eroDeliveryTownEn),
+                getWelshValue(eroDeliveryAreaCy, eroDeliveryAreaEn),
+                getWelshValue(eroDeliveryPostcodeCy, eroDeliveryPostcodeEn)
             )
         }
 
+    private fun PrintRequest.getWelshValue(welshValue: String?, englishValue: String?): String {
+        val isWelshAndWelshPropertiesMissing =
+            certificateLanguage == PrintRequest.CertificateLanguage.CY && eroNameCy == null
```
adacayi, line 129: I checked if the `eroNameCy` is null or not to determine if the Welsh details are missing or not. The reason for that is, as can be seen in Ero English details, some properties can be null like `eroDeliveryTownEn`, `eroDeliveryLocalityEn`.

Assume we had this implementation instead, checking if Welsh properties are missing by checking the `welshValue` is null or not, instead of checking the `eroNameCy`.

<pre>
val isWelshAndWelshPropertiesMissing =
            certificateLanguage == PrintRequest.CertificateLanguage.CY && <b>welshValue</b> == null
</pre>

In that case, if we had an actual Welsh address where some optional fields like `eroDeliveryTownCy` and `eroDeliveryLocalityCy` are null and `eroDeliveryTownEn`  and `                eroDeliveryLocalityEn ` with non null values, than those English values would be mixed with the Welsh address.

I added the `should write file to stream given print request given full English properties and partially missing Welsh properties and certificate language is Welsh` test to specifically check that doesn't happen.

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintRequestsFileProducer.kt
```diff
@@ -111,19 +111,30 @@ class PrintRequestsFileProducer {
...
+                getWelshValue(eroPhoneNumberCy, eroPhoneNumberEn),
+                getWelshValue(eroEmailAddressCy, eroEmailAddressEn),
+                getWelshValue(eroWebsiteCy, eroWebsiteEn),
+                getWelshValue(eroDeliveryStreetCy, eroDeliveryStreetEn),
+                getWelshValue(eroDeliveryPropertyCy, eroDeliveryPropertyEn),
+                getWelshValue(eroDeliveryLocalityCy, eroDeliveryLocalityEn),
+                getWelshValue(eroDeliveryTownCy, eroDeliveryTownEn),
+                getWelshValue(eroDeliveryAreaCy, eroDeliveryAreaEn),
+                getWelshValue(eroDeliveryPostcodeCy, eroDeliveryPostcodeEn)
             )
```
Neil-Massey, line 124: This looks wrong as it appears that we could get a mix of Welsh ERO sourced data and English ERO sourced data - e.g. street populated from `eroDeliveryStreetCy` but property from `eroDeliveryPropertyEn`
I guess in practice if the data is missing from the Welsh translation, it is likely to be missing from the English dataset too.  But would suggest it might be better if a single decision is made to use the Welsh ERO data or the English ERO data.  Maybe the `PrintRequestMapper` could populate the `PrintRequest` with the appropriate English or Welsh data?

adacayi: Thanks, @Neil-Massey , for your comment.
I tried to explain the implementation in the above comment, where I prevent mixing of the English and Welsh values by checking if the Welsh values are missing if `eroNameCy == null`. 

There is also a test covering that scenario as well: `should write file to stream given print request given full English properties and partially missing Welsh properties and certificate language is Welsh`

In that test we expect Welsh values, where some are null being set to "" instead of filled with corresponding English values which are non-null.

adacayi: We wanted to implement the logic in this mapper, to only affect what is going to be written in the psv file, but retaining all the remaining implementation the same as before.

Neil-Massey: Should the database store the `eroWelsh` details that are included in the `PrintRequest`?  (I mean indicate what is sent to the printer?)

adacayi: @NathanRussellValtech and I discussed this before starting the ticket and decided not to store English ERO details as Welsh details in the database in case Welsh details are missing for that Ero.

Neil-Massey: OK - thanks for the clarification @adacayi - when I did the original implementation I expected the PSV writer to simply write out what is stored in the database so was a little surprised at the code 😄 .

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintRequestsFileProducer.kt
```diff
@@ -111,19 +111,30 @@ class PrintRequestsFileProducer {
...
+                getWelshValue(eroNameCy, eroNameEn),
+                getWelshValue(eroPhoneNumberCy, eroPhoneNumberEn),
+                getWelshValue(eroEmailAddressCy, eroEmailAddressEn),
+                getWelshValue(eroWebsiteCy, eroWebsiteEn),
+                getWelshValue(eroDeliveryStreetCy, eroDeliveryStreetEn),
+                getWelshValue(eroDeliveryPropertyCy, eroDeliveryPropertyEn),
+                getWelshValue(eroDeliveryLocalityCy, eroDeliveryLocalityEn),
+                getWelshValue(eroDeliveryTownCy, eroDeliveryTownEn),
+                getWelshValue(eroDeliveryAreaCy, eroDeliveryAreaEn),
+                getWelshValue(eroDeliveryPostcodeCy, eroDeliveryPostcodeEn)
```
vishalgupta-valtech, line 123: I am confused with this implementation.  
e.g If an ero doesn't have a street in welsh, but has a street in english ero, then the returned address will have something like this

`Welsh Ero,
Street of English ero,
Welsh postcode
`

May be Line 114 to 123 should be done in if else loop.  i.e. if ero welsh doesn't exist, then copy over all details from english ero.

adacayi: Thanks, @vishalgupta-valtech , for the comment. 
I can understand the confusion.
Let me try to explain.
The way `getWelshValue` is implemented is, first it decides if we want to replace the existing Welsh value with the English one by checking if the certificate language is Welsh and somehow Welsh details don't exist. I decided to check the eroNameCy field to determine if the Welsh details exist or not, because I assume this value should always be there if the ERO has Welsh details (unlike the `eroDeliveryAreaCy` etc.). Then in that case we return the corresponding English value. Otherwise, it will be either the existing Welsh value or empty string if the value is null.

<pre>
    private fun PrintRequest.getWelshValue(welshValue: String?, englishValue: String?): String {
        val isWelshAndWelshPropertiesMissing =
            certificateLanguage == PrintRequest.CertificateLanguage.CY && <b>eroNameCy == null</b>

        if (isWelshAndWelshPropertiesMissing) {
            return englishValue ?: ""
        }

        return welshValue ?: ""
    }
</pre>

We might move the logic to `PrintRequestMapper` class. I am waiting @Neil-Massey's input on this.

vishalgupta-valtech (commented):
Please could you revisit the `logic` and map this ero welsh/english address elsewhere in some mapper if suitable?

src/main/kotlin/uk/gov/dluhc/printapi/mapper/ElectoralRegistrationOfficeMapper.kt
```diff
@@ -0,0 +1,39 @@
...
+import uk.gov.dluhc.printapi.messaging.models.CertificateLanguage.CY
+import uk.gov.dluhc.printapi.messaging.models.CertificateLanguage.EN
+
+private const val ELECTORAL_REGISTRATION_OFFICER_EN = "Electoral Registration Officer"
+private const val ELECTORAL_REGISTRATION_OFFICER_CY = "Swyddog Cofrestru Etholiadol"
+
+@Mapper
+abstract class ElectoralRegistrationOfficeMapper {
+    @Mapping(target = "name", expression = "java( getName(language) )")
+    protected abstract fun toElectoralRegistrationOfficeFromNotNullContactDetails(
```
adacayi, line 17: Unfortunately, with the MapStruct implemantation, this method doesn't return null when the passed `eroContactDetails` is passed as null. Instead, it generates an instance of the `ElectoralRegistrationOffice`, with the field values set to null. That's why I added the `toElectoralRegistrationOffice`, which calls this method if `eroContactDetails` is not null and otherwise returns null.

Happy to implement a better solution, if there is any.

Here is the code generated by MapStruct:

<pre>
 @Override
    protected ElectoralRegistrationOffice toElectoralRegistrationOfficeFromNotNullContactDetails(EroContactDetailsDto eroContactDetails, CertificateLanguage language) {
        if ( eroContactDetails == null && language == null ) {
            return null;
        }

        ElectoralRegistrationOffice electoralRegistrationOffice = new ElectoralRegistrationOffice();

        if ( eroContactDetails != null ) {
            electoralRegistrationOffice.setPhoneNumber( eroContactDetails.getPhoneNumber() );
            electoralRegistrationOffice.setEmailAddress( eroContactDetails.getEmailAddress() );
            electoralRegistrationOffice.setWebsite( eroContactDetails.getWebsite() );
            electoralRegistrationOffice.setAddress( addressDtoToAddress( eroContactDetails.getAddress() ) );
        }
        electoralRegistrationOffice.setName( getName(language) );

        return electoralRegistrationOffice;
    }

</pre>

Neil-Massey (approved):
LGTM - Thanks @abdullah-sanver-valtech for updating approach to handle Welsh ERO properties in the `PrintRequestMapper`

PR #114, opened 2023-01-17
EIP1-2467 Refactor to support certificates with several print requests

Neil-Massey:
Original implementation had data structure to support multiple print requests per Certificate but the code was simplified to act against the current (only) print request in the various flows.  This PR updates the codebase to select the relevant Print Request based on one of its status, its assigned batch or its request id.  Refactored `getCurrentPrintRequest()` to be private in the Certificate.

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
@@ -104,30 +104,32 @@ class Certificate(
         printRequests.filter { it.getCurrentStatus().status == printRequestStatus }
 
     fun addPrintRequestToBatch(batchId: String) {
-        getPrintRequestsByStatus(Status.PENDING_ASSIGNMENT_TO_BATCH).forEach {
-            it.addPrintRequestStatus(
-                PrintRequestStatus(
-                    status = Status.ASSIGNED_TO_BATCH,
-                    eventDateTime = Instant.now(),
-                    message = null
+        processPrintRequestUpdate {
+            getPrintRequestsByStatus(Status.PENDING_ASSIGNMENT_TO_BATCH).forEach {
+                it.addPrintRequestStatus(
+                    PrintRequestStatus(
+                        status = Status.ASSIGNED_TO_BATCH,
+                        eventDateTime = Instant.now(),
+                        message = null
+                    )
                 )
-            )
-            it.batchId = batchId
+                it.batchId = batchId
+            }
         }
-        assignStatus()
     }
 
     fun addSentToPrintProviderEventForBatch(batchId: String) {
-        getPrintRequestsByBatchId(batchId).forEach {
-            it.addPrintRequestStatus(
-                PrintRequestStatus(
-                    status = Status.SENT_TO_PRINT_PROVIDER,
-                    eventDateTime = Instant.now(),
-                    message = null
+        processPrintRequestUpdate {
+            getPrintRequestsByBatchId(batchId).forEach {
+                it.addPrintRequestStatus(
+                    PrintRequestStatus(
+                        status = Status.SENT_TO_PRINT_PROVIDER,
+                        eventDateTime = Instant.now(),
+                        message = null
+                    )
                 )
-            )
+            }
         }
-        assignStatus()
     }
 
     fun addReceivedByPrintProviderEventForBatch(
@@ -135,16 +137,17 @@ class Certificate(
         eventDateTime: Instant,
         message: String?
     ) {
-        getPrintRequestsByBatchId(batchId).forEach {
-            it.addPrintRequestStatus(
-                PrintRequestStatus(
-                    status = Status.RECEIVED_BY_PRINT_PROVIDER,
-                    eventDateTime = eventDateTime,
-                    message = message
+        processPrintRequestUpdate {
+            getPrintRequestsByBatchId(batchId).forEach {
+                it.addPrintRequestStatus(
+                    PrintRequestStatus(
+                        status = Status.RECEIVED_BY_PRINT_PROVIDER,
+                        eventDateTime = eventDateTime,
+                        message = message
+                    )
                 )
-            )
+            }
         }
-        assignStatus()
     }
 
     fun requeuePrintRequestForBatch(
@@ -153,18 +156,19 @@ class Certificate(
         message: String?,
         newRequestId: String
     ) {
-        getPrintRequestsByBatchId(batchId).forEach {
-            it.addPrintRequestStatus(
-                PrintRequestStatus(
-                    status = Status.PENDING_ASSIGNMENT_TO_BATCH,
-                    eventDateTime = eventDateTime,
-                    message = message
+        processPrintRequestUpdate {
+            getPrintRequestsByBatchId(batchId).forEach {
+                it.addPrintRequestStatus(
+                    PrintRequestStatus(
+                        status = Status.PENDING_ASSIGNMENT_TO_BATCH,
+                        eventDateTime = eventDateTime,
+                        message = message
+                    )
                 )
-            )
-            it.batchId = null
-            it.requestId = newRequestId
+                it.batchId = null
+                it.requestId = newRequestId
+            }
         }
-        assignStatus()
     }
 
     fun addPrintRequestEvent(
@@ -173,15 +177,21 @@ class Certificate(
         eventDateTime: Instant,
         message: String?
     ) {
-        getPrintRequestsByRequestId(requestId).forEach {
-            it.addPrintRequestStatus(
-                PrintRequestStatus(
-                    status = status,
-                    eventDateTime = eventDateTime,
-                    message = message
+        processPrintRequestUpdate {
+            getPrintRequestsByRequestId(requestId).forEach {
+                it.addPrintRequestStatus(
+                    PrintRequestStatus(
+                        status = status,
+                        eventDateTime = eventDateTime,
+                        message = message
+                    )
                 )
-            )
+            }
         }
+    }
+
+    private fun processPrintRequestUpdate(update: () -> Unit) {
+        update.invoke()
         assignStatus()
     }
 
--- a/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
@@ -103,18 +103,16 @@ class Certificate(
     fun getPrintRequestsByStatus(printRequestStatus: Status) =
         printRequests.filter { it.getCurrentStatus().status == printRequestStatus }
 
-    fun addPrintRequestToBatch(batchId: String) {
+    fun addPrintRequestToBatch(printRequest: PrintRequest, batchId: String) {
         processPrintRequestUpdate {
-            getPrintRequestsByStatus(Status.PENDING_ASSIGNMENT_TO_BATCH).forEach {
-                it.addPrintRequestStatus(
-                    PrintRequestStatus(
-                        status = Status.ASSIGNED_TO_BATCH,
-                        eventDateTime = Instant.now(),
-                        message = null
-                    )
+            printRequest.addPrintRequestStatus(
+                PrintRequestStatus(
+                    status = Status.ASSIGNED_TO_BATCH,
+                    eventDateTime = Instant.now(),
+                    message = null
                 )
-                it.batchId = batchId
-            }
+            )
+            printRequest.batchId = batchId
         }
     }
 
--- a/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
@@ -66,6 +66,11 @@ class Certificate(
     @field:NotNull
     var suggestedExpiryDate: LocalDate = issueDate.plusYears(10),
 
+    /**
+     * Certificate status corresponds to the current status of the most recent
+     * [PrintRequest], based on the requestDateTime that is included in the
+     * [uk.gov.dluhc.printapi.messaging.models.SendApplicationToPrintMessage].
+     */
     @field:NotNull
     @Enumerated(EnumType.STRING)
     var status: Status? = null,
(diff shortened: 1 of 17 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
```diff
@@ -95,30 +100,98 @@ class Certificate(
...
+    fun addSentToPrintProviderEventForBatch(batchId: String) {
+        getPrintRequestsByBatchId(batchId).forEach {
+            it.addPrintRequestStatus(
+                PrintRequestStatus(
+                    status = Status.SENT_TO_PRINT_PROVIDER,
+                    eventDateTime = Instant.now(),
+                    message = null
+                )
+            )
+            it.batchId = batchId
```
adacayi, line 129: @Neil-Massey , is this needed here, since we iterate through the printRequests with the batchId equal to this `batchId`?

Neil-Massey: Good spot - will remove 👍

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
```diff
@@ -95,30 +100,97 @@ class Certificate(
...
+            it.addPrintRequestStatus(
+                PrintRequestStatus(
+                    status = Status.ASSIGNED_TO_BATCH,
+                    eventDateTime = Instant.now(),
+                    message = null
+                )
+            )
+            it.batchId = batchId
+        }
+        assignStatus()
```
adacayi, line 117: Do you think it will be better to add an `addPrintRequestStatus` method in this class, which adds the `PrintRequestStatus` to the relevant `PrintRequest` and also calls the `assignStatus` method? This way we won't need to care about updating the status of the certificate each time we add a new `PrintRequestStatus`.
<pre>
getPrintRequestsByStatus(Status.PENDING_ASSIGNMENT_TO_BATCH).forEach {
     <b>       addPrintRequestStatus( it, </b>
                PrintRequestStatus(
                    status = Status.ASSIGNED_TO_BATCH,
                    eventDateTime = Instant.now(),
                    message = null
                )
            )
            it.batchId = batchId
        }

  <b>  private fun addPrintRequestStatus(printRequest: PrintRequest, printRequestStatus: PrintRequestStatus) { </b>
        printRequest.addPrintRequestStatus(printRequestStatus)
        assignStatus()
    }
</pre>

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
```diff
@@ -66,6 +66,11 @@ class Certificate(
     @field:NotNull
     var suggestedExpiryDate: LocalDate = issueDate.plusYears(10),
 
+    /**
+     * Certificate status corresponds to the current status of the most recent
```
adacayi, line 70: I assume from the implementation, this will be the status value of the last PrintRequestStatus that is added to certificate's print requests. If that is the case and we assume we can receive a new print request without the previous one being completely processed, we might end up with a status of a prior request when it is updated.

Neil-Massey:
```
    private fun assignStatus() {
        val currentPrintRequest = getCurrentPrintRequest()
        val currentStatus = currentPrintRequest.getCurrentStatus()
        status = currentStatus.status
    }
```
The status relates to the current print request that selects the PrintRequest with the most recent `requestDateTime` and then for that print request the latest status.

So I don't the scenario you described will occur.  Any "previous" print requests that are updated do not affect the certificate status.

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
```diff
@@ -42,10 +42,7 @@ class CertificateBatchingService(
 
         return limitCertificates(certificatesPendingAssignment).chunked(batchSize).associate { batchOfCertificates ->
             val batchId = idFactory.batchId()
-            batchId to batchOfCertificates.onEach {
-                it.getCurrentPrintRequest().batchId = batchId
-                it.addStatus(ASSIGNED_TO_BATCH)
-            }
+            batchId to batchOfCertificates.onEach { it.addPrintRequestToBatch(batchId) }
```
adacayi, line 45: In the above code when querying certificates by status, we might miss certificates which had a new print request with `PENDING_ASSIGNMENT_TO_BATCH`, but then received a response for a prior print request, assume `VALIDATED_BY_PRINT_PROVIDER`, which will update the status of certificate to `VALIDATED_BY_PRINT_PROVIDER`, hence not being returned by the `certificateRepository.findByStatusOrderByApplicationReceivedDateTimeAsc` method.

adacayi: There is also one more issue here. Assume somehow we have one `printRequest` with status `PENDING_ASSIGNMENT_TO_BATCH` and before it being assigned to a batch we receive a new `printRequest` and its status also becomes `PENDING_ASSIGNMENT_TO_BATCH`.
With the `it.addPrintRequestToBatch(batchId)`, we will be assigning both of them to the same batch, but it will count as one in our limiting logic, since we iterate over certificates and assume we assign one request per certificate to the batch (which was true for the previous implementation), but that might no longer be true.

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
Neil-Massey: Refactored to have separate private method.

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
adacayi: @Neil, We might need to change the logic here as well in the `ProcessPrintBatchService.processBatch`.

<pre>
 @Transactional
    fun processBatch(batchId: String) {
        val certificates = certificateRepository.findByStatusAndPrintRequestsBatchId(ASSIGNED_TO_BATCH, batchId)
        if (certificates.isEmpty()) {
            throw EmptyBatchException(batchId, ASSIGNED_TO_BATCH)
        }
        val fileContents = printFileDetailsFactory.createFileDetailsFromCertificates(batchId, certificates)
        val sftpInputStream = sftpZipInputStreamProvider.createSftpInputStream(fileContents)
        val sftpFilename = filenameFactory.createZipFilename(batchId, certificates.size)
        sftpService.sendFile(sftpInputStream, sftpFilename)
        updateCertificates(batchId, certificates)
    }

</pre>

We are querying by the status in the Certificate, which might have a `ASSIGNED_TO_BATCH` print request, but that might not be status of the Certificate if there was a newer print request added to that certificate, which is not assigned to a batch yet. 

I think we need to query print requests directly from the printRequestRepository based on their status.

Also when generating file details with the  `printFileDetailsFactory.createFileDetailsFromCertificates` we need to take into account that there might be more than one print request for a certificate.

<pre>
  fun createFileDetailsFromCertificates(batchId: String, certificates: List<Certificate>): FileDetails {
        val fileContents = createFromCertificates(certificates)
        return FileDetails(
      </b>      printRequestsFilename = filenameFactory.createPrintRequestsFilename(batchId, certificates.size),</b>
            printRequests = fileContents.printRequests,
            photoLocations = fileContents.photoLocations
        )
    }

</pre>

Neil-Massey: `certificateRepository.findByStatusOrderByApplicationReceivedDateTimeAsc` probably should be refactored as it only handles getting `PENDING_ASSIGNMENT_TO_BATCH` and that's the first status - the code I think is fine and the situation you describe I don't think is possible.

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
```diff
@@ -60,20 +58,20 @@ class PrintResponseProcessingService(
 
     private fun processBatchCertificates(batchResponse: BatchResponse, certificates: List<Certificate>) {
```
adacayi, line 59: @Neil-Massey, I think this method will work properly for multiple `PrintRequests`, but the way we query certificates and pass into this method needs to change.

Here is the code that calls this method:

<pre>
    @Transactional
    fun processBatchResponses(batchResponses: List<BatchResponse>) {
        batchResponses.forEach { batchResponse ->
            val certificates =
               <b> certificateRepository.findByStatusAndPrintRequestsBatchId(
                    SENT_TO_PRINT_PROVIDER,
                    batchResponse.batchId
                )</b>
            processBatchCertificates(batchResponse, certificates)
            certificateRepository.saveAll(certificates)
        }
    }
</pre>

As you can see this queries certificates by their status not querying certificates containing print requests with status `SENT_TO_PRINT_PROVIDER`. So even if there is a print request with the status `SENT_TO_PRINT_PROVIDER` for a given certificate the status of the certificate itself might not reflect it, because that print request status might not be the last status for that certificate. So we lose the updates on that print request with the current implementation.

I think we should query based on print request status above, instead of certificate status to get the certificates to be updated.

Neil-Massey: Agree this one could run into issues processing multiple print requests in parallel.  There is no status on the PrintRequest - am more inclined to remove the status from the initial query.

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
Neil-Massey:
```
With the it.addPrintRequestToBatch(batchId) code is executed, we will be assigning both of them to the same batch, but it will count as one in our limiting logic, since we iterate over certificates and assume we assign one request per certificate to the batch (which was true for the previous implementation), but that might no longer be true.
```
Yes, I had considered this as an extreme edge case as when discussed with Nathan there was no expectation that the Admin will be making multiple re-print requests within a batching window.  (In effect we expect reprints to be occasional and not fired at the system at a high rate). If this does turn out to be a real issue the simplest solution is to reduce the threshold until the limiting code is fixed.

cc  @NathanRussellValtech - in case you think this needs to be included as part of this ticket

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintResponseProcessingService.kt
Neil-Massey: Refactored to return certificates based on the Print Request status and batch.

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
adacayi: Thanks, @Neil-Massey ,
I saw the assignStatus implementation, then going back and forth with the other code, I forgot how it worked. You're right. You can ignore this comment.

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
adacayi: Thanks, @Neil-Massey , for the huddle we had. 

✅ For the first issue I mentioned, I agree that this won't actually happen and I raised it because I misinterpreted the status calculation on the certificate. So it is all good.

✅ For the second one about having 2 requests assigned to a batch, I agree that exceeding the limit because of this is a very extreme case and since the implementation to prevent it is also complicated, I am fine with the current way we do it.

✅ For the third one about `ProcessPrintBatchService.processBatch`, you addressed it as well. Thanks.

Lastly, about the `printRequestsFilename = filenameFactory.createPrintRequestsFilename(batchId, certificates.size)`. Seems like it will be an easy fix to calculate the request size for sending to print by going through the print requests in the certificates with the correct status.

<pre>
    fun createFileDetailsFromCertificates(batchId: String, certificates: List<Certificate>): FileDetails {
        val fileContents = createFromCertificates(certificates)
        val sizeOfAssignedToBatchPrintRequests = certificates.flatMap { it.printRequests }
            .filter { it.getCurrentStatus().status == Status.ASSIGNED_TO_BATCH }.size
        return FileDetails(

            printRequestsFilename = filenameFactory.createPrintRequestsFilename(batchId, sizeOfAssignedToBatchPrintRequests),
            printRequests = fileContents.printRequests,
            photoLocations = fileContents.photoLocations
        )
    }
</pre>

Thanks, again.

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepository.kt
```diff
@@ -31,3 +31,10 @@ interface CertificateRepository : JpaRepository<Certificate, UUID> {
     )
     fun getPrintRequestStatusCount(startInstant: Instant, endInstant: Instant, status: Status): Int
 }
+
+object CertificateRepositoryExtensions {
+    fun CertificateRepository.findByPrintRequestStatusAndBatchId(status: Status, batchId: String): List<Certificate> {
+        return findByPrintRequestsBatchId(batchId)
+            .filter { it.printRequests.stream().anyMatch { printRequest -> printRequest.getCurrentStatus().status == status } }
```
adacayi, line 38: Minor, @Neil-Massey. Seems like there is a slightly shorter version of this in Kotlin without using `.stream()`:

<pre>
    fun CertificateRepository.findByPrintRequestStatusAndBatchId(status: Status, batchId: String): List<Certificate> {
        return findByPrintRequestsBatchId(batchId)
            .filter { <b>it.printRequests.any {</b> printRequest -> printRequest.getCurrentStatus().status == status } }
    }
</pre>

adacayi (approved):
Thanks, @Neil-Massey , for addressing the comments.
LGTM.

There is one minor issue with the size value of the fileName sent to print provider, in case multiple requests are sent to print for the same certificate in the same batch, which we discussed.
 
Also added a minor suggestion for the `findByPrintRequestStatusAndBatchId` method.

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepository.kt
Neil-Massey: Done

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
Neil-Massey: Refactored to filename factory 👍

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintFileDetailsFactory.kt
```diff
@@ -15,7 +17,7 @@ class PrintFileDetailsFactory(
     fun createFileDetailsFromCertificates(batchId: String, certificates: List<Certificate>): FileDetails {
         val fileContents = createFromCertificates(certificates)
         return FileDetails(
-            printRequestsFilename = filenameFactory.createPrintRequestsFilename(batchId, certificates.size),
+            printRequestsFilename = filenameFactory.createPrintRequestsFilename(batchId, certificates),
             printRequests = fileContents.printRequests,
```
adacayi, line 21: @Neil-Massey , I think we need to limit the print requests here to the ones that are in `ASSIGNED_TO_BATCH_STATUS`.

adacayi (changes requested):
Hi @Neil-Massey ,

I realized that we need to make a change to the `createFileDetailsFromCertificates` method in the `PrintFileDetailsFactory`, where we set the requests for the `FileDetails`. 

Currently we are putting all the printRequests there, but we need to only put the ones with the current status `ASSIGNED_TO_BATCH`.

UPDATE

---
This can be ignored, since there is an existing filtering based on the print request status.

Neil-Massey:
Thanks @abdullah-sanver-valtech for the thorough review - gives us confidence that any issues remaining are likely very edge-case.

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintFileDetailsFactory.kt
adacayi: @Neil-Massey , you're right. Ignore my comments on this.

adacayi (approved):

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
```diff
@@ -42,10 +42,7 @@ class CertificateBatchingService(
 
         return limitCertificates(certificatesPendingAssignment).chunked(batchSize).associate { batchOfCertificates ->
             val batchId = idFactory.batchId()
```
adacayi, line 44: Hi @Neil-Massey,

As we agreed, we can ignore the issue of exceeding the daily limit because of multiple print requests for the same certificate within the same batch. 

But, there is another issue. We can exceed the batch size limit for those certificates containing multiple print requests included in a full batch with 50 (batch-size) certificates. So for that batch, there will be more than 50 print request and the batch-size limit will be exceeded. In that case the batch might be rejected all together from the print provider.
 
I think this is a far more likely issue considering the batch size being 50.

I have a suggested solution like this:
<pre>
fun batchCertificates(): Map<String, List<Certificate>> {
        val certificatesPendingAssignment = certificateRepository.findByStatusOrderByApplicationReceivedDateTimeAsc(
            PENDING_ASSIGNMENT_TO_BATCH,
            Pageable.ofSize(maxUnBatchedRecords)
        )

        val printRequestsPendingAssignment = certificatesPendingAssignment.flatMap { it.printRequests }
            .filter { it.getCurrentStatus().status == PENDING_ASSIGNMENT_TO_BATCH }

        return limitPrintRequests(printRequestsPendingAssignment).chunked(batchSize).associate { batchOfPrintRequests ->
            val batchId = idFactory.batchId()
            batchId to batchOfPrintRequests.map { printRequest ->
                val certificate = certificatesPendingAssignment.first { it.printRequests.contains(printRequest) }
                certificate.addPrintRequestToBatch(printRequest, batchId)
                certificate
            }
        }
    }

    private fun limitPrintRequests(printRequestsPendingAssignmentToBatch: List<PrintRequest>): List<PrintRequest> {
        val startOfDay = Instant.now(clock).truncatedTo(DAYS)
        val endOfDay = startOfDay.plus(1, DAYS).minusSeconds(1)
        val countOfRequestsSentToPrintProvider =
            certificateRepository.getPrintRequestStatusCount(startOfDay, endOfDay, ASSIGNED_TO_BATCH)

        if ((printRequestsPendingAssignmentToBatch.size + countOfRequestsSentToPrintProvider) > dailyLimit) {
            logDailyLimit(printRequestsPendingAssignmentToBatch, countOfRequestsSentToPrintProvider, endOfDay)
            return printRequestsPendingAssignmentToBatch.subList(0, dailyLimit - countOfRequestsSentToPrintProvider)
        }

        return printRequestsPendingAssignmentToBatch
    }

    private fun logDailyLimit(
        printRequestsPendingAssignmentToBatch: List<PrintRequest>,
        countOfRequestsSentToPrintProvider: Int,
        endOfDay: Instant
    ) {
        val nextDay = endOfDay.plusSeconds(1)
        logger.warn {
            """Identified ${printRequestsPendingAssignmentToBatch.size} print requests to assign to a batch. 
            Daily print limit is $dailyLimit. 
            $countOfRequestsSentToPrintProvider print requests already sent to print provider today.
            Remaining capacity is ${dailyLimit - countOfRequestsSentToPrintProvider}.
            ${printRequestsPendingAssignmentToBatch.size + countOfRequestsSentToPrintProvider - dailyLimit} print requests won't be assigned to a batch until $nextDay at the earliest."""
        }
    }

</pre>

What do you think?

adacayi (changes requested):
@Neil-Massey, I think there is a more pressing issue then exceeding the daily limit for the print requests when a certificate contains multiple print requests with `PENDING_ASSIGNMENT_TO_BATCH` status. 

We might be exceeding the batch size limit for a given batch, since when preparing a batch we count the number of requests by counting the certificates to be assigned to the batch and if a certificate has multiple requests to be assigned to a batch, this will exceed the batch limit.
 
Since the batch limit is 50, this is much likely to happen when multiple print requests are sent for the same certificate within the same batching period. As a result the batch file might be rejected all together from the print provider.

I added a suggestion for the solution.

What is your opinion on this?

adacayi (approved):

Neil-Massey:
> 

The batch size is not critical, the `50` came from some analysis Nathan and I did - the zip file and PSV file name will report the rare times there are multiple pending print requests for a certificate.  The Print provider does not care about the number per request.

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateBatchingService.kt
adacayi: Thanks for addressing it, @Neil-Massey.

Neil-Massey: Agree - the limit should be applied to the PrintRequests and not the Certificates.  Have included the changes in this PR.

PR #115, opened 2023-01-19
EIP1-2467 Add support to process certificate reprint requests

Neil-Massey:
Print requests for the same Certificate, as identified by the source type and reference are added to the existing certificate instead of creating a second certificate.


src/main/kotlin/uk/gov/dluhc/printapi/database/entity/PrintRequest.kt
```diff
@@ -93,7 +93,7 @@ class PrintRequest(
     @JoinColumn(name = "print_request_id", nullable = false)
     var statusHistory: MutableList<PrintRequestStatus> = mutableListOf(),
 
-    @CreationTimestamp
```
Neil-Massey, line 96: Doesn't populate the reprint Entity that is added to the saved certificate - integration tests fail, so refactored.

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintService.kt
```diff
@@ -3,20 +3,34 @@ package uk.gov.dluhc.printapi.service
...
+            gssCodes = listOf(message.gssCode),
+            sourceType = sourceTypeMapper.toSourceTypeEntity(message.sourceType),
+            sourceReference = message.sourceReference
+        )
+
+        if (certificate != null) {
+            certificateMapper.addPrintRequestToCertificate(message, ero, certificate)
+        } else {
+            certificate = certificateMapper.toCertificate(message, ero)
+        }
```
Neil-Massey, line 32: Explored using Kotlin function scopes but looked more confusing than helpful.

src/test/kotlin/uk/gov/dluhc/printapi/messaging/SendApplicationToPrintMessageListenerIntegrationTest.kt
```diff
@@ -147,33 +79,48 @@ internal class SendApplicationToPrintMessageListenerIntegrationTest : Integratio
...
+        // When
+        sqsMessagingTemplate.convertAndSend(sendApplicationToPrintQueueName, payload)
+
+        // Then
+        await.atMost(5, SECONDS).untilAsserted {
+            wireMockService.verifyEroManagementGetEro(gssCode)
+            val response = certificateRepository.findAll()
+            assertThat(response).hasSize(1)
+            val saved = response[0]
+            assertThat(saved.printRequests).hasSize(2)
```
Neil-Massey, line 121: Not strictly necessary, but does confirm the second print request is loaded on the certificate before the full expected certificate is compared.

src/main/kotlin/uk/gov/dluhc/printapi/mapper/CertificateMapper.kt
```diff
@@ -27,7 +27,7 @@ abstract class CertificateMapper {
     ): Certificate
 
     @AfterMapping
-    protected fun addPrintRequestToCertificate(
+    fun addPrintRequestToCertificate(
```
vishalgupta-valtech, line 30: AfterMapping should be protected, if you wish to use it, then you could add another public fun that eventually after maps and uses same protected method.

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintService.kt
```diff
@@ -3,20 +3,34 @@ package uk.gov.dluhc.printapi.service
...
+    private val sourceTypeMapper: SourceTypeMapper,
     private val certificateMapper: CertificateMapper,
     private val certificateRepository: CertificateRepository
 ) {
     @Transactional
     fun savePrintMessage(message: SendApplicationToPrintMessage) {
         val ero = eroClient.getEro(message.gssCode!!)
-        val certificate = certificateMapper.toCertificate(message, ero)
+
+        var certificate = certificateRepository.findByGssCodeInAndSourceTypeAndSourceReference(
```
vishalgupta-valtech, line 22: It will be better to search using srcRef first. We did something similar in RCa (while deleting) so it may not be relevant. We also created an index on srcRef

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintService.kt
vishalgupta-valtech: V Minor - If cert is null then create createCert else map it. (Instead of !=null)

src/test/kotlin/uk/gov/dluhc/printapi/messaging/SendApplicationToPrintMessageListenerIntegrationTest.kt
```diff
@@ -189,4 +136,116 @@ internal class SendApplicationToPrintMessageListenerIntegrationTest : Integratio
...
+                    status = Status.PENDING_ASSIGNMENT_TO_BATCH,
+                    eventDateTime = Instant.now()
+                )
+            ),
+            userId = userId
+        )
+        return printRequest
+    }
+
+    private fun assertSavedCertificate(
```
vishalgupta-valtech, line 219: Would be it good to extract this assert to its own class so that it can be used for future asserts? (We do something similar in vca)

src/test/kotlin/uk/gov/dluhc/printapi/mapper/CertificateMapperTest.kt
```diff
@@ -134,4 +136,21 @@ class CertificateMapperTest {
...
+        val printRequest = buildPrintRequest()
+        val certificate = buildCertificate(printRequests = listOf(printRequest))
+        val printRequest2 = buildPrintRequest()
+        given(printRequestMapper.toPrintRequest(any(), any())).willReturn(printRequest2)
+
+        // When
+        mapper.addPrintRequestToCertificate(message, ero, certificate)
+
+        // Then
+        assertThat(certificate.printRequests).usingRecursiveComparison().isEqualTo(listOf(printRequest, printRequest2))
```
NathanRussellValtech, line 154: Minor, is it worth doing a verify on printRequestMapper to assert toPrintRequest was called with the correct arguments?

src/test/kotlin/uk/gov/dluhc/printapi/messaging/SendApplicationToPrintMessageListenerIntegrationTest.kt
NathanRussellValtech: Agreed, I'm a big fan of the assertJ API and custom assertions, and I dont think we make enough use of them generally (across all the squads/projects)
I'd like to see it implemented as a custom assertion, and I'd also like to see us get rid of the complexity re `Instant` comparisons with the `BiPredicate` (I vaguely remember refactoring an assertion (not sure which project) that used BiPredicate etc for time comparisons down to little more than the standard assertJ isEqualsTo ??)

src/main/kotlin/uk/gov/dluhc/printapi/mapper/CertificateMapper.kt
NathanRussellValtech: Minor (x2) - I'm not sure why the fun has been made `public`, but I also don't fully follow what @vishalgupta-valtech is saying about different patterns for this use case?

NathanRussellValtech (approved):
LGTM; a couple of minor comments, and if we could do something re the assert method that would be good 👍

src/main/kotlin/uk/gov/dluhc/printapi/mapper/CertificateMapper.kt
Neil-Massey: Was not aware MapStruct methods annotated as `@AfterMapping` should not be public - was made public so could use the method in `uk.gov.dluhc.printapi.service.PrintService`.

Will refactor to have `PrintService` use the `PrintRequestMapper` and then call the `certificate.addPrintRequest` method.

src/test/kotlin/uk/gov/dluhc/printapi/messaging/SendApplicationToPrintMessageListenerIntegrationTest.kt
Neil-Massey: You updated `CertificateRepositoryIntegrationTest` but that works rounding issues by controlling the assigned `Instant` using second precision in saved entity `Instant.now().truncatedTo(SECONDS)` and most other fields are known for the saved entity.  So following was possible :
```
private val IGNORED_FIELDS = arrayOf(".*dateCreated")

            assertThat(actual)
                .usingRecursiveComparison()
                .ignoringFieldsMatchingRegexes(*IGNORED_FIELDS)
                .isEqualTo(expected)
```                

In the integration test the Mappers assign real timestamps (not artificially truncated for comparison convenience) and includes randomly assigned data such as Ids and event dates.  So tricks used in that test class cannot be applied in this Integration test.

If we want to not use the AssertJ recursive comparison (which seemed to be the pattern used on the project) I'll introduce one - so many properties to compare though :-(

vishalgupta-valtech (approved):
LGTM thanks for addressing comments.

src/main/kotlin/uk/gov/dluhc/printapi/service/PrintService.kt
Neil-Massey: This is the same look up as used in the `CertificateFinderService` - possibly could have used that here and handled the `CertificateNotFoundException` exception.

There is already an existing index for this query:
```
        <createIndex indexName="source_details_inx" tableName="certificate" unique="true">
            <column name="gss_code"/>
            <column name="source_type"/>
            <column name="source_reference"/>
        </createIndex>
    </changeSet>
```

PR #118, opened 2023-01-24
Bugfix EIP1-3454 - use new ero management API

NathanRussellValtech:
This PR fixes bug EIP1-3454 by bringing in the latest ERO Management openApi spec (v1.2.1) which introduces a new `name` field on the `contactDetails` object in the REST API response from ERO Management. The new `name` field is then mapped to the `issuingAuthority` and `issuingAuthorityCy` fields in the `Certificate` entity.

The rest of the code required for this overall bugfix is already there and was committed in my last PR for this.


Neil-Massey (approved):
LGTM

PR #122, opened 2023-02-02
EIP1-3610 - liquibase DDL & entity classes for temporary certificate tables

NathanRussellValtech:
This PR implements the liquibase DDL, and entity and repository classes for the new tables required for temporary certificates.

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/TemporaryCertificate.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/TemporaryCertificate.kt
@@ -1,0 +1,132 @@
+package uk.gov.dluhc.printapi.database.entity
+
+import org.hibernate.Hibernate
+import org.hibernate.annotations.CreationTimestamp
+import org.hibernate.annotations.GenericGenerator
+import org.hibernate.annotations.Type
+import org.springframework.data.annotation.LastModifiedBy
+import org.springframework.data.jpa.domain.support.AuditingEntityListener
+import uk.gov.dluhc.printapi.database.repository.UUIDCharType
+import uk.gov.dluhc.printapi.database.repository.UseExistingOrGenerateUUID
+import java.time.Instant
+import java.time.LocalDate
+import java.util.UUID
+import javax.persistence.CascadeType
+import javax.persistence.Entity
+import javax.persistence.EntityListeners
+import javax.persistence.EnumType
+import javax.persistence.Enumerated
+import javax.persistence.FetchType
+import javax.persistence.GeneratedValue
+import javax.persistence.GenerationType
+import javax.persistence.Id
+import javax.persistence.JoinColumn
+import javax.persistence.OneToMany
+import javax.persistence.Table
+import javax.persistence.Version
+import javax.validation.constraints.NotNull
+import javax.validation.constraints.Size
+
+@Table
+@Entity
+@EntityListeners(AuditingEntityListener::class)
+class TemporaryCertificate(
+
+    @Id
+    @Type(type = UUIDCharType)
+    @GeneratedValue(strategy = GenerationType.IDENTITY, generator = "UUID")
+    @GenericGenerator(name = "UUID", strategy = UseExistingOrGenerateUUID.NAME)
+    var id: UUID? = null,
+
+    @field:NotNull
+    @field:Size(max = 20)
+    var certificateNumber: String? = null,
+
+    @field:NotNull
+    @field:Size(max = 80)
+    var gssCode: String? = null,
+
+    @field:NotNull
+    @Enumerated(EnumType.STRING)
+    var sourceType: SourceType? = null,
+
+    @field:NotNull
+    @field:Size(max = 255)
+    var sourceReference: String? = null,
+
+    @field:Size(max = 255)
+    var applicationReference: String? = null,
+
+    @field:NotNull
+    @field:Size(max = 255)
+    var certificateTemplateFilename: String? = null,
+
+    @field:NotNull
+    @field:Size(max = 255)
+    var issuingAuthority: String? = null,
+
+    @field:Size(max = 255)
+    var issuingAuthorityCy: String? = null,
+
+    @field:NotNull
+    @field:Size(max = 255)
+    var firstName: String? = null,
+
+    @field:Size(max = 255)
+    var middleNames: String? = null,
+
+    @field:NotNull
+    @field:Size(max = 255)
+    var surname: String? = null,
+
+    @field:NotNull
+    @Enumerated(EnumType.STRING)
+    var certificateLanguage: CertificateLanguage? = null,
+
+    @field:NotNull
+    @field:Size(max = 255)
+    var photoLocationArn: String? = null,
+
+    @field:NotNull
+    var issueDate: LocalDate = LocalDate.now(),
+
+    @field:NotNull
+    var validOnDate: LocalDate? = null,
+
+    @OneToMany(cascade = [CascadeType.ALL], orphanRemoval = true, fetch = FetchType.LAZY)
+    @JoinColumn(name = "temporary_certificate_id", nullable = false)
+    var statusHistory: MutableList<TemporaryCertificateStatus> = mutableListOf(),
+
+    @field:NotNull
+    @field:Size(max = 255)
+    var userId: String? = null,
+
+    @CreationTimestamp
+    var dateCreated: Instant? = null,
+
+    @field:Size(max = 255)
+    @LastModifiedBy
+    var createdBy: String? = null,
+
+    @Version
+    var version: Long? = null
+) {
+
+    val status: TemporaryCertificateStatus.Status?
+        get() = statusHistory.sortedByDescending { it.dateCreated }.first().status
+
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other == null || Hibernate.getClass(this) != Hibernate.getClass(other)) return false
+        other as TemporaryCertificate
+
+        return id != null && id == other.id
+    }
+
+    override fun hashCode(): Int = javaClass.hashCode()
+
+    @Override
+    override fun toString(): String {
+        return this::class.simpleName + "(id = $id , gssCode = $gssCode, dateCreated = $dateCreated , createdBy = $createdBy , version = $version )"
+    }
+}
--- a/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/TemporaryCertificateStatus.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/TemporaryCertificateStatus.kt
@@ -1,0 +1,71 @@
+package uk.gov.dluhc.printapi.database.entity
+
+import org.hibernate.Hibernate
+import org.hibernate.annotations.CreationTimestamp
+import org.hibernate.annotations.GenericGenerator
+import org.hibernate.annotations.Type
+import org.springframework.data.annotation.LastModifiedBy
+import org.springframework.data.jpa.domain.support.AuditingEntityListener
+import uk.gov.dluhc.printapi.database.repository.UUIDCharType
+import uk.gov.dluhc.printapi.database.repository.UseExistingOrGenerateUUID
+import java.time.Instant
+import java.util.UUID
+import javax.persistence.Entity
+import javax.persistence.EntityListeners
+import javax.persistence.EnumType
+import javax.persistence.Enumerated
+import javax.persistence.GeneratedValue
+import javax.persistence.GenerationType
+import javax.persistence.Id
+import javax.persistence.Table
+import javax.persistence.Version
+import javax.validation.constraints.NotNull
+import javax.validation.constraints.Size
+
+@Table
+@Entity
+@EntityListeners(AuditingEntityListener::class)
+class TemporaryCertificateStatus(
+    @Id
+    @Type(type = UUIDCharType)
+    @GeneratedValue(strategy = GenerationType.IDENTITY, generator = "UUID")
+    @GenericGenerator(name = "UUID", strategy = UseExistingOrGenerateUUID.NAME)
+    var id: UUID? = null,
+
+    @field:NotNull
+    @Enumerated(EnumType.STRING)
+    var status: Status? = null,
+
+    @field:NotNull
+    @field:Size(max = 255)
+    var userId: String? = null,
+
+    @CreationTimestamp
+    var dateCreated: Instant? = null,
+
+    @field:Size(max = 255)
+    @LastModifiedBy
+    var createdBy: String? = null,
+
+    @Version
+    var version: Long? = null
+) {
+    override fun equals(other: Any?): Boolean {
+        if (this === other) return true
+        if (other == null || Hibernate.getClass(this) != Hibernate.getClass(other)) return false
+        other as TemporaryCertificateStatus
+
+        return id != null && id == other.id
+    }
+
+    override fun hashCode(): Int = javaClass.hashCode()
+
+    @Override
+    override fun toString(): String {
+        return this::class.simpleName + "(id = $id , dateCreated = $dateCreated , createdBy = $createdBy , version = $version )"
+    }
+
+    enum class Status {
+        GENERATED
+    }
+}
--- a/src/main/kotlin/uk/gov/dluhc/printapi/database/repository/TemporaryCertificateRepository.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/database/repository/TemporaryCertificateRepository.kt
@@ -1,0 +1,13 @@
+package uk.gov.dluhc.printapi.database.repository
+
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import uk.gov.dluhc.printapi.database.entity.SourceType
+import uk.gov.dluhc.printapi.database.entity.TemporaryCertificate
+import java.util.UUID
+
+@Repository
+interface TemporaryCertificateRepository : JpaRepository<TemporaryCertificate, UUID> {
+
+    fun findByGssCodeInAndSourceTypeAndSourceReference(gssCodes: List<String>, sourceType: SourceType, sourceReference: String): TemporaryCertificate?
+}
--- a/src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
+++ b/src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
@@ -110,4 +110,12 @@
         </createTable>
     </changeSet>
 
+    <changeSet author="nathan.russell@valtech.com" id="0010-EIP1-3610-source-details-index" context="ddl">
+        <createIndex indexName="source_details_inx" tableName="temporary_certificate" unique="true">
+            <column name="gss_code"/>
+            <column name="source_type"/>
+            <column name="source_reference"/>
+        </createIndex>
+    </changeSet>
+
 </databaseChangeLog>
--- a/src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
+++ b/src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
@@ -24,26 +24,15 @@
                 <constraints nullable="false"/>
             </column>
             <column name="application_reference" type="varchar(255)"/>
-            <column name="certificate_template_version" type="varchar(20)">
+            <column name="certificate_template_filename" type="varchar(255)">
                 <constraints nullable="false"/>
             </column>
             <column name="issuing_authority" type="varchar(255)">
                 <constraints nullable="false"/>
             </column>
-            <column name="english_ero_id" type="uuid">
-                <constraints nullable="false"
-                             foreignKeyName="fk_temporary_certificate_english_ero_id"
-                             references="electoral_registration_office(id)"/>
-            </column>
             <column name="issuing_authority_cy" type="varchar(255)">
                 <constraints nullable="true"/>
             </column>
-            <column name="welsh_ero_id" type="uuid">
-                <constraints nullable="true"
-                             foreignKeyName="fk_temporary_certificate_welsh_ero_id"
-                             references="electoral_registration_office(id)"/>
-            </column>
-
             <column name="first_name" type="varchar(255)">
                 <constraints nullable="false"/>
             </column>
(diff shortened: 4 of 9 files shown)


src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
```diff
@@ -0,0 +1,113 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+
+    <changeSet author="nathan.russell@valtech.com" id="0010-EIP1-3610-create-temporary-certificate-table" context="ddl">
```
NathanRussellValtech, line 8: `temporary_certificate` table, based heavily on a flattened combo of `certificate` and `print_request`

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
```diff
@@ -0,0 +1,113 @@
...
+        <createTable tableName="temporary_certificate">
+            <column name="id" type="uuid">
+                <constraints nullable="false" primaryKey="true" unique="true"
+                             primaryKeyName="temporary_certificate_pkey"/>
+            </column>
+            <column name="certificate_number" type="varchar(20)">
```
NathanRussellValtech, line 14: `temporary_certificate_number` felt quite long and wordy ?, but happy for suggestions

(its called `vac_certificate` in `certificate` table)

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
```diff
@@ -0,0 +1,113 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="gss_code" type="varchar(80)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="source_type" type="varchar(20)">
```
NathanRussellValtech, line 20: As with `certificate` I think it's useful having `source_type` (currently only `VOTER_CARD`) and `application_reference` (the "V" reference number). We don't currently do anything/much with these fields, but feels useful to have them?

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
```diff
@@ -0,0 +1,113 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="source_reference" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="application_reference" type="varchar(255)"/>
+            <column name="certificate_template_version" type="varchar(20)">
```
NathanRussellValtech, line 27: This is the equivalent of `vac_version` in `print_request` which for a main cert is currently hardcoded as "A" (it means more to the print provider than us!)

For temporary certs, where we are generating the cert from a PDF template, perhaps its useful to have a template version reference?

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
```diff
@@ -0,0 +1,113 @@
...
+            <column name="welsh_ero_id" type="uuid">
+                <constraints nullable="true"
+                             foreignKeyName="fk_temporary_certificate_welsh_ero_id"
+                             references="electoral_registration_office(id)"/>
+            </column>
+
+            <column name="first_name" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="middle_names" type="varchar(255)"/>
```
NathanRussellValtech, line 50: The current layout for the temp cert does not include the middle name(s), but we collect it for the main cert and it is sent to the print provider ... and presumably they print it on the cert if its provided?
I'm thinking we collect it in the table just in case they change the design in this respect!

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
```diff
@@ -0,0 +1,113 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="version" type="bigint">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet author="nathan.russell@valtech.com" id="0010-EIP1-3610-create-temporary-certificate-status" context="ddl">
+        <createTable tableName="temporary_certificate_status">
```
NathanRussellValtech, line 85: Much like with the `print_request` for the main cert where we have a collection of statuses to be able to show a timeline/history of events ... I'm thinking it might be prudent to do a similar thing here.

At the moment there is only 1 status we know about ("generated"), but it wouldn't surprise me if we were asked to track other events/statuses .... such as being printed, being collected/delivered to the elector, expiry etc

I know for AEDs they are currently talking about 2 statuses (generated and printed), so if that requirement comes out way we will be able to cater for it

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
```diff
@@ -0,0 +1,113 @@
...
+            <column name="photo_location_arn" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="issue_date" type="date">
+                <constraints nullable="false"/>
+            </column>
+            <column name="valid_on_date" type="date">
+                <constraints nullable="false"/>
+            </column>
+            <column name="status" type="varchar(50)">
```
NathanRussellValtech, line 66: I am proposing that this `status` column is the last/most recent `temporary_certificate_status` .... exactly as we do with the main certificate and it's print_requests

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
```diff
@@ -0,0 +1,121 @@
...
+            </column>
+
+            <column name="first_name" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="middle_names" type="varchar(255)"/>
+            <column name="surname" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="certificate_language" type="varchar(20)">
```
adacayi, line 54: @NathanRussellValtech , will we be needing a `certificate_format` as well?

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
```diff
@@ -0,0 +1,121 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="version" type="bigint">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
+    </changeSet>
+
+    <changeSet author="nathan.russell@valtech.com" id="0010-EIP1-3610-source-details-index" context="ddl">
+        <createIndex indexName="source_details_inx" tableName="temporary_certificate" unique="true">
```
adacayi, line 114: @NathanRussellValtech , since we can have multiple template certificates for the same `gss_code`, `source_type` and `source_reference`, we will need to have this index as non-unique.

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
```diff
@@ -0,0 +1,121 @@
...
+            <column name="date_created" type="timestamp">
+                <constraints nullable="false"/>
+            </column>
+            <column name="created_by" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="version" type="bigint">
+                <constraints nullable="false"/>
+            </column>
+        </createTable>
```
adacayi, line 81: @NathanRussellValtech, do you think it will be good to add rollback statements for each changeset?
```
<rollback>
      <dropTable tableName="temporary_certificate" />
</rollback>
```

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
Neil-Massey: Maybe `template-doc` - give the full name including it's version (assumes we include version on the template pdf - that I think we should)

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
```diff
@@ -0,0 +1,121 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="application_reference" type="varchar(255)"/>
+            <column name="certificate_template_version" type="varchar(20)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="issuing_authority" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="english_ero_id" type="uuid">
```
Neil-Massey, line 33: I wonder if we need these as they do not appear on the temporary certificate

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
NathanRussellValtech: Yes, thats a good idea. Yes, we can set the field value to the filename of the template pdf we used as the template. Nice. I'll change the field name 👍

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
NathanRussellValtech: I'll check, but I don't think so.
I _think_ the current `certificate_format` on the `certificate` / `print_request` table is actually a badly named field, and actually represents the format of the explainer document that the print provider print with the main cert. IE. its `standard`, `easy-read`, `large-print` or `braille`
I _think_ thats what its currently used for, and if so, then I dont think we need it in this case. But I'll check

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
NathanRussellValtech: Yes, 💯 agreed. Thanks for pointing it out and keeping me honest! 😁 🤣

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
NathanRussellValtech: Good spot, yes, I'll change that 👍

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
```diff
@@ -0,0 +1,121 @@
...
+            </column>
+            <column name="temporary_certificate_id" type="uuid">
+                <constraints nullable="false"
+                             foreignKeyName="fk_temporary_certificate_status_temporary_certificate"
+                             references="temporary_certificate(id)"/>
+            </column>
+            <column name="status" type="varchar(50)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="event_date_time" type="timestamp">
```
adacayi, line 98: Do you think we need `event_date_time` for this table? 

As far as I can remember, we are getting this value from the print-provider for the actual certificates (which will be different from the `date_created` values), but for the temporary certificate I guess it will be the nearly identical to the `date_created` value (I assume there will be an endpoint, if required, to update the status and the `event_date_time` will be the instant that button is pressed and it will update the status of the temporary certificate synchronously).

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
NathanRussellValtech: Done

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
NathanRussellValtech: Added rollbacks 👍

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
Neil-Massey: We'd possibly have a separate table for AEDs.  
Possibly this is  `temporary_certificate_status_history` with the current status stored directly on the `temporary_certificate`.

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
```diff
@@ -0,0 +1,121 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="date_created" type="timestamp">
+                <constraints nullable="false"/>
+            </column>
+            <column name="created_by" type="varchar(255)">
```
adacayi, line 104: We might need a `user_id` column as well to capture the user updating the status. 

I imagine we don't need this with the actual certificate statuses, because it is not updated manually by a user. For temporary certificates though, seems like, if there is a status update, it will be triggered by a user.

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
Neil-Massey: It's in the `temporary_certificate` table, so `certificate_number` sounds good to me 👍

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
NathanRussellValtech: I've called it `certificate_template_filename` and set it's width to 255

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
```diff
@@ -0,0 +1,121 @@
...
+            <column name="id" type="uuid">
+                <constraints nullable="false" primaryKey="true" unique="true"
+                             primaryKeyName="temporary_certificate_status_pkey"/>
+            </column>
+            <column name="temporary_certificate_id" type="uuid">
+                <constraints nullable="false"
+                             foreignKeyName="fk_temporary_certificate_status_temporary_certificate"
+                             references="temporary_certificate(id)"/>
+            </column>
+            <column name="status" type="varchar(50)">
```
adacayi, line 95: It might be good to have an index on `status` and `date_created` (or `event_date_time` if we decide to retain it)

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
adacayi: Thanks 👍

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
NathanRussellValtech: Agreed; gone 😁

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
NathanRussellValtech: Yes, `history` in the table name suffix. Like it 👍

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
NathanRussellValtech: Good point; yes; I'll get rid of `event_date_time` 👍

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
NathanRussellValtech: Yes, agreed 👍

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
NathanRussellValtech: OK 👍

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
NathanRussellValtech: Ah, OK, yes its as I thought ...
`certificateFormat` has been renamed to `supportingInformationFormat` and is for the format of the supporting info document (ie. the explainer doc)
It was done in EIP1-2781:
```
        <renameColumn
            tableName="print_request"
            oldColumnName="certificate_format"
            newColumnName="supporting_information_format"
            columnDataType="varchar(20)"
        />
```

We dont need this for temporary certificates 👍

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/TemporaryCertificateStatus.kt
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    override fun hashCode(): Int = javaClass.hashCode()
+
+    @Override
+    override fun toString(): String {
+        return this::class.simpleName + "(id = $id , dateCreated = $dateCreated , createdBy = $createdBy , version = $version )"
+    }
+
+    enum class Status {
```
NathanRussellValtech, line 68: I've added the `Status` enum as an inner class. This is because there is already a `Status` enum in the entity package, which is used for the main certificate.
The values in these 2 status enums are different; and we need to be able to differentiate between the 2, so an inner class felt right.

Once this PR is merged then I'm planning on refactoring the existing `Status` enum (that is used for regular certificates) as an inner class of `PrintRequestStatus` - but I'll do that separately to this PR because otherwise this would be massive and impossible to follow

src/test/kotlin/uk/gov/dluhc/printapi/config/LocalStackContainerConfiguration.kt
```diff
@@ -45,12 +45,15 @@ class LocalStackContainerConfiguration {
...
                 ).withEnv(
                     mapOf(
-                        "SERVICES" to "sqs, dynamodb",
+                        "SERVICES" to "sqs",
                         "AWS_DEFAULT_REGION" to DEFAULT_REGION,
                     )
                 )
                     .withReuse(true)
                     .withExposedPorts(DEFAULT_PORT)
+                    .withCreateContainerCmdModifier {
```
NathanRussellValtech, line 54: Not strictly related to this PR but a failing set of tests on GitHub made me come and look at this ....
We no longer use dynamo in print-api, so I have removed that from the local stack config

Plus, something thats been nagging at me for a while .... we can name our TestContainer containers, which means they dont get random names in Docker Desktop. Whilst we re-use the containers, they are not shared amongst our apps (as you would expect). If you work on several app repos (which we all do), you are likely to have several mysql containers, and several localstack containers. Naming them makes a good deal of sense and we should do this in the other apps too 👍 

Screenshot 2023-02-02 at 14 52 09

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
Neil-Massey: Even though there is only a `Name` on the temporary certificate - agree better to store the components separately.

src/main/resources/db/changelog/ddl/0010_create_temporary_certificate_tables.xml
Neil-Massey: Is this going to be `temporary_certificate_status_hostory`?

Neil-Massey (approved):
LGTM - Minor query on table name

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/TemporaryCertificate.kt
```diff
@@ -0,0 +1,132 @@
...
+
+    @field:NotNull
+    var issueDate: LocalDate = LocalDate.now(),
+
+    @field:NotNull
+    var validOnDate: LocalDate? = null,
+
+    @OneToMany(cascade = [CascadeType.ALL], orphanRemoval = true, fetch = FetchType.LAZY)
+    @JoinColumn(name = "temporary_certificate_id", nullable = false)
+    var statusHistory: MutableList<TemporaryCertificateStatus> = mutableListOf(),
```
Neil-Massey, line 98: This looks good 👍

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/TemporaryCertificateStatus.kt
```diff
@@ -0,0 +1,71 @@
...
+        other as TemporaryCertificateStatus
+
+        return id != null && id == other.id
+    }
+
+    override fun hashCode(): Int = javaClass.hashCode()
+
+    @Override
+    override fun toString(): String {
+        return this::class.simpleName + "(id = $id , dateCreated = $dateCreated , createdBy = $createdBy , version = $version )"
```
adacayi, line 65: Minor, @NathanRussellValtech .
I know this is merged, but if you have a change in the following PRs, it might be useful to add the status value to the `toString`.

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/TemporaryCertificateRepository.kt
```diff
@@ -0,0 +1,13 @@
...
+import org.springframework.data.jpa.repository.JpaRepository
+import org.springframework.stereotype.Repository
+import uk.gov.dluhc.printapi.database.entity.SourceType
+import uk.gov.dluhc.printapi.database.entity.TemporaryCertificate
+import java.util.UUID
+
+@Repository
+interface TemporaryCertificateRepository : JpaRepository<TemporaryCertificate, UUID> {
+
+    fun findByGssCodeInAndSourceTypeAndSourceReference(gssCodes: List<String>, sourceType: SourceType, sourceReference: String): TemporaryCertificate?
```
adacayi, line 12: Hi @NathanRussellValtech,

One more for the follow-up PR hopefully. I assume we will expect this method to return multiple temporary certificates, since we can have multiple temporary certificates for a given `gssCode`, `sourceType` and `sourceReference`.

src/test/kotlin/uk/gov/dluhc/printapi/config/LocalStackContainerConfiguration.kt
adacayi: @NathanRussellValtech , minor. Maybe we can add test to the container name, since this container won't be used when we run the application locally. Might prevent some misunderstanding for new-comers.

PR #123, opened 2023-02-02
EIP1-3666 Temporary card explainer document spec

Neil-Massey:
image

The language of the explainer is selected based on the ERO Nation code included in the gssCode.

### DIFF
--- a/src/main/resources/openapi/PrintAPIs.yaml
+++ b/src/main/resources/openapi/PrintAPIs.yaml
@@ -99,9 +99,9 @@ paths:
 
 
   #
-  # Temporary Card Explainer Document
+  # Temporary Certificate Explainer Document
   # --------------------------------------------------------------------------------
-  '/eros/{eroId}/temp-card/{gssCode}/explainer-document':
+  '/eros/{eroId}/temporary-certificate/{gssCode}/explainer-document':
     parameters:
       - name: eroId
         description: The ID of the Electoral Registration Office responsible for the application.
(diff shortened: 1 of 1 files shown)


src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -97,6 +97,95 @@ paths:
...
+          description: Not Found
+      operationId: generate-temporary-certificate-explainer-document
+      security:
+        - eroUserCognitoUserPoolAuthorizer: [ ]
+      x-amazon-apigateway-integration:
+        type: HTTP_PROXY
+        uri: '${base_uri}/eros/{eroId}/temporary-certificate/{gssCode}/explainer-document'
+        requestParameters:
+          integration.request.path.eroId: method.request.path.eroId
+          integration.request.path.applicationId: method.request.path.applicationId
```
NathanRussellValtech, line 180: Should be gssCode here?

NathanRussellValtech (approved):
LGTM - a minor correction needed for the request path parameter; plus the spec itself will need a version bump; but other than that 👍

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -97,6 +97,95 @@ paths:
         connectionId: '${vpc_connection_id}'
         httpMethod: GET
 
+
+  #
+  # Temporary Certificate Explainer Document
+  # --------------------------------------------------------------------------------
+  '/eros/{eroId}/temporary-certificate/{gssCode}/explainer-document':
```
Neil-Massey, line 104: Possibly, can move the `gssCode` to  a request body - included in the URL as initially thought this would be a `GET`

src/main/resources/openapi/PrintAPIs.yaml
Neil-Massey: Thanks 👍

PR #124, opened 2023-02-02
EIP1-3610 - Refactored Certificate status

NathanRussellValtech:
This PR refactors the `Status` enum class as used with the `Certificate` entity, and moves it as an inner class of `PrintRequestStatus` (in the same way that I did with Temporary Certificates in my last PR)

It looks like a massive PR, but for the most part its just a load of import changes and reference changes from `Status` to `PrintRequestStatus.Status`


src/main/kotlin/uk/gov/dluhc/printapi/database/entity/PrintRequestStatus.kt
```diff
@@ -64,6 +64,20 @@ class PrintRequestStatus(
 
     @Override
     override fun toString(): String {
-        return this::class.simpleName + "(id = $id , dateCreated = $dateCreated , createdBy = $createdBy , version = $version )"
+        return this::class.simpleName + "(id = $id , status = $status , dateCreated = $dateCreated , createdBy = $createdBy , version = $version )"
```
NathanRussellValtech, line 67: I've added the `status` field to the `toString` method in this class, and also `TemporaryCertificateStatus` as per a late suggestion in my last PR - FYI @abdullah-sanver-valtech

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/TemporaryCertificateStatus.kt
```diff
@@ -62,7 +62,7 @@ class TemporaryCertificateStatus(
 
     @Override
     override fun toString(): String {
-        return this::class.simpleName + "(id = $id , dateCreated = $dateCreated , createdBy = $createdBy , version = $version )"
+        return this::class.simpleName + "(id = $id , status = $status , dateCreated = $dateCreated , createdBy = $createdBy , version = $version )"
```
NathanRussellValtech, line 65: FYI @abdullah-sanver-valtech

src/main/kotlin/uk/gov/dluhc/printapi/database/mapper/PrintRequestStatusDtoMapper.kt
```diff
@@ -0,0 +1,10 @@
+package uk.gov.dluhc.printapi.database.mapper
+
+import uk.gov.dluhc.printapi.database.entity.PrintRequestStatus
+import uk.gov.dluhc.printapi.dto.PrintRequestStatusDto
+
+class PrintRequestStatusDtoMapper {
```
NathanRussellValtech, line 6: I've renamed the mapper class - it looks like a delete and new class, but its just a rename

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/TemporaryCertificateRepository.kt
```diff
@@ -9,5 +9,5 @@ import java.util.UUID
 @Repository
 interface TemporaryCertificateRepository : JpaRepository<TemporaryCertificate, UUID> {
 
-    fun findByGssCodeInAndSourceTypeAndSourceReference(gssCodes: List<String>, sourceType: SourceType, sourceReference: String): TemporaryCertificate?
+    fun findByGssCodeInAndSourceTypeAndSourceReference(gssCodes: List<String>, sourceType: SourceType, sourceReference: String): List<TemporaryCertificate>
```
NathanRussellValtech, line 12: FYI @abdullah-sanver-valtech

matt-wills-valtech (approved):
Funnily enough, I had the same thought when I saw this the other day - `Status` is a generic name and having it within the context of the class that uses it is much clearer.

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
```diff
@@ -76,7 +76,7 @@ class Certificate(
      */
     @field:NotNull
     @Enumerated(EnumType.STRING)
-    var status: Status? = null,
+    var status: PrintRequestStatus.Status? = null,
```
Neil-Massey, line 79: Can this be added as an import?

Neil-Massey (approved):
LGTM - Just query about having an `import` in `Certificate`

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/PrintRequestStatus.kt
adacayi: Thank you, @NathanRussellValtech

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/TemporaryCertificateRepository.kt
adacayi: Thanks, for addressing it @NathanRussellValtech.

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
NathanRussellValtech: I _could_ and I kinda thought about it, but I decided against it because then we are back to referencing `Status` in isolation without the context of what kind of status it is.

eg: it would be this:
```
import uk.gov.dluhc.printapi.database.entity.PrintRequestStatus.Status

    @field:NotNull
    @Enumerated(EnumType.STRING)
    var status: Status? = null,
```
I know it would be there in the import, but it felt a bit lost

adacayi (approved):

PR #125, opened 2023-02-03
EIP1-3439 / EIP1-3721 - openAPI spec for generating a temporary certificate

NathanRussellValtech:
This PR updates the openAPI spec with the new endpoint for generating a temporary certificate
(it also updates the TestContainer names based on a request from @abdullah-sanver-valtech that I saw after I'd merged my last PR)

Screenshot 2023-02-03 at 09 56 58

### DIFF
--- a/src/main/resources/openapi/PrintAPIs.yaml
+++ b/src/main/resources/openapi/PrintAPIs.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Print APIs
-  version: '1.2.0'
+  version: '1.3.0'
   description: Print APIs
   contact:
     name: Krister Bone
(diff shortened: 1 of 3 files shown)


src/test/kotlin/uk/gov/dluhc/printapi/config/LocalStackContainerConfiguration.kt
```diff
@@ -52,7 +52,7 @@ class LocalStackContainerConfiguration {
                     .withReuse(true)
                     .withExposedPorts(DEFAULT_PORT)
                     .withCreateContainerCmdModifier {
-                        it.withName("print-api-localstack")
+                        it.withName("print-api-integration-test-localstack")
```
adacayi, line 55: Thanks, @NathanRussellValtech , for the change.

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -98,6 +98,90 @@ paths:
...
+        $ref: '#/components/requestBodies/GenerateTemporaryCertificate'
+      responses:
+        '201':
+          description: created
+          headers:
+            Content-Disposition:
+              description: Content-Disposition header
+              schema:
+                type: string
+                example: attachment; filename="temporary-certificate-V3JSZC4CRH.pdf"
```
Neil-Massey, line 164: As multiple temporary certificates can be created for same application reference, possible including the certificate number in the filename is better.

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -269,6 +353,80 @@ components:
...
+          maxLength: 255
+          description: Surname of the Elector
+          example: Smith
+        certificateLanguage:
+          $ref: '#/components/schemas/CertificateLanguage'
+        photoLocation:
+          type: string
+          description: The location of the Elector's photo. Typically an S3 ARN
+          maxLength: 1024
+        validOnDate:
```
adacayi, line 414: One quick question, @NathanRussellValtech , just to confirm. 
Will we be getting the userId information of the user making the request from the bearer token that's passed through?

adacayi (approved):

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -269,6 +353,80 @@ components:
...
+          description: Middle names of the Elector
+          example: Malcolm
+        surname:
+          type: string
+          maxLength: 255
+          description: Surname of the Elector
+          example: Smith
+        certificateLanguage:
+          $ref: '#/components/schemas/CertificateLanguage'
+        photoLocation:
```
Neil-Massey, line 410: As there are a few different photos in the system, possible `certificatePhotoLocation` - but probably not necessary given request context. Although the description could possibly be more explicit.

Neil-Massey (approved):
LGTM - Few minor comments

src/main/resources/openapi/PrintAPIs.yaml
Neil-Massey: Nathan mentioned that was his preferred approach at our refinement 👍

NathanRussellValtech: Yes. I think it would be better to get the user if from the authentication token, rather than allowing it to be passed in the request (and potentially abused)

src/main/resources/openapi/PrintAPIs.yaml
NathanRussellValtech: Good idea; yes, I'll do that 👍

src/main/resources/openapi/PrintAPIs.yaml
NathanRussellValtech: Yep, agreed. I'll do that 👍

PR #128, opened 2023-02-06
EIP1-3439 / EIP1-3722 - Request validation with 400 scenarios

NathanRussellValtech:
This PR introduces the request body validation and HTTP 400 scenarios for the REST API to generate a temporary certificate


src/main/kotlin/uk/gov/dluhc/printapi/rest/TemporaryCertificateController.kt
```diff
@@ -1,18 +1,37 @@
...
 @RestController
 @CrossOrigin
-class TemporaryCertificateController {
+class TemporaryCertificateController(
+    private val generateTemporaryCertificateRequestValidator: GenerateTemporaryCertificateRequestValidator
+) {
+
+    @InitBinder("generateTemporaryCertificateRequest")
+    fun initGenerateTemporaryCertificateRequestValidator(binder: WebDataBinder) {
+        binder.addValidators(generateTemporaryCertificateRequestValidator)
```
NathanRussellValtech, line 24: We need to bind a custom validator to the `generateTemporaryCertificateRequest` request parameter. This is over and above the standard JSR-303 validation that is already implemented by virtue of the fact that the openAPI codegen using standard validation annotations.
EG: this in the openAPI spec:
```
        gssCode:
          type: string
          example: E12345678
          minLength: 9
          maxLength: 9
          description: GSS code of the ERO responsible for the application
```
Becomes this in the generated model class:
```
    @get:Size(min=9,max=9)
    @Schema(example = "E12345678", required = true, description = "GSS code of the ERO responsible for the application")
    @field:JsonProperty("gssCode", required = true) val gssCode: kotlin.String,
```

We need to add a custom validator to validate the `validOnDate` field in the request. The business requirements are that the date cannot be in the past, and cannot be greater than 30 days in the future. It is not possible to express this logic in openAPI specs, hence we need the custom validator.

(to be fair it's a different type of validation. JSR-303 validation is static _format_ validation . The validation required for `validOnDate` is a business rule rather than format validation)

src/main/kotlin/uk/gov/dluhc/printapi/rest/TemporaryCertificateController.kt
```diff
@@ -1,18 +1,37 @@
...
 
     @PostMapping("/eros/{eroId}/temporary-certificate")
     @PreAuthorize(HAS_ERO_VC_ADMIN_AUTHORITY)
-    fun generateTemporaryCertificate(@PathVariable eroId: String) {
+    fun generateTemporaryCertificate(
+        @PathVariable eroId: String,
+        @RequestBody @Valid generateTemporaryCertificateRequest: GenerateTemporaryCertificateRequest,
+        authentication: Authentication
+    ) {
+        val userId = authentication.name
```
NathanRussellValtech, line 34: I didn't need the authentication in this PR, but I added it seeing as I was here - we will need it in the next PR

src/main/kotlin/uk/gov/dluhc/printapi/validator/GenerateTemporaryCertificateRequestValidator.kt
```diff
@@ -0,0 +1,41 @@
...
+import org.springframework.validation.Errors
+import org.springframework.validation.Validator
+import uk.gov.dluhc.printapi.models.GenerateTemporaryCertificateRequest
+import java.time.Clock
+import java.time.LocalDate
+
+@Component
+class GenerateTemporaryCertificateRequestValidator(
+    private val clock: Clock,
+    @Value("\${api.print-api.generate-temporary-certificate.valid-on-date-maximum-advance-days}") private val maxAdvanceDays: Long
```
NathanRussellValtech, line 14: I struggled long and hard to come up with a suitable property name and variable name.
What I'm trying to express is the maximum number of days after "today" that a temporary certificate's `validOnDate` field can be set. IE. how far in the future can you set a temp cert's `validOnDate`

Suggestions on better names gratefully received.....

... or indeed, do we even need it in _this_ repo and PR ?
So, the business rule for generating temp certs is that they cannot be in the past and cannot be greater than x days in the future. 
I think `print-api` should validate that you cannot set the date in the past ... but I'm not sure about the future date part of this ....
My thinking is that the UI web client doesnt call this directly. The process is that the UI calls a VCA REST API (EIP1-1278), which in turn calls this REST API
The VCA REST API will definitely need to do this validation, to validate the date is not in the past and is not greater than x days in the future.
But do we need to do the "greater than x days in the future" validation in _both_ APIs?
Doing it in both APIs means we need to define the property in both APIs, and their values will need to be the same.

I wonder if we don't need it here?
What do you think?

src/test/kotlin/uk/gov/dluhc/printapi/rest/GenerateTemporaryCertificateIntegrationTest.kt
```diff
@@ -19,6 +25,10 @@ internal class GenerateTemporaryCertificateIntegrationTest : IntegrationTest() {
     fun `should return forbidden given no bearer token`() {
         webTestClient.post()
             .uri(URI_TEMPLATE, ERO_ID)
+            .body(
```
NathanRussellValtech, line 28: Existing integration tests now need the request body passing in, now that I've added it to the request method signature in the controller.

src/test/kotlin/uk/gov/dluhc/printapi/rest/GenerateTemporaryCertificateIntegrationTest.kt
```diff
@@ -63,8 +81,76 @@ internal class GenerateTemporaryCertificateIntegrationTest : IntegrationTest() {
...
+                Mono.just(buildGenerateTemporaryCertificateRequest()),
+                GenerateTemporaryCertificateRequest::class.java
+            )
             .exchange()
             .expectStatus()
             .isForbidden
     }
+
+    @Test
```
NathanRussellValtech, line 93: These 2 integration tests are not testing the validation rules per se, but are about testing that the request handling has invoked both the standard JSR-303 validation, plus our custom initBinder validation

In respect of the initBinder test, one could argue that its testing one of the scenarios that is also covered in the unit test. Whilst this is true, the objective is not about unit testing the validation class/rules, and is more about testing the validator is bound to the request and is invoked.

src/main/kotlin/uk/gov/dluhc/printapi/rest/TemporaryCertificateController.kt
Neil-Massey: I would expected the `onDate` validations to be in the service layer as another business valdiation.

src/main/kotlin/uk/gov/dluhc/printapi/validator/GenerateTemporaryCertificateRequestValidator.kt
Neil-Massey: `api.print-api.generate-temporary-certificate.on-date.max-calendar-days-in-future`

NathanRussellValtech: > `api.print-api.generate-temporary-certificate.on-date.max-calendar-days-in-future`

Ah yes, nice. I like that @Neil-Massey . My only change would be `on-date` should be `valid-on-date` - but I like the separation of the elements, and the explicit mention of calendar days 👍

Neil-Massey: Good question regarding the necessity to do the validation in both APIs.
Usually both public API endpoints are treated as independent API and both have the same rules.  I can see the potential for configuration discrepancy though.  If I had to choose one to know the correct validation rule I'd have it in print-api (aka Certificate API).

src/test/kotlin/uk/gov/dluhc/printapi/rest/GenerateTemporaryCertificateIntegrationTest.kt
```diff
@@ -63,8 +81,76 @@ internal class GenerateTemporaryCertificateIntegrationTest : IntegrationTest() {
...
+        val actual = response.responseBody.blockFirst()
+        assertThat(actual)
+            .hasStatus(400)
+            .hasError("Bad Request")
+            .hasMessageContaining("Validation failed for object='generateTemporaryCertificateRequest'. Error count: 1")
+            .hasValidationError("Error on field 'gssCode': rejected value [$invalidGssCode], size must be between 9 and 9")
+    }
+
+    @Test
+    fun `should return bad request given initBinder validation failure`() {
```
Neil-Massey, line 126: Usually the method names are more Business focussed (this one and the JSR303 seem too technical)

src/main/kotlin/uk/gov/dluhc/printapi/rest/TemporaryCertificateController.kt
NathanRussellValtech: Agreed. As discussed I'm going to remove this validator class and anything related to it, and re-introduce it in my next PR where I introduce the service layer 👍

src/main/kotlin/uk/gov/dluhc/printapi/validator/GenerateTemporaryCertificateRequestValidator.kt
NathanRussellValtech: Yes, agreed. Lets do the business validation here in this API (but in my next PR)
VCA will call this API, and if it fails with a date too far in the future/past then this API will return a 400, which VCA can pass straight back to the UI 👍 (or deal with it in some way if we think there is value in that)

src/test/kotlin/uk/gov/dluhc/printapi/rest/GenerateTemporaryCertificateIntegrationTest.kt
NathanRussellValtech: Yes, agreed. Now that I've removed the initBinder validation, and we only have the JSR303 validation, I've renamed that test so that it simply talks about an invalid request 👍

Neil-Massey (approved):
LGTM

src/test/kotlin/uk/gov/dluhc/printapi/rest/GenerateTemporaryCertificateIntegrationTest.kt
```diff
@@ -31,6 +40,10 @@ internal class GenerateTemporaryCertificateIntegrationTest : IntegrationTest() {
             .uri(URI_TEMPLATE, ERO_ID)
             .bearerToken(UNAUTHORIZED_BEARER_TOKEN)
             .contentType(MediaType.APPLICATION_JSON)
+            .body(
+                Mono.just(buildGenerateTemporaryCertificateRequest()),
+                GenerateTemporaryCertificateRequest::class.java
+            )
```
adacayi, line 46: Minor suggestion, @NathanRussellValtech ,
We can have an extension function `withABody` to simplify this:
<pre>
        webTestClient.post()
            .uri(URI_TEMPLATE, ERO_ID)
            .bearerToken(getBearerToken(eroId = ERO_ID, groups = listOf("ero-$ERO_ID", "ero-admin-$ERO_ID")))
            .contentType(MediaType.APPLICATION_JSON)
<b>            .withABody(buildGenerateTemporaryCertificateRequest())</b>
</pre>
<pre>
    private fun <T : Any> WebTestClient.RequestBodySpec.withABody(requestBody: T): WebTestClient.RequestBodySpec =
        body(
            Mono.just(requestBody),
            requestBody.javaClass
        ) as WebTestClient.RequestBodySpec
</pre>

NathanRussellValtech: Oh yes, I like that 👍 
I'll do that now 😁

adacayi (approved):
Added one minor suggestion.

PR #134, opened 2023-02-08
EIP1-3802 - Calculate certificate delivery info removal date

matt-wills-valtech:
We've recently discovered we need to remove the address and addressee fields that are stored against each certificate's print requests much sooner than anticipated. Since this data is not on the printed certificate itself, we need to remove it 28 working days after a certificate is "issued" (as opposed to 28 working days following approval for the application's data in VCA).

There is currently some debate re what the "issue" date is exactly, but that is outside the focus of this PR.

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/client/BankHolidayDataClient.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/client/BankHolidayDataClient.kt
@@ -1,0 +1,70 @@
+package uk.gov.dluhc.printapi.client
+
+import com.fasterxml.jackson.annotation.JsonProperty
+import mu.KotlinLogging
+import java.time.LocalDate
+
+private val logger = KotlinLogging.logger {}
+
+interface BankHolidayDataClient {
+
+    /**
+     * Retrieves bank holiday dates for the requested [BankHolidayDivision] (e.g. England and Wales), starting from the
+     * provided [LocalDate].
+     *
+     * @param division The UK nation (e.g. Scotland) that the bank holidays are required for.
+     * @param fromDate The [LocalDate] from which the bank holidays start (defaults to today).
+     * @param toDate The [LocalDate] when the bank holidays end (defaults to an arbitrary 100 days in the future).
+     * @return a [List] of [LocalDate] values representing upcoming bank holidays.
+     */
+    fun getBankHolidayDates(
+        division: BankHolidayDivision,
+        fromDate: LocalDate = LocalDate.now(),
+        toDate: LocalDate = LocalDate.now().plusDays(100)
+    ): List<LocalDate>
+}
+
+/**
+ * The bank holiday data we retrieve from gov.uk is structured according to the UK nation (or "division" in the JSON).
+ */
+enum class BankHolidayDivision(val value: String) {
+    ENGLAND_AND_WALES("england-and-wales"),
+    SCOTLAND("scotland"),
+    NORTHERN_IRELAND("northern-ireland");
+
+    companion object {
+        @JvmStatic
+        fun fromGssCode(gssCode: String): BankHolidayDivision {
+            return when (gssCode.substring(0, 1)) {
+                "E", "W" -> ENGLAND_AND_WALES
+                "S" -> SCOTLAND
+                "N" -> NORTHERN_IRELAND
+                else -> {
+                    logger.warn("Unknown country prefix for gssCode: [$gssCode], defaulting to $ENGLAND_AND_WALES")
+                    ENGLAND_AND_WALES
+                }
+            }
+        }
+    }
+}
+
+/**
+ * Data class to deserialize the JSON from gov.uk's Bank Holidays API (https://www.api.gov.uk/gds/bank-holidays).
+ */
+data class BankHolidayData(
+    @JsonProperty("division")
+    val division: String?,
+    @JsonProperty("events")
+    val events: List<BankHoliday>?
+)
+
+data class BankHoliday(
+    @JsonProperty("name")
+    val name: String?,
+    @JsonProperty("date")
+    val date: LocalDate?,
+    @JsonProperty("notes")
+    val notes: String?,
+    @JsonProperty("bunting")
+    val bunting: Boolean?
+)
--- a/src/main/kotlin/uk/gov/dluhc/printapi/client/BankHolidayDataClient.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/client/BankHolidayDataClient.kt
@@ -35,10 +35,10 @@ enum class BankHolidayDivision(val value: String) {
     companion object {
         @JvmStatic
         fun fromGssCode(gssCode: String): BankHolidayDivision {
-            return when (gssCode.substring(0, 1)) {
-                "E", "W" -> ENGLAND_AND_WALES
-                "S" -> SCOTLAND
-                "N" -> NORTHERN_IRELAND
+            return when (gssCode.first()) {
+                'E', 'W' -> ENGLAND_AND_WALES
+                'S' -> SCOTLAND
+                'N' -> NORTHERN_IRELAND
                 else -> {
                     logger.warn("Unknown country prefix for gssCode: [$gssCode], defaulting to $ENGLAND_AND_WALES")
                     ENGLAND_AND_WALES
--- a/src/main/kotlin/uk/gov/dluhc/printapi/client/BankHolidayDataFileClient.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/client/BankHolidayDataFileClient.kt
@@ -1,0 +1,49 @@
+package uk.gov.dluhc.printapi.client
+
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.databind.type.MapType
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.core.io.Resource
+import org.springframework.stereotype.Component
+import java.time.LocalDate
+import javax.annotation.PostConstruct
+
+@Component
+class BankHolidayDataFileClient(
+    @Value("classpath:bank-holidays.json") private val resource: Resource,
+    private val mapper: ObjectMapper
+) : BankHolidayDataClient {
+
+    private var bankHolidayData: Map<String, BankHolidayData> = emptyMap()
+
+    @PostConstruct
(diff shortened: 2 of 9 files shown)


sentioservices (approved):
LGTM

src/main/resources/application.yml
```diff
@@ -29,6 +29,8 @@ api:
     generate-temporary-certificate:
       valid-on-date:
         max-calendar-days-in-future: 30
+    retention.period:
+      certificate-delivery-info: "P28D" # 28 working days
```
vishalgupta-valtech, line 33: ℹ️ infra PR required as a pre-requisite 👍🏼

vishalgupta-valtech (approved):
LGTMV 👍🏼

src/main/kotlin/uk/gov/dluhc/printapi/client/BankHolidayDataFileClient.kt
```diff
@@ -0,0 +1,49 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper
+import com.fasterxml.jackson.databind.type.MapType
+import org.springframework.beans.factory.annotation.Value
+import org.springframework.core.io.Resource
+import org.springframework.stereotype.Component
+import java.time.LocalDate
+import javax.annotation.PostConstruct
+
+@Component
+class BankHolidayDataFileClient(
```
matt-wills-valtech, line 12: I've added a sub-task to [EIP1-3669](https://technologyprogramme.atlassian.net/browse/EIP1-3669) to improve this in future (so that we're retrieving up-to-date bank holiday data from one place).

src/main/kotlin/uk/gov/dluhc/printapi/client/BankHolidayDataFileClient.kt
```diff
@@ -0,0 +1,49 @@
...
+    @Value("classpath:bank-holidays.json") private val resource: Resource,
```
NathanRussellValtech, line 13: I'm sure this works, but is this the right syntax for an `@Value` annotation ??
```
@Value("classpath:bank-holidays.json") 
```
As I understand it the annotation is used to resolve SPEL expressions, and 99.9% of the time used to get property values. Therefore I would expect this line to be something like
```
@Value("\${bank-holidays.data-file.location}") 
```
or similar. And then it would be defined in `application.properties` as
```
bank-holidays:
    data-file:
        location: classpath:bank-holidays.json
```
I suspect its working more by luck than judgement (on the basis that `classpath:bank-holidays.json` is being resolved as a string literal SPEL expression)

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateRemovalDateResolver.kt
```diff
@@ -0,0 +1,58 @@
...
+     * @param gssCode The Certificate's GSS code which is used to determine if there are any upcoming bank holidays for
+     * the UK nation concerned.
+     * @return A [LocalDate] representing when a Certificate's delivery info should be removed
+     */
+    fun getCertificateDeliveryInfoRemovalDate(issueDate: LocalDate, gssCode: String): LocalDate =
+        with(getTotalDaysForWorkingDays(issueDate, dataRetentionConfig.certificateDeliveryInfo.toDays(), gssCode)) {
+            issueDate.plusDays(this.toLong())
+        }
+
+    private fun getTotalDaysForWorkingDays(issueDate: LocalDate, requiredWorkingDays: Long, gssCode: String): Int {
```
Neil-Massey, line 38: Could this return a `Long` instead of an `Int`? (This removes the  need of `toLong()` on line 35)

src/main/kotlin/uk/gov/dluhc/printapi/client/BankHolidayDataClient.kt
```diff
@@ -0,0 +1,70 @@
...
+ */
+enum class BankHolidayDivision(val value: String) {
+    ENGLAND_AND_WALES("england-and-wales"),
+    SCOTLAND("scotland"),
+    NORTHERN_IRELAND("northern-ireland");
+
+    companion object {
+        @JvmStatic
+        fun fromGssCode(gssCode: String): BankHolidayDivision {
+            return when (gssCode.substring(0, 1)) {
```
Neil-Massey, line 38: Could use `gssCode.first()`

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateRemovalDateResolver.kt
```diff
@@ -0,0 +1,58 @@
...
+     * period is currently 28 (configurable) working days from the certificate's "issue" date, which means weekends and
+     * bank holidays should be excluded when making this calculation. If the issue date was more than 28 working days
+     * ago, the removal date will be in the past.
+     *
+     * @param issueDate The date the Certificate was issued.
+     * @param gssCode The Certificate's GSS code which is used to determine if there are any upcoming bank holidays for
+     * the UK nation concerned.
+     * @return A [LocalDate] representing when a Certificate's delivery info should be removed
+     */
+    fun getCertificateDeliveryInfoRemovalDate(issueDate: LocalDate, gssCode: String): LocalDate =
```
NathanRussellValtech, line 33: Minor question about the design / use case of this method ....

The kdoc (which is truly excellent BTW 👍 ) refers to a `Certificate` . I wonder whether we mean this in a generic sense, or the entity? - ie the specific type of Certificate (I've read it and assumed the entity, but I could be wrong)

On the assumption that we mean `Certificate` as in the entity, then I think the kdoc should be generalised to describe Certificate as meaning the main certificate for a VAC, but also mention a Temporary Certificate and an AED.

Having said that I wonder if what might be a better approach is if we do actually refer to the entity types, and provide overloaded methods .... 🤔 

We could make this method `getCertificateDeliveryInfoRemovalDate` private, then perhaps have methods for each entity type:
```
fun getPersonalInformationRemovalDate(certificate: Certificate) : Int

fun getPersonalInformationRemovalDate(temporaryCertificate: TemporaryCertificate) : Int

fun getPersonalInformationRemovalDate(aed: AnonymousElectorDocument) : Int // we've not written this entity class yet, but you get the idea)
```

Within those public methods you can get the relevant fields (issueDate and gssCode) from the entity to be able to call the private method

What do you think?

src/test/kotlin/uk/gov/dluhc/printapi/client/BankHolidayDataFileClientTest.kt
```diff
@@ -0,0 +1,30 @@
+package uk.gov.dluhc.printapi.client
+
+import org.assertj.core.api.Assertions.assertThat
+import org.junit.jupiter.params.ParameterizedTest
+import org.junit.jupiter.params.provider.CsvSource
+import uk.gov.dluhc.printapi.client.BankHolidayDivision.ENGLAND_AND_WALES
+import uk.gov.dluhc.printapi.config.IntegrationTest
+import java.time.LocalDate
+
+internal class BankHolidayDataFileClientTest : IntegrationTest() {
```
NathanRussellValtech, line 10: Do you really want this to be an integration test?
Integration tests are slow, and all you're really doing is unit testing the `getBankHolidayDates` , based on data that is not in the control of the test (ie. the json file)

Does the design of the client class you are testing allow you to pass in a data file that might be a little closer to the test class?

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateRemovalDateResolver.kt
matt-wills-valtech: A long would be inappropriate.

src/main/kotlin/uk/gov/dluhc/printapi/client/BankHolidayDataClient.kt
matt-wills-valtech: Nice, done.

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateRemovalDateResolver.kt
matt-wills-valtech: It deliberately only refers to Certificate (as in the "business" type, rather than a specific class, though it's also the entity). The kdoc of the class:

```
Responsible for determining the removal date for a Certificate's data. The removal date varies, depending on the category of data and the type of Certificate concerned.
```

Whether "Certificate" is the right term to group these types together is another matter, but I think it works for `Certificate` and `TemporaryCertificate`.

It's difficult to know what lies in store for temp certs & AEDs, so I've not attempted to make this generic, but thought specific methods could be added, or refactoring could be done, as necessary.

src/main/kotlin/uk/gov/dluhc/printapi/client/BankHolidayDataFileClient.kt
matt-wills-valtech: Bear in mind it's a tactical solution so I was hoping not to modify application.yml, unless we have to.

I lifted this approach from the interweb somewhere, but happy to change it if necessary.

src/test/kotlin/uk/gov/dluhc/printapi/client/BankHolidayDataFileClientTest.kt
matt-wills-valtech: The idea is that each implementation of `BankHolidayDataClient` knows how to retrieve the data it needs. In this case, I want to test that `BankHolidayDataFileClient` reads and deserialises the JSON file correctly, in addition to returning the correct dates. We will have a different test when we start retrieving the bank holiday data from S3 (which we consider using Wiremock for).

src/main/kotlin/uk/gov/dluhc/printapi/client/BankHolidayDataFileClient.kt
matt-wills-valtech: It's a fair challenge as I know you'd normally use a `@Value` annotation as you said. I stole it from here: https://stackoverflow.com/a/58705928. I'll dig a bit deeper.

NathanRussellValtech (approved):
The PR roulette wheel has given you a reluctant approval from me 😁

matt-wills-valtech:
> The PR roulette wheel has given you a reluctant approval from me 😁

"reluctant"?! Bloody cheek of it! It's a thing of beauty 😉

PR #139, opened 2023-02-09
Feature/EIP1-3801 Listener to set initialRetentionRemovalDate

matt-wills-valtech:
This PR implements an SQS listener which sets the `initialRetentionRemovalDate` on the `Certificate` entity.

A scheduled job will remove relevant data (currently just the address PII data) from the Certificate's `PrintRequests` once we have reached the `initialRetentionRemovalDate`.


src/main/kotlin/uk/gov/dluhc/printapi/config/DataRetentionConfiguration.kt
```diff
@@ -7,5 +7,5 @@ import java.time.Duration
 @ConfigurationProperties(prefix = "api.print-api.retention.period")
 @ConstructorBinding
 data class DataRetentionConfiguration(
-    val certificateDeliveryInfo: Duration
+    val certificateInitialRetentionPeriod: Duration
```
matt-wills-valtech, line 10: Naming is hard, but hopefully there is sufficient kdoc where this is used (plus in the `Certificate` entity wrt `initialRetentionRemovalDate`) to explain what this means.

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateDataRetentionService.kt
```diff
@@ -0,0 +1,42 @@
...
+            val certificate = certificateRepository.findByGssCodeAndSourceTypeAndSourceReference(
+                gssCode = gssCode,
+                sourceType = sourceType,
+                sourceReference = sourceReference
+            )?.also {
+                it.initialRetentionRemovalDate =
+                    certificateRemovalDateResolver.getCertificateInitialRetentionPeriodRemovalDate(it.issueDate, gssCode)
+            } ?: throw CertificateNotFoundException(sourceType, sourceReference).also { logger.warn(it.message) }
+
+            certificateRepository.save(certificate)
```
matt-wills-valtech, line 39: I think this is unnecessary, but I prefer to be explicit.

src/main/kotlin/uk/gov/dluhc/printapi/config/DataRetentionConfiguration.kt
NathanRussellValtech: 💯 👍 
My only comment here would be re: class type. I think the variable name is accurate in that its a period of time
There are 2 classes related to this - `java.time.Duration` and `java.time.Period` , but the variable name doesnt match the class used
I don't know the difference between a `Duration` and a `Period` - and I suspect changing the class type down the stack to a `Period` might be a pain?, so maybe the variable name is a `certificateInitialRetentionDuration` ??

I don't know - like you said, naming is hard!

I tell you what, let's pretend I never mentioned this, and we'll leave it well alone !! 😁 👍 🤣

src/main/kotlin/uk/gov/dluhc/printapi/exception/CertificateNotFoundException.kt
```diff
@@ -2,10 +2,10 @@ package uk.gov.dluhc.printapi.exception
...
 
-class CertificateNotFoundException(eroId: String, sourceType: SourceType, sourceReference: String) :
-    RuntimeException(
-        if (sourceType == SourceType.VOTER_CARD)
-            "Certificate for eroId = $eroId and application id = $sourceReference not found"
-        else
-            "Certificate for eroId = $eroId, sourceType = $sourceType and $sourceReference not found"
-    )
+class CertificateNotFoundException : RuntimeException {
+    constructor(eroId: String, sourceType: SourceType, sourceReference: String) :
```
NathanRussellValtech, line 6: Overloaded constructor - nice 👍 
I'd forgotten the syntax for that because we don't use it often ; so good to see a nice example of it 👍 
Thanks

src/main/kotlin/uk/gov/dluhc/printapi/messaging/ApplicationRemovedMessageListener.kt
```diff
@@ -0,0 +1,28 @@
...
+/**
+ * Implementation of [MessageListener] to handle [ApplicationRemovedMessage] messages
+ */
+@Component
+class ApplicationRemovedMessageListener(
+    val certificateDataRetentionService: CertificateDataRetentionService
+) : MessageListener<ApplicationRemovedMessage> {
+
+    @SqsListener("\${sqs.application-removed-queue-name}")
+    override fun handleMessage(@Valid @Payload payload: ApplicationRemovedMessage) {
```
NathanRussellValtech, line 22: Just so I'm clear .... a `ApplicationRemovedMessage` is for a VAC Application? And this listener can also be used to handle doing the data retention for a VAC Application's Temporary Certificate(s) as well as and when we get there?

Does that mean we are proposing other message types in the future, such as `AnonymousApplicationRemovedMessage`, `PostalApplicationRemovedMessage`, `OverseasApplicationRemovedMessage` etc? (not that those last 2 have 'certificates' I dont think, but you get the idea)

Or, does `ApplicationRemovedMessage` have a field in it to say whether its a regular VAC or an Anonymous application ?

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateDataRetentionService.kt
```diff
@@ -0,0 +1,42 @@
...
+    private val certificateRemovalDateResolver: CertificateRemovalDateResolver
+) {
+    /**
+     * Sets the initialRetentionRemovalDate on a [uk.gov.dluhc.printapi.database.entity.Certificate], after the
+     * originating application is removed from the source system (e.g. VCA).
+     *
+     * @param message An [ApplicationRemovedMessage] sent from the source system.
+     */
+    @Transactional
+    fun handleSourceApplicationRemoved(message: ApplicationRemovedMessage) {
```
NathanRussellValtech, line 26: Rightly or wrongly I think our normal pattern is that the service method will take a DTO rather than the SQS interface model type, which means your listener class above needs to map the model message into a DTO

Any thoughts on that approach ?

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateDataRetentionService.kt
```diff
@@ -0,0 +1,42 @@
...
+        with(message) {
+            val sourceType = sourceTypeMapper.mapSqsToEntity(sourceType)
+            val certificate = certificateRepository.findByGssCodeAndSourceTypeAndSourceReference(
+                gssCode = gssCode,
+                sourceType = sourceType,
+                sourceReference = sourceReference
+            )?.also {
+                it.initialRetentionRemovalDate =
+                    certificateRemovalDateResolver.getCertificateInitialRetentionPeriodRemovalDate(it.issueDate, gssCode)
+            } ?: throw CertificateNotFoundException(sourceType, sourceReference).also { logger.warn(it.message) }
```
NathanRussellValtech, line 37: We might want to think about the handling of this exception.
I think you're right to handle the eventuality that we might not find the certificate in the repository, but this service method is being called by a listener, which means (unless you catch and swallow it in the listener) the SQS message will fail, and the message will be re-queued and re-tried up to the retry limit (6 times ?), then DLQ'ed

If the certificate is not found in the repository, is there any milage in trying again? Do we think it might find the certificate in the database on a subsequent retry ?

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateDataRetentionService.kt
NathanRussellValtech: Does that mean it will get "saved" twice? Once here, and once at the TX boundary ?
(I know it will only get committed once at the TX boundary .... but what about the JPA managed fields? specifically version? Could the version number jump up 2 versions ?)

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateRemovalDateResolver.kt
```diff
@@ -20,18 +20,22 @@ class CertificateRemovalDateResolver(
 ) {
 
     /**
-     * Calculates the date that the delivery info for a Certificate's print requests should be removed. The retention
-     * period is currently 28 (configurable) working days from the certificate's "issue" date, which means weekends and
-     * bank holidays should be excluded when making this calculation. If the issue date was more than 28 working days
-     * ago, the removal date will be in the past.
+     * Calculates the date that certain [uk.gov.dluhc.printapi.database.entity.Certificate] related data should be
```
NathanRussellValtech, line 23: Thank you; good clarification 👍

src/test/kotlin/uk/gov/dluhc/printapi/logging/CorrelationIdMdcIntegrationTest.kt
```diff
@@ -130,7 +131,7 @@ internal class CorrelationIdMdcIntegrationTest : IntegrationTest() {
             batchPrintRequestsJob.run()
 
             // Then
-            await.atMost(3, TimeUnit.SECONDS).untilAsserted {
+            await.atMost(5, TimeUnit.SECONDS).untilAsserted {
```
NathanRussellValtech, line 134: Thank you 👍

src/test/kotlin/uk/gov/dluhc/printapi/messaging/ApplicationRemovedMessageListenerTest.kt
```diff
@@ -0,0 +1,44 @@
...
+                buildPrintRequest(delivery = buildDelivery()),
+                buildPrintRequest(delivery = buildDelivery())
+            )
+        )
+        certificateRepository.save(certificate)
+        val payload = buildApplicationRemovedMessage(
+            sourceReference = certificate.sourceReference!!,
+            gssCode = certificate.gssCode!!
+        )
+        val expectedInitialRemovalDate = LocalDate.of(2023, 5, 16)
```
NathanRussellValtech, line 31: Minor, and I know you kinda mentioned it above, but might be worth a comment on the expectedInitialRemovalDate explaining why its 16th May 2023
```
// Certificate issue date is 01/04/2023. 28 working days from that date is 16/05/23 because there are xx weekend days, plus xx bank holidays (Good Friday and Easter Monday)
```
Something like that?

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateDataRetentionService.kt
matt-wills-valtech: Some cleverness happens and no it doesn't bump the version twice if you call `save()` :)

This answers it better than I can: https://stackoverflow.com/a/46708295

src/test/kotlin/uk/gov/dluhc/printapi/messaging/ApplicationRemovedMessageListenerTest.kt
```diff
@@ -0,0 +1,44 @@
...
+
+        // When
+        sqsMessagingTemplate.convertAndSend(applicationRemovedQueueName, payload)
+
+        // Then
+        await.atMost(5, TimeUnit.SECONDS).untilAsserted {
+            val response = certificateRepository.findAll()
+            assertThat(response).hasSize(1)
+            val saved = response[0]
+            assertThat(saved.initialRetentionRemovalDate).isEqualTo(expectedInitialRemovalDate)
```
NathanRussellValtech, line 41: Minor, we already have a `CertificateAssert` class (`uk.gov.dluhc.printapi.testsupport.assertj.assertions.CertificateAssert`), so perhaps you could add a `hasInitialRetentionRemovalDate` method on it and use that?

src/main/kotlin/uk/gov/dluhc/printapi/messaging/ApplicationRemovedMessageListener.kt
matt-wills-valtech: I considered including a type field, but I don't think there's any data that needs to be removed after the initial retention period for temp certs or AEDs (since they have no address). I didn't want to overthink it, so decided I could always add a type later if necessary.

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/testdata/entity/CertificateBuilder.kt
```diff
@@ -61,6 +62,7 @@ fun buildCertificate(
     sourceReference: String = aValidSourceReference(),
     applicationReceivedDateTime: Instant = aValidApplicationReceivedDateTime(),
     applicationReference: String = aValidApplicationReference(),
+    issueDate: LocalDate = aValidIssueDate()
```
NathanRussellValtech, line 65: 👍

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateDataRetentionService.kt
matt-wills-valtech: I had the same thought, but followed the approach Squad A have already established in this repo 🔥 😉

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateDataRetentionService.kt
NathanRussellValtech: 👍 Cool, thanks

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateDataRetentionService.kt
NathanRussellValtech: 😮

src/main/kotlin/uk/gov/dluhc/printapi/messaging/ApplicationRemovedMessageListener.kt
NathanRussellValtech: Thats fair. Thanks 👍

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateDataRetentionService.kt
matt-wills-valtech: That's a good point - I'll change it to log an error. However, I expect (and sincerely hope) this never happens!

src/main/kotlin/uk/gov/dluhc/printapi/config/DataRetentionConfiguration.kt
matt-wills-valtech: A valid observation. Changed to `Period` as it's better suited anyway (since we're not interested in the time, only the number of days).

src/test/kotlin/uk/gov/dluhc/printapi/messaging/ApplicationRemovedMessageListenerTest.kt
matt-wills-valtech: Sorry but I don't think that's necessary (given I alluded to it above and I've added quite a bit of kdoc in places).

NathanRussellValtech: We might have to agree to disagree on that 😁 👍 
Yes you've alluded to it above, yes you have plenty of really good kdoc; but from reading the test alone I'm not sure it's as obvious as it could be ¯\_(ツ)_/¯
But I'm happy to stand down on this 😁

NathanRussellValtech (approved):
LGTM; a few minor comments, but there's some really nice stuff in here 👍 
Thanks

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepository.kt
```diff
@@ -19,6 +19,8 @@ interface CertificateRepository : JpaRepository<Certificate, UUID> {
 
     fun findByStatusOrderByApplicationReceivedDateTimeAsc(certificateStatus: Status, pageable: Pageable): List<Certificate>
 
+    fun findByGssCodeAndSourceTypeAndSourceReference(gssCode: String, sourceType: SourceType, sourceReference: String): Certificate?
```
vishalgupta-valtech, line 22: 👍🏼

src/test/kotlin/uk/gov/dluhc/printapi/messaging/ApplicationRemovedMessageListenerTest.kt
matt-wills-valtech: Done, but to be honest it felt a bit overkill at this stage since a) this is currently only used in one place and b) I had to add this sizeable method to our Assertions class:

```
    /**
     * Verifies that the actual Certificate's initialRetentionRemovalDate is equal to the given one.
     * @param initialRetentionRemovalDate the given initialRetentionRemovalDate to compare the actual Certificate's one.
     * @return this assertion object.
     * @throws AssertionError - if the actual Certificate's initialRetentionRemovalDate is not equal to the given one.
     */
    fun hasInitialRetentionRemovalDate(initialRetentionRemovalDate: LocalDate?): CertificateAssert {
        // check that actual Certificate we want to make assertions on is not null.
        isNotNull

        // overrides the default error message with a more explicit one
        val assertjErrorMessage = "\nExpecting initialRetentionRemovalDate of:\n  <%s>\nto be:\n  <%s>\nbut was:\n  <%s>"

        // null safe check
        val actualInitialRetentionRemovalDate = actual!!.initialRetentionRemovalDate
        if (!Objects.deepEquals(actualInitialRetentionRemovalDate, initialRetentionRemovalDate)) {
            failWithMessage(assertjErrorMessage, actual, initialRetentionRemovalDate, actualInitialRetentionRemovalDate)
        }

        // return the current assertion for method chaining
        return this
    }
```

That's not to say I'm not a fan of these custom assertion methods btw - I think they read really nicely - just that I wouldn't normally add them for one off use cases. If we start using it in more places though then it's definitely worth it.

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateRemovalDateResolver.kt
vishalgupta-valtech: +1

src/main/resources/application.yml
```diff
@@ -23,14 +23,15 @@ sqs:
...
+  application-removed-queue-name: ${SQS_APPLICATION_REMOVED_QUEUE_NAME}
 
 api:
   print-api:
     generate-temporary-certificate:
       valid-on-date:
         max-calendar-days-in-future: 30
     retention.period:
-      certificate-delivery-info: "P28D" # 28 working days
+      certificate-initial-retention-period: "P28D" # 28 working days
```
vishalgupta-valtech, line 34: Could this be simplified `retention.period.certificate-initial-retention-period`?

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/testdata/model/SourceApplicationRemovedMessageBuilder.kt
```diff
@@ -0,0 +1,16 @@
+package uk.gov.dluhc.printapi.testsupport.testdata.model
+
+import uk.gov.dluhc.printapi.messaging.models.ApplicationRemovedMessage
+import uk.gov.dluhc.printapi.messaging.models.SourceType
+import uk.gov.dluhc.printapi.testsupport.testdata.aValidSourceReference
+import uk.gov.dluhc.printapi.testsupport.testdata.getRandomGssCode
+
+fun buildApplicationRemovedMessage(
```
vishalgupta-valtech, line 8: Could rename the class name to be ApplicationRemovedMessageBuilder as we took `Source` out of it's name

vishalgupta-valtech (approved):
LGTM 🚀

src/main/resources/application.yml
matt-wills-valtech: You mean place on one line? I was thinking other retention periods could be added in the future - e.g. for temporary certs.

sentioservices (approved):
LGTM

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateDataRetentionService.kt
```diff
@@ -0,0 +1,40 @@
...
+        with(message) {
+            val sourceType = sourceTypeMapper.mapSqsToEntity(sourceType)
+            certificateRepository.findByGssCodeAndSourceTypeAndSourceReference(
+                gssCode = gssCode,
+                sourceType = sourceType,
+                sourceReference = sourceReference
+            )?.also {
+                it.initialRetentionRemovalDate = certificateRemovalDateResolver
+                    .getCertificateInitialRetentionPeriodRemovalDate(it.issueDate, gssCode)
+                certificateRepository.save(it)
```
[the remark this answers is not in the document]
matt-wills-valtech: Indeed, but I've already left a comment about that (if you can find it amongst all the ones Nathan left! 😆).

PR #140, opened 2023-02-09
EIP1-3439 / EIP1-3776 - Use real Explainer Documents

NathanRussellValtech:
This PR replaces the previous Temporary Certificate Explainer Document templates (our dev versions) with the real versions supplied by Malcolm Seaborn (UX / UCD)

The resultant PDFs with the fields populated look like this:

### English
Screenshot 2023-02-09 at 19 13 06

### Welsh
Screenshot 2023-02-09 at 19 13 29


src/main/resources/application.yml
```diff
@@ -91,25 +91,25 @@ temporary-certificate:
     english:
       path: "classpath:temporary-certificate-template/Explainer Document (English).pdf"
       placeholder:
-        ero-name: "eroName"
```
NathanRussellValtech, line 94: The field names in the template PDFs have changed (unsurprisingly)

src/test/resources/application-integration-test.yml
```diff
@@ -22,6 +22,9 @@ spring:
         format_sql : false
         show_sql : false
 
+  codec:
```
NathanRussellValtech, line 25: We need this new property in the integration test spring context (`spring.codec.max-in-memory-size`) to allow the `WebTestClient` that is used in the integration tests to handle a large response body
(we have an integration test that generates and downloads the explainer doc - now we have the real explainer document template PDFs, the resultant http response is larger (almost 1MB !!)))

src/main/kotlin/uk/gov/dluhc/printapi/service/temporarycertificate/PdfFactory.kt
```diff
@@ -16,6 +16,7 @@ class PdfFactory {
         PdfReader(ResourceUtils.getFile(templateDetails.path).inputStream()).use { reader ->
             val stamper = PdfStamper(reader, outputStream)
             stamper.cleanMetadata()
+            stamper.setFormFlattening(true)
```
NathanRussellValtech, line 19: Flatten the form fields when the stamper is closed, meaning the resultant PDF has the field values, but they are no longer editable form fields (dont want the vc-admin or someone creating a PDF, then being able to edit it themselves afterwards!)

src/test/kotlin/uk/gov/dluhc/printapi/rest/GenerateTemporaryCertificateExplainerDocumentIntegrationTest.kt
```diff
@@ -148,8 +147,5 @@ internal class GenerateTemporaryCertificateExplainerDocumentIntegrationTest : In
 
         val pdfContent = result.responseBody
         assertThat(pdfContent).isNotNull
-        PdfReader(pdfContent).use { reader ->
```
NathanRussellValtech, line 151: The resultant PDF that is generated and returned from the REST API no longer has form fields

src/test/resources/application-integration-test.yml
NathanRussellValtech: I would agree that your approach is more explicit. The property I have used is not exactly self-explanatory !!
I'll change it to use your approach 👍

NathanRussellValtech: Done

src/main/resources/application.yml
Neil-Massey: It's almost as if we knew that would happen and went to the effort of externalising their names 😄

src/main/kotlin/uk/gov/dluhc/printapi/service/temporarycertificate/PdfFactory.kt
Neil-Massey: Nice - wasn't aware of this feature; have updated my PR to include this change 👍

matt-wills-valtech: Minor - perhaps add a comment to that effect?

matt-wills-valtech (approved):

adacayi (approved):

Neil-Massey (approved):
LGTM 👍

PR #142, opened 2023-02-10
EIP1-511 - openAPI doc for endpoint to get Temporary Certificates for an application

NathanRussellValtech:
This PR adds the new API spec for the REST API to get the Temporary Certificate summaries for an application

Screenshot 2023-02-10 at 14 06 46

### DIFF
--- a/src/main/resources/openapi/PrintAPIs.yaml
+++ b/src/main/resources/openapi/PrintAPIs.yaml
@@ -488,11 +488,23 @@ components:
           format: date-time
           description: The timestamp in ISO-8601 UTC that the Temporary Certificate was generated
           example: '2022-11-17T10:53:25Z'
+        issueDate:
+          type: string
+          format: date
+          description: The date that the Temporary Certificate was issued
+          example: '2022-11-17'
+        validOnDate:
+          type: string
+          format: date
+          description: The date on which the Temporary Certificate is/was valid
+          example: '2022-11-19'
       required:
         - certificateNumber
         - status
         - userId
         - dateTime
+        - issueDate
+        - validOnDate
     TemporaryCertificateStatus:
       title: TemporaryCertificateStatus
       type: string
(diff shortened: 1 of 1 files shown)


src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -375,6 +454,51 @@ components:
...
+          description: The unique identifier of the Certificate. A 20 character alphanumeric string.
+          example: DEV1W0XDH1368LWE1J40
+        status:
+          $ref: '#/components/schemas/TemporaryCertificateStatus'
+        userId:
+          type: string
+          format: email
+          example: fred.blogs@some-domain.co.uk
+          description: UserId of the user who generated the Temporary Certificate
+        dateTime:
```
adacayi, line 486: Is this the `issueDate` @NathanRussellValtech ? If the answer is yes, naming it as `issueDate` will make it more clear in my opinion. `dateTime` seems too generic and I wasn't able to figure it out without reading the description.

adacayi:
@NathanRussellValtech , do you think `validOnDate` will be useful in the response as well?

NathanRussellValtech:
> @NathanRussellValtech , do you think `validOnDate` will be useful in the response as well?

I think you're right @abdullah-sanver-valtech - I think we want both `validOnDate` and `issueDate` (rather than `dateTime`)
I'll make that change now 👍

NathanRussellValtech:
> > @NathanRussellValtech , do you think `validOnDate` will be useful in the response as well?
> 
> I think you're right @abdullah-sanver-valtech - I think we want both `validOnDate` and `issueDate` (rather than `dateTime`) I'll make that change now 👍

@abdullah-sanver-valtech - OK, I've just pushed a commit that adds `issueDate` and `validOnDate` as per your suggestions.
Both of these are dates, rather than date times. I've left the original `dateTime` field as actually its correct, but I've renamed it to `dateTimeGenerated` to make it clear. It reflects the date and time that the Temp Cert was generated, which is (or could be) different to the `issueDate`, and almost certainly will be different to the `validOnDate`

adacayi (approved):
LGTM, @NathanRussellValtech ,
Thanks for addressing the comments.

PR #147, opened 2023-02-14
EIP1-511 - Implemented controller through to service and repository

NathanRussellValtech:
This PR implements the main controller method, through service, through repository; to implement the happy path for the REST API to get Temporary Certificate Summaries for a given application

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/TemporaryCertificate.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/TemporaryCertificate.kt
@@ -128,6 +128,11 @@ class TemporaryCertificate(
     val status: TemporaryCertificateStatus.Status?
         get() = statusHistory.sortedByDescending { it.dateCreated }.first().status
 
+    val dateTimeGenerated: Instant?
+        get() = statusHistory
+            .sortedBy { it.dateCreated }
+            .first { it.status == TemporaryCertificateStatus.Status.GENERATED }.dateCreated
+
     override fun equals(other: Any?): Boolean {
         if (this === other) return true
         if (other == null || Hibernate.getClass(this) != Hibernate.getClass(other)) return false
--- a/src/main/kotlin/uk/gov/dluhc/printapi/rest/TemporaryCertificateController.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/rest/TemporaryCertificateController.kt
@@ -13,18 +13,23 @@ import org.springframework.web.bind.annotation.PathVariable
 import org.springframework.web.bind.annotation.PostMapping
 import org.springframework.web.bind.annotation.RequestBody
 import org.springframework.web.bind.annotation.RestController
+import uk.gov.dluhc.printapi.database.entity.SourceType
 import uk.gov.dluhc.printapi.dto.PdfFile
 import uk.gov.dluhc.printapi.mapper.GenerateTemporaryCertificateMapper
+import uk.gov.dluhc.printapi.mapper.TemporaryCertificateSummaryMapper
 import uk.gov.dluhc.printapi.models.GenerateTemporaryCertificateRequest
 import uk.gov.dluhc.printapi.models.TemporaryCertificateSummariesResponse
 import uk.gov.dluhc.printapi.service.temporarycertificate.ExplainerPdfService
 import uk.gov.dluhc.printapi.service.temporarycertificate.TemporaryCertificateService
+import uk.gov.dluhc.printapi.service.temporarycertificate.TemporaryCertificateSummaryService
 import java.io.ByteArrayInputStream
 import javax.validation.Valid
 
 @RestController
 @CrossOrigin
 class TemporaryCertificateController(
+    private val temporaryCertificateSummaryService: TemporaryCertificateSummaryService,
+    private val temporaryCertificateSummaryMapper: TemporaryCertificateSummaryMapper,
     private val explainerPdfService: ExplainerPdfService,
     private val temporaryCertificateService: TemporaryCertificateService,
     private val generateTemporaryCertificateMapper: GenerateTemporaryCertificateMapper,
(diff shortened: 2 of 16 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/database/entity/TemporaryCertificate.kt
```diff
@@ -128,6 +128,11 @@ class TemporaryCertificate(
     val status: TemporaryCertificateStatus.Status?
         get() = statusHistory.sortedByDescending { it.dateCreated }.first().status
 
+    val dateTimeGenerated: Instant?
```
Neil-Massey, line 131: Does this have to be optional? `Instant?`

src/main/kotlin/uk/gov/dluhc/printapi/rest/TemporaryCertificateController.kt
```diff
@@ -35,9 +40,18 @@ class TemporaryCertificateController(
...
+    ): TemporaryCertificateSummariesResponse =
+        TemporaryCertificateSummariesResponse(
+            temporaryCertificateSummaryService.getTemporaryCertificateSummaries(
+                eroId,
+                SourceType.VOTER_CARD,
+                applicationId
+            ).map {
+                temporaryCertificateSummaryMapper.toApiTemporaryCertificateSummary(it)
+            }.sortedByDescending {
+                it.dateTimeGenerated
```
Neil-Massey, line 52: seems like a UI concern?

src/test/kotlin/uk/gov/dluhc/printapi/rest/GetTemporaryCertificateSummariesByApplicationIdIntegrationTest.kt
```diff
@@ -33,4 +42,76 @@ internal class GetTemporaryCertificateSummariesByApplicationIdIntegrationTest :
...
+
+        val temporaryCertificate1 = buildTemporaryCertificate(
+            sourceType = SourceType.VOTER_CARD,
+            sourceReference = APPLICATION_ID,
+            gssCode = eroResponse.localAuthorities[0].gssCode,
+        ).also {
+            temporaryCertificateRepository.save(it)
+        }
+
+        Thread.sleep(1000) // sleep for 1 second to ensure that the 2 temp certs are saved at different times and we can therefore predict and assert the order
```
Neil-Massey, line 61: A whole second ... maybe can update the entity directly? (may not be worth the effort)

src/test/kotlin/uk/gov/dluhc/printapi/service/temporarycertificate/TemporaryCertificateFinderServiceTest.kt
```diff
@@ -0,0 +1,62 @@
...
+    private lateinit var eroService: EroService
+
+    @Mock
+    private lateinit var temporaryCertificateRepository: TemporaryCertificateRepository
+
+    @InjectMocks
+    private lateinit var temporaryCertificateFinderService: TemporaryCertificateFinderService
+
+    @Test
+    fun `should get TemporaryCertificates given one exists for the provided details`() {
```
Neil-Massey, line 33: or if none exist (nothing specific in function whether one is found or not). (`should get TemporaryCertificates given the provided details`)

src/test/kotlin/uk/gov/dluhc/printapi/rest/GetTemporaryCertificateSummariesByApplicationIdIntegrationTest.kt
NathanRussellValtech: Yeah, I couldn't remember the best way to do this (or way's we've done it in other places such as VCA)
In the end I decided it wasnt worth the effort - sleep for a second (yup, I know! 🙄 ), then assert the order and most of the fields, but not the dateTimeGenerated ¯\_(ツ)_/¯

src/main/kotlin/uk/gov/dluhc/printapi/service/temporarycertificate/TemporaryCertificateSummaryService.kt
```diff
@@ -0,0 +1,25 @@
...
+
+    fun getTemporaryCertificateSummaries(eroId: String, sourceType: SourceType, sourceReference: String):
+        List<TemporaryCertificateSummaryDto> {
+        val temporaryCertificates = temporaryCertificateFinderService.getTemporaryCertificates(
+            eroId,
+            sourceType,
+            sourceReference
+        )
+        return temporaryCertificates.map {
+            mapper.toDtoTemporaryCertificateSummary(it)
```
Neil-Massey, line 22: Still struggle with the `entity` -> `dto` -> `Api` mapping chain 😄

src/main/kotlin/uk/gov/dluhc/printapi/rest/TemporaryCertificateController.kt
NathanRussellValtech: Yes and no ....
For most of our APIs (in all our apps) that return a collection of things where a date is important, the UI have asked that we return them in a sorted order. It helps them; and I kinda get it.
Sure, its a UI concern .... but it also feels like an API concern in that it should be part of our implementation to return things in a given sort order.
I've implemented the sorting here at the controller - its the closest point to the UI, so it feels like the right approach?

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/TemporaryCertificate.kt
NathanRussellValtech: Technically, yes; in practice, no.
As with the `status` field above, its based on an element in the statusHistory collection. There is no guarantee that the collection will not be null, or wont be empty, or in this case will always have a `GENERATED` status.

src/main/kotlin/uk/gov/dluhc/printapi/service/temporarycertificate/TemporaryCertificateSummaryService.kt
NathanRussellValtech: You love it really 😁

src/main/kotlin/uk/gov/dluhc/printapi/rest/TemporaryCertificateController.kt
Neil-Massey: Agree - I liked that it's in the controller (rather than in the service layer) 👍

src/test/kotlin/uk/gov/dluhc/printapi/rest/GetTemporaryCertificateSummariesByApplicationIdIntegrationTest.kt
Neil-Massey: Print is small enough a test suite it's cool 👍

src/test/kotlin/uk/gov/dluhc/printapi/service/temporarycertificate/TemporaryCertificateFinderServiceTest.kt
NathanRussellValtech: OK, will do that 👍

NathanRussellValtech: Done

PR #157, opened 2023-02-17
EIP1-3824 Anonymous Elector Document (AED) persistence

Neil-Massey:


src/main/resources/db/changelog/ddl/0013_create_anonymous_elector_document_tables.xml
```diff
@@ -0,0 +1,151 @@
...
+            <column name="electoral_number" type="varchar(20)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="aed_template_filename" type="varchar(255)">
+                <constraints nullable="false"/>
+            </column>
+            <column name="issue_date" type="date">
+                <constraints nullable="false"/>
+            </column>
+            <column name="request_date_time" type="timestamp">
```
Neil-Massey, line 90: Not sure how this will differ from `date_created` - and therefore whether it's required

src/main/resources/db/changelog/ddl/0013_create_anonymous_elector_document_tables.xml
```diff
@@ -0,0 +1,151 @@
...
+                <constraints nullable="false"/>
+            </column>
+            <column name="user_id" type="varchar(255)">
```
Neil-Massey, line 93: Not sure how this will differ from `created_by ` - and therefore whether it's required

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AedPrintRequest.kt
```diff
@@ -0,0 +1,101 @@
...
+    @field:NotNull
+    @field:Size(max = 255)
+    var userId: String? = null,
+
+    @OneToMany(cascade = [CascadeType.ALL], orphanRemoval = true, fetch = FetchType.EAGER)
+    @JoinColumn(name = "print_request_id", nullable = false)
+    var statusHistory: MutableList<AedPrintRequestStatus> = mutableListOf(),
+
+    // managed using JPA pre-persist annotated method below
+    var dateCreated: Instant? = null,
```
matt-wills-valtech, line 61: Out of interest, why are we using the pre-persist method and why do you have `@CreationTimestamp` on `AedPrintRequestStatus` instead?

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AedPrintRequestStatus.kt
```diff
@@ -0,0 +1,68 @@
...
+    override fun hashCode(): Int = javaClass.hashCode()
+
+    @Override
+    override fun toString(): String {
+        return this::class.simpleName + "(id = $id , status = $status , dateCreated = $dateCreated , createdBy = $createdBy , version = $version )"
+    }
+
+    enum class Status {
+        GENERATED,
+        PRINTED,
```
matt-wills-valtech, line 66: Is the record only ever created after the document is successfully generated?

NathanRussellValtech (approved):
LGTM, other than the changeset file not being included in the changelog master? Don't know if that was deliberate or not?

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
```diff
@@ -0,0 +1,134 @@
...
+
+    @field:Size(max = 255)
+    var middleNames: String? = null,
+
+    @field:NotNull
+    @field:Size(max = 255)
+    var surname: String? = null,
+
+    @OneToOne(cascade = [CascadeType.ALL])
+    var address: Address,
```
matt-wills-valtech, line 80: I wonder if we should put the name fields and possibly the address within a separate object - e.g. `PersonalDetails`, or something like that? My understanding is that we need to keep this data for re-prints after the "application" data has been removed from VCA. I'm about to start refining the data retention side of things for AEDs, but I have a feeling we need to remove this personal data sooner (after 15 months?) than the data that's on the document itself. In which case, it will need to be nullable.

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AedPrintRequestStatus.kt
Neil-Massey: Yes - `AedPrintRequest` would only be added to the `Anonymous_Elector_Document` once the PDF is generated.  If a request comes for a certificate to be reissued a new `AedPrintRequest` and PDF are created.

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
Neil-Massey: As an embeddable or an `Entity` (similar to the `Delivery` that has separate table and entity)?
I considered both but wasn't sure modelling the tables/entities for data retention policy was a valid approach.  If we didn't have an existing `Address` Entity would have probably have modelled as an embeddable.

But given you're also questioning I think it's probably a valid approach - certainly the data is a logical unit 👍 Probably will call it Contact Details and will include phone number and email too 👍

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AedPrintRequest.kt
Neil-Massey: It was how it was defined in the `PrintRequest` that this is copied from and it was needed there as the created date is used for sorting and it needs to be populated earlier than the `@CreationTimestamp` is populated.  Although we can use the `requestDateTime` in this flow - will refactor 👍

Neil-Massey:
> LGTM, other than the changeset file not being included in the changelog master? Don't know if that was deliberate or not?

Fixed 👍

PR #163, opened 2023-02-20
Feature/EIP1-4053 Add initial retention data removed column

matt-wills-valtech:


src/main/resources/db/changelog/ddl/0013_initial_retention_data_removed_column.xml
```diff
@@ -0,0 +1,37 @@
...
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+
+    <changeSet author="matt.wills@valtech.com" id="0013_EIP1-4053_alter_certificate - add initial_retention_data_removed column" context="ddl">
+
+        <addColumn tableName="certificate">
+            <column name="initial_retention_data_removed" type="boolean" value="false"
+                    remarks="Indicates if initial retention period data has been removed from a certificate">
+                <constraints nullable="true"/>
```
sentioservices, line 13: should this be nullable? The kotlin class attribute isn't optional.

src/main/resources/db/changelog/ddl/0013_initial_retention_data_removed_column.xml
```diff
@@ -0,0 +1,37 @@
...
+
+        <dropIndex tableName="certificate" indexName="certificate_initial_retention_removal_date_idx"/>
+
+        <createIndex tableName="certificate" indexName="certificate_initial_retention_removal_idx">
+            <column name="source_type"/>
+            <column name="initial_retention_data_removed"/>
+            <column name="initial_retention_removal_date"/>
+        </createIndex>
+
+        <rollback>
```
Neil-Massey, line 25: Does `certificate_initial_retention_removal_idx` need dropping before column can be dropped?

sentioservices (approved):
LGTM, just a query wrt the nullablility of the new column

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
```diff
@@ -81,6 +81,11 @@ class Certificate(
      */
     var initialRetentionRemovalDate: LocalDate? = null,
 
+    /**
+     * Set to true after the initial retention period data is removed.
+     */
+    var initialRetentionDataRemoved: Boolean = false,
```
Neil-Massey, line 87: Does this need to be `Boolean?` - asking as noticed the Liquibase column allows nulls.

Neil-Massey (approved):
LGTM - Just question on null handling

src/main/resources/db/changelog/ddl/0013_initial_retention_data_removed_column.xml
sentioservices: Good point. According to the mysql docs it looks like the column is just removed from the index, so it would cause issues when the changeset is tried on subsequent occasions.

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
matt-wills-valtech: Made column non-nullable.

src/main/resources/db/changelog/ddl/0013_initial_retention_data_removed_column.xml
matt-wills-valtech: Done.

sentioservices (approved):
LGTM, thanks for addressing comments

PR #164, opened 2023-02-20
Feature/EIP1-4053 Add initial retention data removed column (#163)

matt-wills-valtech:
Apply new initial retention data removed column to hotfix


vishalgupta-valtech (approved):
LGTM, I can't think of a better column name 😉

sentioservices (approved):
LGTM

PR #165, opened 2023-02-20
EIP1-4053 - Add DeliveryRepository to ensure Delivery data is removed

matt-wills-valtech:
More in-depth testing has revealed it's not enough just to set various values to `null` and then call `save()` on the `CertificateRepository`. That's not has how `CascadeType.ALL` works and is something I was always unsure about (I know, I should have tested this more thoroughly to begin with but haste got the better of me! 🙄 ).

Instead I've reluctantly introduced `DeliveryRepository` so I can call `delete()` on each one. This subsequently removes all child `address` entities as well (i.e. exactly how `CascadeType.ALL` is intended to work).

### DIFF
--- a/src/test/kotlin/uk/gov/dluhc/printapi/jobs/InitialRetentionPeriodDataRemovalJobIntegrationTest.kt
+++ b/src/test/kotlin/uk/gov/dluhc/printapi/jobs/InitialRetentionPeriodDataRemovalJobIntegrationTest.kt
@@ -19,6 +19,10 @@ internal class InitialRetentionPeriodDataRemovalJobIntegrationTest : Integration
         val certificate3 = buildCertificate(initialRetentionRemovalDate = LocalDate.now())
         val certificate4 = buildCertificate(initialRetentionRemovalDate = LocalDate.now().plusDays(1))
         certificateRepository.saveAll(listOf(certificate1, certificate2, certificate3, certificate4))
+        val deliveryId1 = certificate1.printRequests[0].delivery!!.id!!
+        val deliveryId2 = certificate2.printRequests[0].delivery!!.id!!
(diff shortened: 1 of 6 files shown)


NathanRussellValtech (approved):
LGTM - I understand the objective here is not only to remove the `Delivery` entity from the `PrintRequest` entity (ie. by nulling the property which at the DB level will null the linking ID), but also to actually delete the `Delivery` record from the table (otherwise we are left with essentially an orphan record, but crucially a record that still contains PII !!)

src/test/kotlin/uk/gov/dluhc/printapi/jobs/InitialRetentionPeriodDataRemovalJobIntegrationTest.kt
```diff
@@ -29,6 +33,10 @@ internal class InitialRetentionPeriodDataRemovalJobIntegrationTest : Integration
         assertThat(certificateRepository.findById(certificate2.id!!).get()).initialRetentionPeriodDataIsRemoved()
         assertThat(certificateRepository.findById(certificate3.id!!).get()).hasInitialRetentionPeriodData()
         assertThat(certificateRepository.findById(certificate4.id!!).get()).hasInitialRetentionPeriodData()
+        assertThat(deliveryRepository.findById(deliveryId1)).isEmpty
+        assertThat(deliveryRepository.findById(deliveryId2)).isEmpty
+        assertThat(deliveryRepository.findById(deliveryId3)).isNotEmpty
+        assertThat(deliveryRepository.findById(deliveryId4)).isNotEmpty
```
matt-wills-valtech, line 39: As an improvement, separate to this, I'll look into adding a Repository class for the `Address` entity within this integration test, so that I can confirm it's been deleted. I've tested it manually for now.

PR #168, opened 2023-02-21
EIP1-3823 Service to generate Anonymous Elector Document (AED) PDF document

Neil-Massey:
Example English AED
image

Example Welsh AED
image

Adds processing of the templates to generate AED documents, includes minor updates to the API spec and Entity model.


src/main/kotlin/uk/gov/dluhc/printapi/config/ElectorDocumentPdfTemplateProperties.kt
```diff
@@ -2,22 +2,40 @@ package uk.gov.dluhc.printapi.config
...
 ) {
-    data class English(
+    data class English<T>(
         val path: String,
-        val placeholder: Placeholder,
+        val placeholder: T,
         val images: Images
     ) {
-        data class Placeholder(
+        data class TemporaryCertificatePlaceholder(
```
adacayi, line 38: Hi @Neil-Massey ,

I need to admit, it took some time for me to understand the new structure, especially because we have `TemporaryCertificatePlaceholder` classes under `English` and `Welsh` classes but not the `AnonymousElectorDocumentPlaceholder`. The reason for it is, I was expecting AED structure would be similar to Temporary Certificate. I can now see that the AED placeholders don't change by the language and that is why it is not duplicated under English and Welsh classes. 

However, I assume it would be much easier and quicker for me to understand if we have it as below:

```
@ConfigurationProperties(prefix = "temporary-certificate.certificate-pdf")
@ConstructorBinding
class TemporaryCertificatePdfTemplateProperties(
    english: English<EnglishTemporaryCertificatePlaceholder>,
    welsh: Welsh<WelshTemporaryCertificatePlaceholder>
) : ElectorDocumentPdfTemplateProperties<EnglishTemporaryCertificatePlaceholder, WelshTemporaryCertificatePlaceholder>(
    english,
    welsh
)

@ConfigurationProperties(prefix = "anonymous-elector-document.certificate-pdf")
@ConstructorBinding
class AnonymousElectorDocumentPdfTemplateProperties(
    english: English<AnonymousElectorDocumentPlaceholder>,
    welsh: Welsh<AnonymousElectorDocumentPlaceholder>
) : ElectorDocumentPdfTemplateProperties<AnonymousElectorDocumentPlaceholder, AnonymousElectorDocumentPlaceholder>(
    english,
    welsh
)

abstract class ElectorDocumentPdfTemplateProperties<EP, WP>(
    val english: English<EP>,
    val welsh: Welsh<WP>
) {
    data class English<T>(
        val path: String,
        val placeholder: T,
        val images: Images
    )

    data class Welsh<T>(
        val path: String,
        val placeholder: T,
        val images: Images
    )

    data class EnglishTemporaryCertificatePlaceholder(
        val electorName: String,
        val localAuthorityNameEn: String,
        val dateOfIssue: String,
        val validOnDate: String,
        val certificateNumber: String
    )

    data class WelshTemporaryCertificatePlaceholder(
        val electorName: String,
        val localAuthorityNameEn: String,
        val localAuthorityNameCy: String,
        val dateOfIssue: String,
        val validOnDate: String,
        val certificateNumber: String
    )

    data class AnonymousElectorDocumentPlaceholder(
        val electoralRollNumber: String,
        val dateOfIssue: String,
        val certificateNumber: String
    )

    data class Images(
        val voterPhoto: PhotoProperties
    )
    
    data class PhotoProperties(
        val pageNumber: Int = 1,
        val absoluteXMm: Float,
        val absoluteYMm: Float,
        val fitWidthMm: Float,
        val fitHeightMm: Float
    )
}

```
We wouldn't be duplicating the image class as well in this approach.
What are your thoughts?

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
```diff
@@ -41,40 +41,40 @@ class AnonymousElectorDocument(
...
+    var sourceReference: String,
 
     @field:Size(max = 255)
-    var applicationReference: String? = null,
+    var applicationReference: String?,
 
     @field:NotNull
     @Enumerated(EnumType.STRING)
-    var certificateLanguage: CertificateLanguage? = null,
+    var certificateLanguage: CertificateLanguage?,
```
adacayi, line 59: Minor. I think we can remove nullable for this field as well.
<pre>
@field:NotNull
@Enumerated(EnumType.STRING)
var certificateLanguage: CertificateLanguage,
</pre>

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
```diff
@@ -97,14 +97,14 @@ class AnonymousElectorDocument(
 ) {
 
     val status: AedPrintRequestStatus.Status?
-        get() = printRequests.sortedByDescending { it.requestDateTime }.first().status
+        get() = printRequests.maxBy { it.requestDateTime }.status
```
adacayi, line 100: Minor, @Neil-Massey. I think we can use the `getLatestPrintRequest` here to make the intent clearer and avoiding duplication of the logic.

<pre>
    val status: AedPrintRequestStatus.Status?
        get() = getLatestPrintRequest().status
</pre>

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
```diff
@@ -97,14 +97,14 @@ class AnonymousElectorDocument(
...
 
     fun addPrintRequest(newPrintRequest: AedPrintRequest): AnonymousElectorDocument {
         printRequests += newPrintRequest
         return this
     }
 
-    private fun getCurrentPrintRequest(): AedPrintRequest = printRequests.sortedByDescending { it.requestDateTime }.first()
+    fun getLatestPrintRequest(): AedPrintRequest = printRequests.maxBy { it.requestDateTime }
```
adacayi, line 107: 👍

src/main/kotlin/uk/gov/dluhc/printapi/config/ElectorDocumentPdfTemplateProperties.kt
Neil-Massey: Looks good - I wanted to revisit after I had a working version committed (I tried several approaches before this generified ones). Will update with your suggestion 👍

src/main/kotlin/uk/gov/dluhc/printapi/dto/GenerateAnonymousElectorDocumentDto.kt
```diff
@@ -0,0 +1,19 @@
...
+
+data class GenerateAnonymousElectorDocumentDto(
+    val gssCode: String,
+    val sourceType: SourceType,
+    val sourceReference: String,
+    val applicationReference: String,
+    val electoralRollNumber: String,
+    val photoLocation: String,
+    val certificateLanguage: CertificateLanguage,
+    val supportingInformationFormat: SupportingInformationFormat,
```
adacayi, line 11: Minor, @Neil-Massey. 

I think it will be good to rename the `supportingInfoFormat` field of the `GenerateAnonymousElectorDocumentRequest` in the open api specs as `supportingInformationFormat` as well so the mapping would be more convenient and it will be consistent with the naming of the other fields for the `supportingInformationFormat`.

src/main/kotlin/uk/gov/dluhc/printapi/mapper/AnonymousElectorDocumentMapper.kt
```diff
@@ -0,0 +1,47 @@
...
+    @Autowired
+    protected lateinit var idFactory: IdFactory
+
+    @Autowired
+    protected lateinit var printRequestMapper: AedPrintRequestMapper
+
+    @Mapping(target = "photoLocationArn", source = "aedRequest.photoLocation")
+    @Mapping(target = "certificateNumber", expression = "java( idFactory.vacNumber() )")
+    @Mapping(target = "printRequests", expression = "java( kotlin.collections.CollectionsKt.mutableListOf() )")
+    @Mapping(target = "contactDetails", expression = "java( toContactDetails( aedRequest ) )")
```
adacayi, line 31: I just checked and this seems to work @Neil-Massey:
`    @Mapping(target = "contactDetails", source = "aedRequest")`

src/main/kotlin/uk/gov/dluhc/printapi/mapper/AnonymousElectorDocumentMapper.kt
```diff
@@ -0,0 +1,47 @@
...
+    abstract fun toAnonymousElectorDocument(
+        aedRequest: GenerateAnonymousElectorDocumentDto,
+        aedTemplateFilename: String
+    ): AnonymousElectorDocument
+
+    abstract fun toContactDetails(aedRequest: GenerateAnonymousElectorDocumentDto): AedContactDetails
```
adacayi, line 37: We can remove this method if we do the below mapping instead:

`    @Mapping(target = "contactDetails", source = "aedRequest")`

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
```diff
@@ -41,40 +41,40 @@ class AnonymousElectorDocument(
...
+    var applicationReference: String?,
 
     @field:NotNull
     @Enumerated(EnumType.STRING)
-    var certificateLanguage: CertificateLanguage? = null,
+    var certificateLanguage: CertificateLanguage?,
 
     @Enumerated(EnumType.STRING)
-    var supportingInformationFormat: SupportingInformationFormat? = null,
+    var supportingInformationFormat: SupportingInformationFormat?,
```
adacayi, line 62: @Neil-Massey, I see that the supporting information format is in the `PrintRequest` for VACs, so assumed it might be different for different print requests. Is it not the case for the `AED`s as well? Can we not have 2 different print requests for AEDs in the future with 2 different supporting information formats? 
I saw there is only one supporting information format for AED, but can that change in the future?

src/main/kotlin/uk/gov/dluhc/printapi/mapper/AnonymousElectorDocumentMapper.kt
Neil-Massey: Thanks 👍

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
```diff
@@ -41,40 +41,40 @@ class AnonymousElectorDocument(
...
 
     @field:NotNull
     @field:Size(max = 255)
-    var photoLocationArn: String? = null,
+    var photoLocationArn: String,
```
adacayi, line 66: Similar comment for as the above one. Seems like for the `VAC`s we have `photoLocationArn` in the `PrintRequest` entity, rather than the `Certificate`, possibly allowing different `photoLocationArn`s for different print requests  of the same application. 

Is that not the case for the `AED`s? 

Can we not have two different `photoLocationArn`s for two different print requests for the same `AED` application?

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
Neil-Massey: Yeah - I think you are right, have updated the schema too.

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
Neil-Massey: Will need to double check with the BAs - they only described reprint flows for LSD and electoral roll number re-indexing.  Will address in next PR if any changes required.

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
Neil-Massey: As above.

adacayi (approved):
LGTM, @Neil-Massey ,
Thanks, for addressing the comments.

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AedContactDetails.kt
```diff
@@ -35,14 +35,14 @@ class AedContactDetails(
...
-    var firstName: String? = null,
+    var firstName: String,
 
     @field:Size(max = 255)
     var middleNames: String? = null,
 
     @field:NotNull
     @field:Size(max = 255)
-    var surname: String? = null,
+    var surname: String,
```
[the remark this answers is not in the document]
Neil-Massey: Yes - I envisioned the reference would be nulled, in similar way to `Delivery` for a VAC certificate. 👍

PR #169, opened 2023-02-21
EIP1-3765 Remove duplicate print requests from count

Neil-Massey:
Found issue with the number of requests being included in the SQS message being more than it should as double counting print requests.


NathanRussellValtech (approved):
LGTM 👍

PR #171, opened 2023-02-24
EIP1-3229 Add endpoint to return a summary of the Anonymous Elector Documents (AEDs)

Neil-Massey:
image

### DIFF
--- a/src/main/resources/openapi/PrintAPIs.yaml
+++ b/src/main/resources/openapi/PrintAPIs.yaml
@@ -458,7 +458,7 @@ paths:
         in: path
         required: true
       - name: applicationId
-        description: The identifier of the Anonymous Elector Document.
+        description: The identifier of the Anonymous Voter Card Application.
         schema:
           type: string
         in: path
@@ -501,8 +501,8 @@ paths:
               application/json: |
                 {}
     get:
-      summary: Returns the Anonymous Elector Document summary for an application
-      description: Returns the Anonymous Elector Document summary for an application
+      summary: Returns the Anonymous Elector Document summary for an anonymous application
+      description: Returns the Anonymous Elector Document summary for an anonymous application
       tags:
         - Anonymous Elector Documents (AED)
       responses:
@@ -510,7 +510,7 @@ paths:
           $ref: '#/components/responses/AnonymousElectorDocumentSummaries'
         '404':
           description: Not Found
-      operationId: get-anonymous-elector-document-summary-by-application-id
+      operationId: get-anonymous-elector-document-summaries-by-application-id
       security:
         - eroUserCognitoUserPoolAuthorizer: [ ]
       x-amazon-apigateway-integration:
@@ -912,11 +912,11 @@ components:
           type: string
           format: email
           example: fred.blogs@some-domain.co.uk
-          description: UserId of the user who made the print request update
+          description: UserId of the user who made the update to the AED
         dateTime:
           type: string
           format: date-time
-          description: The timestamp in ISO-8601 UTC when the print request status update was received
+          description: The timestamp in ISO-8601 UTC when the status was updated on the AED
           example: '2022-11-17T10:53:25Z'
         electoralRollNumber:
           type: string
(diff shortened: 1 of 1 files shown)


src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -446,6 +446,87 @@ paths:
...
+  '/eros/{eroId}/anonymous-elector-documents/applications/{applicationId}':
+    parameters:
+      - name: eroId
+        description: The ID of the Electoral Registration Office responsible for the application.
+        schema:
+          type: string
+        in: path
+        required: true
+      - name: applicationId
+        description: The identifier of the Anonymous Elector Document.
```
matt-wills-valtech, line 461: Is that right, or is it the ID of the AED's originating application (aka `sourceReference`)?

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -446,6 +446,87 @@ paths:
...
+        '200':
+          $ref: '#/components/responses/AnonymousElectorDocumentSummaries'
+        '404':
+          description: Not Found
+      operationId: get-anonymous-elector-document-summary-by-application-id
+      security:
+        - eroUserCognitoUserPoolAuthorizer: [ ]
+      x-amazon-apigateway-integration:
+        type: HTTP_PROXY
+        uri: '${base_uri}/eros/{eroId}/anonymous-elector-documents/applications/{applicationId}'
```
matt-wills-valtech, line 518: To contradict myself slightly here, I think `/applications` is ok here as it's clearly in the path/context of `/anonymous-elector-documents`.

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -803,6 +884,63 @@ components:
         - photoLocation
         - validOnDate
 
+    AnonymousElectorDocumentSummariesResponse:
+      title: AnonymousElectorDocumentSummariesResponse
+      type: object
+      description: A response containing a collection of Anonymous Elector Document summaries
```
matt-wills-valtech, line 890: Will the most recent one be returned first?

src/main/resources/openapi/PrintAPIs.yaml
Neil-Massey: You're right, it is the sourceReference, and will be found using `findByGssCodeInAndSourceTypeAndSourceReference`.  

Agree, wording is wrong, the Certificates & Temporary Certificates use the following wording:
```
The identifier of the Voter Card Application.
```

So maybe, this one should be:
```
The identifier of the Anonymous Voter Card Application.
```

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -803,6 +884,63 @@ components:
...
+          $ref: '#/components/schemas/AnonymousElectorDocumentStatus'
+        userId:
+          type: string
+          format: email
+          example: fred.blogs@some-domain.co.uk
+          description: UserId of the user who made the print request update
+        dateTime:
+          type: string
+          format: date-time
+          description: The timestamp in ISO-8601 UTC when the print request status update was received
```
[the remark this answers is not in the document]
Neil-Massey: Should be the timestamp corresponding to when the status changed.

src/main/resources/openapi/PrintAPIs.yaml
Neil-Massey: Let's say we will do that - description updated 👍

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -803,6 +884,63 @@ components:
         - photoLocation
         - validOnDate
 
+    AnonymousElectorDocumentSummariesResponse:
+      title: AnonymousElectorDocumentSummariesResponse
+      type: object
+      description: A response containing a collection of Anonymous Elector Document summaries sorted with latest changes ordered first.
```
matt-wills-valtech, line 890: 👍

src/main/resources/openapi/PrintAPIs.yaml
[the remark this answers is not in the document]
Neil-Massey: To clarify, the endpoint returns only 1 summary record for each AED and that has its latest status and the timestamp of when that status was assigned.

matt-wills-valtech (approved):
LGTM, thanks for finishing this off 👍

PR #172, opened 2023-02-27
EIP1-4179 Refactor Anonymous Elector Document data model

Neil-Massey:
Refactor Anonymous Elector Document data model to not have print requests - each new AED request creates a new AED document.


NathanRussellValtech (approved):
LGTM, but make sure @matt-wills-valtech is also happy before merging (he understands the requirements better than me) 👍

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
```diff
@@ -73,9 +73,27 @@ class AnonymousElectorDocument(
...
+
+    @field:NotNull
+    @field:Size(max = 255)
+    var aedTemplateFilename: String,
+
+    @field:NotNull
+    var issueDate: LocalDate,
+
+    @field:NotNull
+    var requestDateTime: Instant,
```
matt-wills-valtech, line 88: Is this the same as `dateCreated`? If so, do we need it?

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocumentStatus.kt
```diff
@@ -25,7 +25,7 @@ import javax.validation.constraints.Size
 @Table
 @Entity
 @EntityListeners(AuditingEntityListener::class)
-class AedPrintRequestStatus(
+class AnonymousElectorDocumentStatus(
```
matt-wills-valtech, line 28: 👍

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
```diff
@@ -97,15 +115,15 @@ class AnonymousElectorDocument(
...
-    val status: AedPrintRequestStatus.Status?
-        get() = getLatestPrintRequest().status
+    val status: AnonymousElectorDocumentStatus.Status?
+        get() = getLatestStatus().status
 
-    fun addPrintRequest(newPrintRequest: AedPrintRequest): AnonymousElectorDocument {
-        printRequests += newPrintRequest
+    fun addStatus(newStatus: AnonymousElectorDocumentStatus): AnonymousElectorDocument {
+        statusHistory += newStatus
         return this
```
matt-wills-valtech, line 123: Out of interest, why does it return `this`?

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
```diff
@@ -97,15 +115,15 @@ class AnonymousElectorDocument(
     var version: Long? = null
 ) {
 
-    val status: AedPrintRequestStatus.Status?
-        get() = getLatestPrintRequest().status
+    val status: AnonymousElectorDocumentStatus.Status?
```
matt-wills-valtech, line 118: Is this field used anywhere?

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
Neil-Massey: Following same pattern as in `TemporaryCertificate` and `Certificate` - useful when using some of the Kotlin scope functions.

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
Neil-Massey: This would be set during mapping, instead of waiting for the JPA to populate it whilst persisting - this means it can be set at construction.  Also, it records when request is made, whereby the `created` is for audit trail.  It's similar to way we have both `userId` and `createdBy` properties.  Lastly, this is only moving from one table to another - so hasn't been altered by the refactor.

matt-wills-valtech: Thanks for the explanation. IMHO it feels a bit unnecessary, but I'm not going to worry about it 👍

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
Neil-Massey: Removed (can be added if/when needed)

src/main/resources/db/changelog/ddl/0015_refactor_anonymous_elector_document_tables.xml
```diff
@@ -0,0 +1,190 @@
...
+                <column name="date_created" type="timestamp">
+                    <constraints nullable="false"/>
+                </column>
+                <column name="created_by" type="varchar(255)">
+                    <constraints nullable="false"/>
+                </column>
+                <column name="version" type="bigint">
+                    <constraints nullable="false"/>
+                </column>
+            </createTable>
```
matt-wills-valtech, line 181: Not sure how we'd recover the previous data, but that's currently not a concern in prod 👍

matt-wills-valtech (approved):
LGTM, thanks for accommodating the change in requirements.

src/main/resources/db/changelog/ddl/0015_refactor_anonymous_elector_document_tables.xml
Neil-Massey: Yeah, good point - I guess it would still be in the new `anonymous_elector_document_status` table.  Rollbacks where data is concerned is likely to need manual intervention - luckily we don't have any data yet 👍

PR #174, opened 2023-02-27
EIP1-4154 - Retention period columns

matt-wills-valtech:
This PR adds `finalRetentionRemovalDate` to the `Certificate` entity, so that we can set the date when any remaining certificate data should be removed.

Whilst in the area, I spotted that `AnonymousElectorDocument` was missing `initialRetentionDataRemoved`, which I had to add to `Certificate` retrospectively (presumably after @Neil-Massey initially created the AED entity), so I've included that here. It then seemed unreasonable not to add `finalRetentionRemovalDate` as well, even though both these columns are strictly part of [EIP1-3753](https://technologyprogramme.atlassian.net/browse/EIP1-3753) and [EIP1-3567](https://technologyprogramme.atlassian.net/browse/EIP1-3567) respectively.

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
@@ -73,11 +73,12 @@ class Certificate(
     var suggestedExpiryDate: LocalDate = issueDate.plusYears(10),
 
     /**
-     * The legislation stipulates there are two retention periods for certificate related data. The first (initial)
-     * period applies to PII data that is not on the printed certificate itself (e.g. the addressee/address on the
-     * envelope), which needs to be removed 28 (configurable) working days after the certificate is "issued".
-     * For standard (non-temporary) certificates, the retention period is considerably longer and is currently specified
-     * as the tenth 1st July.
+     * The legislation stipulates there are three retention periods for certificate related data. The first (initial)
+     * period (which this field relates to), applies to PII data that is not on the printed certificate itself (e.g. the
+     * addressee/address on the envelope). This needs to be removed 28 (configurable) working days after the
+     * certificate is "issued".
+     * The second retention period applies to Temporary Certificates (see [TemporaryCertificate]) and the third (final)
+     * retention period is handled by `finalRetentionRemovalDate` below.
      */
     var initialRetentionRemovalDate: LocalDate? = null,
 
@@ -85,6 +86,12 @@ class Certificate(
      * Set to true after the initial retention period data is removed.
      */
     var initialRetentionDataRemoved: Boolean = false,
+
+    /**
+     * The date that all remaining certificate data should be removed after the third (final) retention period. This is
+     * currently specified as the tenth 1st July in the legislation.
+     */
+    var finalRetentionRemovalDate: LocalDate? = null,
 
     /**
      * Certificate status corresponds to the current status of the most recent
--- a/src/main/resources/db/changelog/ddl/0016_final_retention_data_removed_column.xml
+++ b/src/main/resources/db/changelog/ddl/0016_final_retention_data_removed_column.xml
@@ -1,0 +1,80 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+
+    <changeSet author="matt.wills@valtech.com" id="0016_EIP1-4154_alter_certificate - add final_retention_removal_date column" context="ddl">
+
+        <addColumn tableName="certificate">
+            <column name="final_retention_removal_date" type="date" afterColumn="initial_retention_data_removed"
+                    remarks="The date any remaining certificate data should be removed after the final retention period">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+
+        <createIndex tableName="certificate" indexName="certificate_final_retention_removal_date_idx">
+            <column name="source_type"/>
+            <column name="final_retention_removal_date"/>
+        </createIndex>
+
+        <rollback>
+            <dropColumn
+                tableName="certificate"
+                columnName="final_retention_removal_date"/>
+        </rollback>
+    </changeSet>
+
+    <changeSet author="matt.wills@valtech.com" id="0016_EIP1-4154_alter_aed - add data retention columns" context="ddl">
+
+        <addColumn tableName="anonymous_elector_document">
+            <column name="initial_retention_data_removed" type="boolean" valueBoolean="false" afterColumn="initial_retention_removal_date"
+                    remarks="Indicates if initial retention period data has been removed from an AED">
+                <constraints nullable="false"/>
+            </column>
+        </addColumn>
+
+        <addColumn tableName="anonymous_elector_document">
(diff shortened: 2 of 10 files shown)


src/test/kotlin/uk/gov/dluhc/printapi/database/repository/AnonymousElectorDocumentRepositoryIntegrationTest.kt
```diff
@@ -0,0 +1,70 @@
...
+import org.junit.jupiter.api.Nested
+import org.junit.jupiter.api.Test
+import uk.gov.dluhc.printapi.config.IntegrationTest
+import uk.gov.dluhc.printapi.database.entity.SourceType.VOTER_CARD
+import uk.gov.dluhc.printapi.database.repository.AnonymousElectorDocumentRepositoryExtensions.findPendingRemovalOfFinalRetentionData
+import uk.gov.dluhc.printapi.database.repository.AnonymousElectorDocumentRepositoryExtensions.findPendingRemovalOfInitialRetentionData
+import uk.gov.dluhc.printapi.testsupport.testdata.entity.buildAnonymousElectorDocument
+import java.time.LocalDate
+
+internal class AnonymousElectorDocumentRepositoryIntegrationTest : IntegrationTest() {
```
matt-wills-valtech, line 13: I'm not a massive fan of having repository tests when we can test this functionality higher up the stack, so I may consider removing this in a future PR.

Neil-Massey (approved):
LGTM

src/main/resources/db/changelog/ddl/0016_final_retention_data_removed_column.xml
```diff
@@ -0,0 +1,80 @@
...
+    <changeSet author="matt.wills@valtech.com" id="0016_EIP1-4154_alter_aed - add data retention columns" context="ddl">
+
+        <addColumn tableName="anonymous_elector_document">
+            <column name="initial_retention_data_removed" type="boolean" valueBoolean="false" afterColumn="initial_retention_removal_date"
+                    remarks="Indicates if initial retention period data has been removed from an AED">
+                <constraints nullable="false"/>
+            </column>
+        </addColumn>
+
+        <addColumn tableName="anonymous_elector_document">
```
sentioservices, line 38: fyi it looks like addColumn supports multiple column elements

sentioservices (commented):
LGTM, just one minor comment

src/main/resources/db/changelog/ddl/0016_final_retention_data_removed_column.xml
```diff
@@ -0,0 +1,80 @@
...
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+
+        <createIndex tableName="certificate" indexName="certificate_final_retention_removal_date_idx">
+            <column name="source_type"/>
+            <column name="final_retention_removal_date"/>
+        </createIndex>
+
+        <rollback>
```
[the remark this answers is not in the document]
matt-wills-valtech: I was hoping that removing `final_retention_removal_date` would automatically remove the index, but does it just remove the column from the index? I'll do some reading!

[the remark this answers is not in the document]
matt-wills-valtech: > If columns are dropped from a table, the columns are also removed from any index of which they are a part. If all columns that make up an index are dropped, the index is dropped as well. 

Oops, I'll drop the index :)

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
```diff
@@ -86,6 +87,12 @@ class Certificate(
      */
     var initialRetentionDataRemoved: Boolean = false,
 
+    /**
+     * The date that all remaining certificate data should be removed after the third (final) retention period. This is
+     * currently specified as the tenth 1st July in the legislation.
```
vishalgupta-valtech, line 92: `as the tenth 1st July ` ? (Similar comment in file above). Is that a UK legislation called `tenth 1st July`?

vishalgupta-valtech (approved):
lgtmw

src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Certificate.kt
matt-wills-valtech: It's ten (10 x) 1st Julys from now 🤯

src/main/resources/db/changelog/ddl/0016_final_retention_data_removed_column.xml
```diff
@@ -0,0 +1,93 @@
...
+                <column name="initial_retention_removal_date"/>
+            </createIndex>
+        </rollback>
+    </changeSet>
+
+    <changeSet author="matt.wills@valtech.com" id="0016_EIP1-4154_alter_aed elector_roll_number width" context="ddl">
+        <modifyDataType
+            columnName="electoral_roll_number"
+            newDataType="varchar(30)"
+            tableName="anonymous_elector_document"/>
```
matt-wills-valtech, line 83: @Neil-Massey  - I've made this change here as we know we need it and I thought I'd prevent the need for another liquibase script.

I'm not sure where else it needs changing, other than in the swagger spec (`GenerateAnonymousElectorDocumentRequest`), but I'd rather do that separately.

Neil-Massey: The only other change is to the OpenAPI regex for the property definition.

src/main/resources/db/changelog/ddl/0016_final_retention_data_removed_column.xml
```diff
@@ -0,0 +1,93 @@
...
+                <column name="source_type"/>
+                <column name="initial_retention_removal_date"/>
+            </createIndex>
+        </rollback>
+    </changeSet>
+
+    <changeSet author="matt.wills@valtech.com" id="0016_EIP1-4154_alter_aed elector_roll_number width" context="ddl">
+        <modifyDataType
+            columnName="electoral_roll_number"
+            newDataType="varchar(30)"
```
sentioservices, line 82: This column was previously NOT NULL. I think you can set it using:
```
newDataType="varchar(30) NOT NULL"
```

src/main/resources/db/changelog/ddl/0016_final_retention_data_removed_column.xml
```diff
@@ -0,0 +1,93 @@
...
+            tableName="anonymous_elector_document"/>
+
+        <rollback>
+            <modifyDataType
+                columnName="electoral_roll_number"
+                newDataType="varchar(7)"
```
sentioservices, line 88: Similarly `NOT NULL` should be applied during rollback.

src/main/resources/db/changelog/ddl/0016_final_retention_data_removed_column.xml
sentioservices: We could also change uk.gov.dluhc.printapi.testsupport.testdata.RandomUtilsKt#aValidElectoralRollNumber
It currently returns a value of length 6

matt-wills-valtech: I was in two minds about changing the DB column as it's obviously not part of this JIRA, but made sense whilst I'm "in the area". I think we should change the API spec (including the regex) separately 👍

src/main/resources/db/changelog/ddl/0016_final_retention_data_removed_column.xml
matt-wills-valtech: Ah yes. Fixed 👍

sentioservices (approved):
LGTM, thanks for addressing comments

PR #177, opened 2023-03-02
EIP1-4155 - Scheduled task to remove remaining Certificate data

matt-wills-valtech:


src/main/resources/application.yml
```diff
@@ -66,6 +66,11 @@ jobs:
     # Runs at 21:35, 01:35 and 05:35.
     # Running outside business hours to avoid delivery info suddenly disappearing.
     cron: "0 35 1,5,21 * * *"
+  remove-vca-final-retention-period-data:
+    name: "RemoveVcaFinalRetentionPeriodData"
+    # Runs at 21:45, 01:45 and 05:45.
+    # Running outside business hours to avoid certificate data suddenly disappearing.
+    cron: "0 45 1,5,21 * * *"
```
sentioservices, line 73: Should this only run once a year on the first of July?

src/main/resources/application-dev.yml
```diff
@@ -9,3 +9,6 @@ jobs:
   remove-vca-initial-retention-period-data:
     name: "RemoveVcaInitialRetentionPeriodData"
     cron: "0 */10 * ? * *" # every 10 minutes
+  remove-vca-final-retention-period-data:
+    name: "RemoveVcaFinalRetentionPeriodData"
+    cron: "0 */10 * ? * *" # every 10 minutes
```
sentioservices, line 14: What about this cron expression "0 5/10 7-19 ? * MON-FRI" to try and avoid log output from other jobs by starting at 5 past the hour.
`At second :00, every 10 minutes starting at minute :05, every hour between 07am and 19pm, every day between Monday and Friday, every month `

src/main/resources/application-int.yml
```diff
@@ -2,3 +2,6 @@ jobs:
   remove-vca-initial-retention-period-data:
     name: "RemoveVcaInitialRetentionPeriodData"
     cron: "0 */10 * ? * *" # every 10 minutes
+  remove-vca-final-retention-period-data:
+    name: "RemoveVcaFinalRetentionPeriodData"
+    cron: "0 */10 * ? * *" # every 10 minutes
```
sentioservices, line 7: similar to comment on application-dev.yml

src/test/kotlin/uk/gov/dluhc/printapi/jobs/FinalRetentionPeriodDataRemovalJobIntegrationTest.kt
```diff
@@ -0,0 +1,49 @@
...
+import uk.gov.dluhc.printapi.database.entity.PrintRequest
+import uk.gov.dluhc.printapi.testsupport.TestLogAppender
+import uk.gov.dluhc.printapi.testsupport.testdata.entity.buildCertificate
+import java.time.LocalDate
+import java.util.UUID
+
+internal class FinalRetentionPeriodDataRemovalJobIntegrationTest : IntegrationTest() {
+
+    @Autowired
+    private lateinit var testPrintRequestRepository: TestPrintRequestRepository
```
sentioservices, line 19: I'll like the way you define and autowire the TestPrintRequestRepository object but doesn't defining Autowired attributes outside IntegrationTest class result in a new spring context being created?

sentioservices (approved):
LGTM, just some minors comments around cron timings

src/main/resources/application.yml
matt-wills-valtech: That would make more sense, or at least to run it far less often 😄 

However, I was wondering if we want to tie the job's scheduled to the current legislation so tightly? Or if there was ever a use case to set the date manually. Personally I think once a day is ok as there will be so little overhead, but granted not 3 times.

src/test/kotlin/uk/gov/dluhc/printapi/jobs/FinalRetentionPeriodDataRemovalJobIntegrationTest.kt
matt-wills-valtech: Oh I hadn't considered that. I assumed it would create and wire in the bean in the current context. Why would it create a new one?

sentioservices: I think currently it uses the same context for all of the int-tests. If it added a bean into the context for this test it would need to remove it for all the rest, so I think it creates a new spring-context where the context differs between tests.

sentioservices (approved):
LGTM, thanks for addressing concerns.

PR #182, opened 2023-03-06
EIP1-4186: EIP1:4352: API spec for updating AED status

vishalgupta-valtech:
image

**Updated to use `certificateNumber`** 
image

### DIFF
--- a/src/main/resources/openapi/PrintAPIs.yaml
+++ b/src/main/resources/openapi/PrintAPIs.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Print APIs
-  version: '1.8.1'
+  version: '1.9.1'
   description: Print APIs
   contact:
     name: Krister Bone
@@ -614,7 +614,7 @@ paths:
       description: |
         Enable CORS by returning correct headers
       tags:
-        - Anonymous Elector Documents (AED)
+        - Anonymous Elector Documents (AED) Explainer Document
       responses:
         200:
           description: Default response for CORS method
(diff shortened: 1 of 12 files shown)


src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -682,11 +682,115 @@ paths:
...
+  # --------------------------------------------------------------------------------
+  '/eros/{eroId}/anonymous-elector-documents/{documentNumber}/status':
+    parameters:
+      - name: eroId
+        description: The ID of the Electoral Registration Office responsible for an AED application.
+        schema:
+          type: string
+        in: path
+        required: true
+      - name: documentNumber
```
Neil-Massey, line 696: Is this the same as the `certificate_number` in the database?

sentioservices (approved):
LGTM

matt-wills-valtech (changes requested):
I don't think we should make a partial change to switch to `documentNumber` before we're received confirmation.

In the meantime, we could opt for `certificateNumber` and change it everywhere in the print-api if the need arises (not least because it may take a while to get a decision!), but let's decide tomorrow 👍

vishalgupta-valtech:
> Robin has confirmed we should stick with certificate number:
> 
> image



> Robin has confirmed we should stick with certificate number:
> 
> image

Thank you @matt-wills-valtech . I have updated the PR and now using a certificateNumber in endpoint url 👍🏼 
image

matt-wills-valtech (approved):
LGTM

vishalgupta-valtech:
Follow up PR that corrects a little typo https://github.com/cabinetoffice/eip-ero-print-api/pull/183

PR #184, opened 2023-03-08
EIP1-4307 Remove remaining certificate data and photo

matt-wills-valtech:
This PR removes the remaining certificate data from the database (after the removal of the "initial" retention period data), plus the certificate's photo from S3.

This data is removed on the 1st July each year and, due to the expected high volumes, we are retrieving the essential details of each certificate in batches of 10,000 at a time and placing each one on a queue. The certificate data is then removed from both S3 and the database in a separate transaction for each one.


src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateDataRetentionService.kt
```diff
@@ -62,6 +64,8 @@ class CertificateDataRetentionService(
         logger.info { "Finding certificates with sourceType $sourceType to remove final retention period data from" }
         with(certificateRepository.findPendingRemovalOfFinalRetentionData(sourceType = sourceType)) {
             forEach {
+                // TODO - EIP1-4481 - get s3 arn from certificate instead of print request
+                s3CertificatePhotoService.removePhoto(it.printRequests[0].photoLocationArn!!)
                 certificateRepository.delete(it)
```
matt-wills-valtech, line 69: I'd like to discuss the efficiency of iterating over each certificate and deleting from both S3 and the database each time.

src/main/kotlin/uk/gov/dluhc/printapi/service/S3PhotoService.kt
```diff
@@ -0,0 +1,28 @@
...
+     * Removes the photo that is on the printed "Elector Document" (i.e. Certificate/AED) from S3.
+     */
+    fun removePhoto(photoS3Arn: String) {
+        try {
+            with(parseS3Arn(photoS3Arn)) {
+                s3Client.deleteObject(DeleteObjectRequest.builder().bucket(bucket).key(path).build())
+                logger.info { "Deleted photo with S3 arn [$photoS3Arn]" }
+            }
+        } catch (e: SdkException) {
+            logger.warn { "Unable to delete photo with S3 arn [$photoS3Arn] due to error [${e.cause?.message ?: e.cause}]" }
```
matt-wills-valtech, line 25: I'm going to change this as there are probably cases where we want to throw an exception and rollback the transaction in the `CertificateDataRetentionService` (for example when there's a server error).

I thought it makes sense to be forgiving in case the photo cannot be found (although there's no mention of an exception being thrown in this scenario in the `S3Client` javadocs). I'll do some more investigation.

If the photo cannot be found, then it makes sense to continue to remove the data from the database, but otherwise we should probably roll back the transaction. However, this would mean the transaction is rolled back for all certificates in the list, which may not be right. Hence why I'd like to give more thought to the transaction boundaries. Suggestions welcomed 😄

src/main/kotlin/uk/gov/dluhc/printapi/service/CertificateDataRetentionService.kt
sentioservices: The way we coped with the transaction boundary within VCA was to use ApplicationRemovalTransactionalService. I think a similar approach here would work to prevent rolling back the whole list. 
Also, in this scenario because we're not using SQS to independently process each certificate, we might need a try/catch(Exception) inside the for each block so that if we do get blocked deleting one certificate we can continue to process the rest.

sentioservices (commented):
Just a couple of initial thoughts...

src/test/kotlin/uk/gov/dluhc/printapi/messaging/RemoveCertificateMessageListenerTest.kt
```diff
@@ -0,0 +1,67 @@
...
+        sqsMessagingTemplate.convertAndSend(removeCertificateQueueName, payload)
+
+        // Then
+        await.atMost(5, TimeUnit.SECONDS).untilAsserted {
+            assertThat(certificateRepository.findById(certificateId)).isEmpty
+            assertThat(certificatePhotoExists(s3PhotoBucket, s3PhotoPath)).isFalse
+        }
+    }
+
+    private fun addCertificatePhotoToS3(bucket: String, path: String) {
```
matt-wills-valtech, line 46: I plan to put this into a common S3 test support class.

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/CertificateRepository.kt
```diff
@@ -59,10 +63,30 @@ object CertificateRepositoryExtensions {
...
-    fun CertificateRepository.findPendingRemovalOfFinalRetentionData(sourceType: SourceType): List<Certificate> {
+    fun CertificateRepository.findPendingRemovalOfFinalRetentionData(
+        sourceType: SourceType,
+        batchIndex: Int,
+        batchSize: Int = 10000
+    ): Page<CertificateRemovalSummary> {
+        val batchRequest = PageRequest.of(
+            batchIndex,
+            batchSize,
+            Sort.by(Sort.Direction.ASC, "issueDate")
```
sentioservices, line 74: shouldn't the sort by field be `final_retention_removal_date` so that it can use the index.
if you added the new s3arn column that Vishal is adding to the index certificate_final_retention_removal_idx then the query should never even load the table's data making it pretty fast.

matt-wills-valtech: The `final_retention_removal_date` will be the same for each entry, so using `issueDate` is the better option.

Good point about re index. @vishalgupta-valtech - would you mind modifying it in your liquibase script? 😄

vishalgupta-valtech: Index created to include `id` 
image

matt-wills-valtech: Thank you 🙌  I've changed the `Sort.by` column in the repository 👍

src/test/kotlin/uk/gov/dluhc/printapi/messaging/RemoveCertificateMessageListenerTest.kt
matt-wills-valtech: Moved into an extension function.

src/test/kotlin/uk/gov/dluhc/printapi/jobs/FinalRetentionPeriodDataRemovalJobIntegrationTest.kt
```diff
@@ -2,46 +2,70 @@ package uk.gov.dluhc.printapi.jobs
...
-        assertThat(testPrintRequestRepository.findById(certificate3.printRequests[0].id!!)).isNotEmpty
-        assertThat(testPrintRequestRepository.findById(certificate4.printRequests[0].id!!)).isNotEmpty
-        assertThat(TestLogAppender.hasLog("Removed remaining data after final retention period from certificate with sourceReference ${certificate1.sourceReference}", Level.INFO)).isTrue
-        assertThat(TestLogAppender.hasLog("Removed remaining data after final retention period from certificate with sourceReference ${certificate2.sourceReference}", Level.INFO)).isTrue
+        await.atMost(5, TimeUnit.SECONDS).untilAsserted {
+            assertThat(certificateRepository.findById(certificate1.id!!)).isEmpty
+            assertThat(certificateRepository.findById(certificate2.id!!)).isEmpty
+            assertThat(certificateRepository.findById(certificate3.id!!)).isNotEmpty
+            assertThat(certificateRepository.findById(certificate4.id!!)).isNotEmpty
+            assertThat(testPrintRequestRepository.findById(certificate1.printRequests[0].id!!)).isEmpty
```
vishalgupta-valtech, line 61: You could use `CertificateAssert` class if required. It has `hasNoPrintRequests` function.

src/main/kotlin/uk/gov/dluhc/printapi/messaging/RemoveCertificateListener.kt
```diff
@@ -0,0 +1,22 @@
...
+ * Implementation of [MessageListener] to handle [RemoveCertificateMessage] messages
+ */
+@Component
+class RemoveCertificateListener(
+    val certificateDataRetentionService: CertificateDataRetentionService
+) : MessageListener<RemoveCertificateMessage> {
+
+    @SqsListener("\${sqs.remove-certificate-queue-name}")
+    override fun handleMessage(@Valid @Payload payload: RemoveCertificateMessage) {
+        certificateDataRetentionService.removeFinalRetentionPeriodData(payload)
```
vishalgupta-valtech, line 20: It will be nicer to add a logger info stmt

src/main/kotlin/uk/gov/dluhc/printapi/service/S3PhotoService.kt
```diff
@@ -0,0 +1,29 @@
...
+private val logger = KotlinLogging.logger {}
+
+@Service
+class S3PhotoService(
+    private val s3Client: S3Client
+) {
+    /**
+     * Removes the photo that is on the printed "Elector Document" (i.e. Certificate/AED) from S3.
+     */
+    fun removePhoto(photoS3Arn: String) {
```
vishalgupta-valtech, line 18: Could this be generic method `removeObjectFromS3(s3Arn: String)` - Not a blocker

src/test/kotlin/uk/gov/dluhc/printapi/jobs/FinalRetentionPeriodDataRemovalJobIntegrationTest.kt
matt-wills-valtech: I'm making sure the print_request has been removed from the database entirely, rather than just asserting the `PrintRequest` has been removed from the `Certificate` entity. For this reason, I'm using `TestPrintRequestRepository` which I added in a previous PR (I didn't want to add a `PrintRequestRepository` in the main scope).

src/main/kotlin/uk/gov/dluhc/printapi/messaging/RemoveCertificateListener.kt
matt-wills-valtech: I considered adding a debug one. Trouble is there could be over 1 million rows on 1st July each year, so not sure it would be that useful :)

src/main/kotlin/uk/gov/dluhc/printapi/service/S3PhotoService.kt
matt-wills-valtech: I considered that but thought we'll only ever get/remove photos within S3 in the print-api. By making it generic I think you lose some of the readability.

src/main/kotlin/uk/gov/dluhc/printapi/jobs/FinalRetentionPeriodDataRemovalJob.kt
```diff
@@ -14,6 +14,6 @@ class FinalRetentionPeriodDataRemovalJob(
     @Scheduled(cron = "\${jobs.remove-vca-final-retention-period-data.cron}")
     @SchedulerLock(name = "\${jobs.remove-vca-final-retention-period-data.name}")
     fun removeVoterCardFinalRetentionPeriodData() {
```
vishalgupta-valtech, line 16: Is it ok to have voterCard in the function name `removeVoterCardFinalRetentionPeriodData` ?

src/main/resources/application.yml
```diff
@@ -24,20 +24,23 @@ sqs:
   process-print-response-file-queue-name: ${SQS_PROCESS_PRINT_RESPONSE_FILE_QUEUE_NAME}
   process-print-response-queue-name: ${SQS_PROCESS_PRINT_RESPONSE_QUEUE_NAME}
   application-removed-queue-name: ${SQS_APPLICATION_REMOVED_QUEUE_NAME}
+  remove-certificate-queue-name: ${SQS_REMOVE_CERTIFICATE_QUEUE_NAME}
```
vishalgupta-valtech, line 27: Kindly update README with this new var `SQS_REMOVE_CERTIFICATE_QUEUE_NAME`

src/main/resources/openapi/sqs/print-api-sqs-messaging.yaml
```diff
@@ -165,6 +176,21 @@ components:
         - sourceType
         - gssCode
 
+    RemoveCertificateMessage:
+      description: The SQS message sent to remove an individual Certificate record from the database and its corresponding photo from S3.
```
vishalgupta-valtech, line 180: Minor: If you add `title: RemoveCertificateMessage` then swagger ui will show it's name

vishalgupta-valtech (approved):
LGTM, few minor points. 🚀

src/main/kotlin/uk/gov/dluhc/printapi/jobs/FinalRetentionPeriodDataRemovalJob.kt
matt-wills-valtech: Yes because I'm planning on adding separate jobs to remove temp certs and AEDs. This is so that their timings can vary if necessary, but also because they live in separate database tables, so it will just make it easier.

sentioservices (approved):
LGTM. The tests make quite a complex problem much easier to understand 👍

vishalgupta-valtech:
> LGTM. The tests make quite a complex problem much easier to understand 👍

I couldn't agree more. My hawk eyes couldn't find much. 👍🏼

vishalgupta-valtech:
- [ ] Checklist - Infra PR merged https://github.com/cabinetoffice/eip-ero-infra/pull/596 before merging this PR

PR #187, opened 2023-03-10
EIP1-3229: EIP1:4572: Updates GET API spec for Getting AEDs with applicationId as queryParam

vishalgupta-valtech:
image

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/rest/AnonymousElectorDocumentController.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/rest/AnonymousElectorDocumentController.kt
@@ -4,8 +4,8 @@ import org.springframework.beans.factory.annotation.Qualifier
 import org.springframework.core.io.InputStreamResource
 import org.springframework.http.HttpHeaders
 import org.springframework.http.HttpHeaders.CONTENT_DISPOSITION
-import org.springframework.http.HttpStatus
 import org.springframework.http.HttpStatus.CREATED
+import org.springframework.http.HttpStatus.OK
 import org.springframework.http.MediaType.APPLICATION_PDF
 import org.springframework.http.MediaType.APPLICATION_PDF_VALUE
 import org.springframework.http.ResponseEntity
@@ -59,7 +59,7 @@ class AnonymousElectorDocumentController(
 
     @GetMapping
     @PreAuthorize(HAS_ERO_VC_ANONYMOUS_ADMIN_AUTHORITY)
-    @ResponseStatus(HttpStatus.OK)
+    @ResponseStatus(OK)
     fun getAnonymousElectorDocumentSummaries(
         @PathVariable eroId: String,
         @RequestParam applicationId: String,
--- a/src/main/kotlin/uk/gov/dluhc/printapi/rest/AnonymousElectorDocumentController.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/rest/AnonymousElectorDocumentController.kt
@@ -4,6 +4,7 @@ import org.springframework.beans.factory.annotation.Qualifier
 import org.springframework.core.io.InputStreamResource
 import org.springframework.http.HttpHeaders
 import org.springframework.http.HttpHeaders.CONTENT_DISPOSITION
+import org.springframework.http.HttpStatus
 import org.springframework.http.HttpStatus.CREATED
 import org.springframework.http.MediaType.APPLICATION_PDF
 import org.springframework.http.MediaType.APPLICATION_PDF_VALUE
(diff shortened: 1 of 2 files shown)


src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -583,6 +583,7 @@ paths:
         uri: '${base_uri}/eros/{eroId}/anonymous-elector-documents'
         requestParameters:
           integration.request.path.eroId: method.request.path.eroId
+          integration.request.querystring.applicationId: method.request.querystring.applicationId
```
vishalgupta-valtech, line 586: Reference : `https://docs.aws.amazon.com/apigateway/latest/developerguide/how-to-method-settings-execution-console.html` for `query` string `applicationId`. 

We might need to do the same for other GET endpoints

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -564,11 +564,11 @@ paths:
...
         - Anonymous Elector Documents (AED)
       parameters:
-        - name: eroId
-          description: The ID of the Electoral Registration Office responsible for the application.
+        - name: applicationId
+          description: An identifier of an Anonymous Voter Card Application as a `query param`
           schema:
             type: string
-          in: path
+          in: query
```
vishalgupta-valtech, line 571: ℹ️ It's a query param instead of path.

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -756,28 +758,18 @@ components:
         - printRequestSummaries
     PrintRequestSummary:
       title: PrintRequestSummary
-      type: object
       description: An object containing print request summary
+      type: object
+      allOf:
+        - $ref: '#/components/schemas/UpdatedBy'
```
vishalgupta-valtech, line 764: I have introduced a new mandatory object type that can now be reused across both summary responses
ℹ️ The response structure hasn't changed, so this is not a breaking change

sentioservices (approved):
LGTM

src/main/kotlin/uk/gov/dluhc/printapi/rest/AnonymousElectorDocumentController.kt
```diff
@@ -52,6 +57,16 @@ class AnonymousElectorDocumentController(
         }
     }
 
+    @GetMapping
+    @PreAuthorize(HAS_ERO_VC_ANONYMOUS_ADMIN_AUTHORITY)
+    @ResponseStatus(HttpStatus.OK)
```
Neil-Massey, line 62: Maybe import

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -352,7 +353,7 @@ paths:
       description: Returns the Temporary Certificate summaries for a Voter Card Application
       parameters:
         - name: applicationId
-          description: The identifier of the Voter Card Application.
+          description: The identifier of the Voter Card Application as a `query param`.
```
Neil-Massey, line 356: I don't think " as a `query param`" is needed in the description - the schema makes that very clear with the `in: query`

matt-wills-valtech (approved):
LGTM

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -370,7 +371,7 @@ paths:
         uri: '${base_uri}/eros/{eroId}/temporary-certificates'
         requestParameters:
           integration.request.path.eroId: method.request.path.eroId
-          integration.request.path.applicationId: method.request.path.applicationId
+          integration.request.querystring.applicationId: method.request.querystring.applicationId
```
Neil-Massey, line 374: Is this needed?  (When Nathan added other query parameters I checked VCA spec and it didn't look like we specifically mapped query parameters)

vishalgupta-valtech: I am unsure 🤷🏼‍♂️  as applicationId is definitely not a `path` variable, so just corrected it to be a `querystring` param.

src/main/resources/openapi/PrintAPIs.yaml
matt-wills-valtech: I don't think it does any harm, especially as `applicationId` is in the path of the vast majority of our endpoints.

src/main/resources/openapi/PrintAPIs.yaml
matt-wills-valtech: I'm not sure either, but I'd prefer to have consistency with the other relevant endpoints as well.

src/main/resources/openapi/PrintAPIs.yaml
Neil-Massey: I think if the query parameter name is the same in the AWS API GW and the endpoint the request is forwarded the passthrough behaviour means no explicit mapping is required.

src/main/resources/openapi/PrintAPIs.yaml
NathanRussellValtech: 100% that it needs removing as a path variable (I must have forgotten that when I did it 😬 )
But I'm not sure about adding it as a querystring - I'm not sure it's needed ?
(When I made this change here in print-api to make this a query string param, I took my steer from the Search Application's endpoint in VCA ..... that endpoint has loads of query string params, and we dont define them in the APIG config)

vishalgupta-valtech: Thank you all . I have removed these and `applicationId queryString` won't be mapped in APIG integration as we all agreed/discussed. PR updated 👍🏼

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -370,7 +370,6 @@ paths:
         uri: '${base_uri}/eros/{eroId}/temporary-certificates'
         requestParameters:
           integration.request.path.eroId: method.request.path.eroId
-          integration.request.path.applicationId: method.request.path.applicationId
```
NathanRussellValtech, line 373: 👍 Thank you

NathanRussellValtech (approved):
LGTM 👍

PR #190, opened 2023-03-13
Printer responsible for Issued Date (DRAFT - DO NOT MERGE).

kristerbone:
This PR is DRAFT and for review with EIP and Print Provider.

The issued date needs to be when the certificate is actually printed, therefore the print provider needs to supply the actual date of printing / issued date.

Where the certificate is a re-print, EROP needs to supply the original issuedDate as previously supplied by the printer.

### DIFF
--- a/src/main/resources/yamlschema/PrintRequest.yaml
+++ b/src/main/resources/yamlschema/PrintRequest.yaml
@@ -19,11 +19,21 @@ properties:
     maxLength: 255
     description: Issuing authority (LA/ERO/VJB)
     example: Camden Borough Council
+  issueDate:
+    type: string
+    format: date
+    description: |-
+      Only supplied in the case of a re-print. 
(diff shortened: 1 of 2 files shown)


vishalgupta-valtech:
Added temporary label to not merge, please remove the label once PR is ready/approved to be merged

NathanRussellValtech (commented):
You'll need to update the version number of the spec files - we update the version for any and all changes to our specs (no matter how small)

In addition. these print provider yml specs are a little different in this respect. Whilst there are several different files (more than the 2 you're updating here), they are all considered "holistically" as one spec. Each file contains a version number, and its the same version number. So you'll want to bump the spec version number in all yml files that make up this spec (4 yml files); plus the `README` in the same folder contains the version number complete with a change history table.

vishalgupta-valtech:
This is a month old PR so closing this @kristerbone

PR #193, opened 2023-03-14
NO-JIRA Refactor frequently failing tests

Neil-Massey:


src/test/kotlin/uk/gov/dluhc/printapi/logging/CorrelationIdMdcIntegrationTest.kt
```diff
@@ -132,43 +136,18 @@ internal class CorrelationIdMdcIntegrationTest : IntegrationTest() {
...
-                        "Batch \\[.{32}\\] containing .{1} print requests submitted to queue",
-                        Level.INFO
-                    )
-                ).hasCorrelationId(expectedCorrelationId)
-                assertThat(
-                    TestLogAppender.getLogEventMatchingRegex(
-                        "Completed batching certificate Print Requests",
-                        Level.INFO
-                    )
-                ).hasCorrelationId(expectedCorrelationId)
```
Neil-Massey, line 159: As there is a race condition and frequently the flow aborts with an `InsufficientPrintRequestsInBatchException` error. This test fails, but the intention of the test is that the correlation id is the same in the logs, so propose getting a minimum of 4 logs and verifying that they have the same correlation ID.

NathanRussellValtech (approved):
A thing of absolute beauty 👍

PR #194, opened 2023-03-14
EIP1-4634, EIP1-4635 - Listener and scheduled job to remove temporary cert data

matt-wills-valtech:
Removes `TemporaryCertificate` data after 2 years (or more accurately on the second 1st July after its issue date, which is confusing I know!)


src/main/kotlin/uk/gov/dluhc/printapi/jobs/FinalRetentionPeriodDataRemovalJob.kt
```diff
@@ -5,15 +5,18 @@ import org.springframework.scheduling.annotation.Scheduled
...
 class FinalRetentionPeriodDataRemovalJob(
-    private val certificateDataRetentionService: CertificateDataRetentionService
+    private val certificateDataRetentionService: CertificateDataRetentionService,
+    private val temporaryCertificateDataRetentionService: TemporaryCertificateDataRetentionService
 ) {
 
     @Scheduled(cron = "\${jobs.remove-vca-final-retention-period-data.cron}")
     @SchedulerLock(name = "\${jobs.remove-vca-final-retention-period-data.name}")
     fun removeVoterCardFinalRetentionPeriodData() {
+        temporaryCertificateDataRetentionService.removeTemporaryCertificateData(sourceType = VOTER_CARD)
```
matt-wills-valtech, line 19: I considered creating a separate scheduled method, but I think it's ok to remove the temporary cert data at the same time (both get removed on 1st July).

vishalgupta-valtech: Your Could use the same job, but may be control deleting of temp cert and aed data with 2 flags. 

Retain-temp-cert: true
Retain-aed-data : true

matt-wills-valtech: I don't see a need if I'm honest. I can't see why we'd want to disable the removal of data (other than a feature flag, which we don't need in this instance)?

I originally made the number of days to retain data configurable, partly because counting in days gives you a lot of flexibility (you can set it to a high number to cover months and years if necessary), but mainly because the legislation and requirements were less clear, but I think it's different here. There's both process and legislation involved and if in the unlikely event it changed from say months to days then we have to make a code change anyway.

src/main/kotlin/uk/gov/dluhc/printapi/service/TemporaryCertificateDataRetentionService.kt
```diff
@@ -0,0 +1,60 @@
...
+            )?.also {
+                it.finalRetentionRemovalDate = getFinalRetentionPeriodRemovalDate(it.issueDate)
+                temporaryCertificateRepository.save(it)
+            } ?: logger.error { "Temporary certificate with sourceType = $sourceType and sourceReference = $sourceReference not found" }
+        }
+    }
+
+    @Transactional
+    fun removeTemporaryCertificateData(sourceType: SourceType) {
+        with(temporaryCertificateRepository.findPendingRemoval(sourceType = sourceType)) {
```
sentioservices, line 45: I presume there won't be as many Temporary Certificates as VCAs / year so it's ok to delete them all in 1 transaction?

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/TemporaryCertificateRepository.kt
```diff
@@ -10,6 +10,8 @@ import java.util.UUID
 @Repository
 interface TemporaryCertificateRepository : JpaRepository<TemporaryCertificate, UUID> {
 
+    fun findByGssCodeAndSourceTypeAndSourceReference(gssCode: String, sourceType: SourceType, sourceReference: String): TemporaryCertificate?
+
```
sentioservices, line 14: Looks like there isn't a TemporaryCertificateRepositoryIntegrationTest class, was that a conscious decision because it's covered in RestIntTest?

src/test/kotlin/uk/gov/dluhc/printapi/service/TemporaryCertificateDataRetentionServiceTest.kt
```diff
@@ -0,0 +1,85 @@
...
+            // Then
+            assertThat(temporaryCertificate.finalRetentionRemovalDate).isEqualTo(expectedFinalRetentionRemovalDate)
+            assertThat(temporaryCertificate).hasFinalRetentionRemovalDate(expectedFinalRetentionRemovalDate)
+            verify(sourceTypeMapper).mapSqsToEntity(message.sourceType)
+            verify(temporaryCertificateRepository).findByGssCodeAndSourceTypeAndSourceReference(message.gssCode, VOTER_CARD, message.sourceReference)
+            verify(temporaryCertificateRepository).save(temporaryCertificate)
+        }
+
+        @Test
+        fun `should throw exception when temporary certificate doesn't exist`() {
```
sentioservices, line 61: `should log not found when temporary certificate doesn't exist` ?

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/assertj/assertions/TemporaryCertificateAssert.kt
```diff
@@ -0,0 +1,59 @@
...
+import org.assertj.core.util.CheckReturnValue
+import uk.gov.dluhc.printapi.database.entity.TemporaryCertificate
+import java.time.LocalDate
+import java.util.Objects
+import javax.annotation.Generated
+
+/**
+ * [TemporaryCertificate] specific assertions - Generated by CustomAssertionGenerator.
+ */
+@Generated(value = ["assertj-assertions-generator"])
```
sentioservices, line 13: Just for my own underanding why is `@Generated` here?

sentioservices (commented):
LGTM just a couple of questions

src/main/kotlin/uk/gov/dluhc/printapi/database/repository/TemporaryCertificateRepository.kt
matt-wills-valtech: That's right (except I'm relying on `FinalRetentionPeriodDataRemovalJobIntegrationTest`). I'm not keen on having that many ITs in general, but also this method is a copy & paste from the CertificateRepository and it seemed a bit unnecessary. I'm relying on `FinalRetentionPeriodDataRemovalJobIntegrationTest` to expose any regression issues.

src/main/kotlin/uk/gov/dluhc/printapi/service/TemporaryCertificateDataRetentionService.kt
matt-wills-valtech: Indeed, I'm assuming the vast majority of electors won't need a temp cert. Also, we're only removing entities from the DB here (not S3 as well), so I'd hope it will be a quick, reasonably efficient transaction.

src/test/kotlin/uk/gov/dluhc/printapi/service/TemporaryCertificateDataRetentionServiceTest.kt
matt-wills-valtech: Oops, copy & paste error. I'll fix the one I copied it from too.

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/assertj/assertions/TemporaryCertificateAssert.kt
matt-wills-valtech: Ah yes, copy & paste again - this time from `CertificateAssert`. I asked @Neil-Massey and he said `CertificateAssert` was generated automatically for him. This is magic I didn't know about! I'll remove the annotation here, but it would be good to find out about this voodoo.

sentioservices: Interesting! This https://plugins.jetbrains.com/plugin/10345-assertions2assertj or https://joel-costigliola.github.io/assertj/assertj-assertions-generator.html

vishalgupta-valtech (approved):
LGTM

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/assertj/assertions/TemporaryCertificateAssert.kt
vishalgupta-valtech: Nice 👍🏼

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/assertj/assertions/TemporaryCertificateAssert.kt
```diff
@@ -0,0 +1,57 @@
+package uk.gov.dluhc.printapi.testsupport.assertj.assertions
+
+import org.assertj.core.api.AbstractObjectAssert
+import org.assertj.core.util.CheckReturnValue
+import uk.gov.dluhc.printapi.database.entity.TemporaryCertificate
+import java.time.LocalDate
+import java.util.Objects
+
+/**
+ * [TemporaryCertificate] specific assertions - Generated by CustomAssertionGenerator.
```
sentioservices, line 10: Not a blocker but you could also remove `Generated by CustomAssertionGenerator` in your next PR. Sorry I missed it first time around.

sentioservices (approved):
LGTM

PR #199, opened 2023-03-17
EIP1-4776 Ensure print request event statuses are ordered deterministically

Neil-Massey:
Just to assist discussion


Neil-Massey:
Alternative approach identified as the timestamp can be stored and processed with greater precision

PR #200, opened 2023-03-17
EIP1-4653 enhance correlation id

gurpiarbassi:
- Ensure correlation id is also returned in HTTP response
- Ensure correlation id is carried through in WebClient requests or Webclient is able to generate a new one.

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
@@ -1,17 +1,25 @@
 package uk.gov.dluhc.printapi.config
 
+import mu.KotlinLogging
 import org.aspectj.lang.ProceedingJoinPoint
 import org.aspectj.lang.annotation.Around
 import org.aspectj.lang.annotation.Aspect
 import org.slf4j.MDC
+import org.springframework.http.HttpHeaders
 import org.springframework.messaging.Message
 import org.springframework.messaging.support.GenericMessage
 import org.springframework.stereotype.Component
+import org.springframework.web.reactive.function.client.ClientRequest
+import org.springframework.web.reactive.function.client.ClientResponse
+import org.springframework.web.reactive.function.client.ExchangeFilterFunction
+import org.springframework.web.reactive.function.client.ExchangeFunction
 import org.springframework.web.servlet.HandlerInterceptor
-import java.lang.Exception
+import reactor.core.publisher.Mono
 import java.util.UUID
 import javax.servlet.http.HttpServletRequest
 import javax.servlet.http.HttpServletResponse
+
+private val logger = KotlinLogging.logger {}
 
 /**
  * MVC Interceptor and AOP beans that set the correlation ID MDC variable for inclusion in all log statements.
@@ -25,11 +33,14 @@ const val CORRELATION_ID_HEADER = "x-correlation-id"
  * HTTP header `x-correlation-id` if set. This allows for passing and logging a consistent correlation ID between
  * disparate systems or processes.
  */
+
 @Component
 class CorrelationIdMdcInterceptor : HandlerInterceptor {
 
     override fun preHandle(request: HttpServletRequest, response: HttpServletResponse, handler: Any): Boolean {
-        MDC.put(CORRELATION_ID, request.getHeader(CORRELATION_ID_HEADER) ?: generateCorrelationId())
+        val correlationId = request.getHeader(CORRELATION_ID_HEADER) ?: generateCorrelationId()
+        MDC.put(CORRELATION_ID, correlationId)
+        response.setHeader(CORRELATION_ID_HEADER, correlationId)
         return true
     }
 
@@ -40,6 +51,46 @@ class CorrelationIdMdcInterceptor : HandlerInterceptor {
         ex: Exception?
     ) {
         MDC.remove(CORRELATION_ID)
+    }
+}
+
+/**
+ * Webclient exchange filter that sets the correlation ID MDC variable of either a new value, or the
+ * current value found in the MDC context. This allows for passing and logging a consistent correlation ID between
+ * disparate systems or processes using the spring Webclient.
+ */
+@Component
+class CorrelationIdMdcExchangeFilter(private val mdcExchangeFilter: MDCExchangeFilter) : ExchangeFilterFunction {
+
+    /*
+        This is modelled as a set in case we need to talk to another system within the gov space that doesn't use 'x-correlation-id'.
+        Another commonly used identifier is 'X-Request-Id'. This allows us to send our 'x-correlation-id' as well as their specified one.
+    */
+    private val correlationHeaderNames: Set<String> = setOf(CORRELATION_ID_HEADER)
+
+    override fun filter(request: ClientRequest, next: ExchangeFunction): Mono<ClientResponse> {
+        val currentCorrelationId = getCurrentCorrelationId()
+        val clientRequestModified = setCorrelationIdInRequest(request, currentCorrelationId)
+
+        return next.filter(mdcExchangeFilter).exchange(clientRequestModified)
+    }
+
+    private fun setCorrelationIdInRequest(request: ClientRequest, correlationId: String): ClientRequest {
+        return ClientRequest.from(request)
+            .headers { headers: HttpHeaders -> correlationHeaderNames.forEach { correlationHeaderName -> headers[correlationHeaderName] = correlationId } }
+            .build()
+    }
+
+    @Component
+    class MDCExchangeFilter : ExchangeFilterFunction {
+        override fun filter(request: ClientRequest, next: ExchangeFunction): Mono<ClientResponse> {
+            val contextMap = MDC.getCopyOfContextMap()
+            return next.exchange(request).doOnEach { _ ->
+                if (contextMap != null) {
+                    MDC.setContextMap(contextMap)
+                }
+            }
+        }
     }
 }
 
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
```diff
@@ -25,11 +33,14 @@ const val CORRELATION_ID_HEADER = "x-correlation-id"
...
  */
+
 @Component
 class CorrelationIdMdcInterceptor : HandlerInterceptor {
 
     override fun preHandle(request: HttpServletRequest, response: HttpServletResponse, handler: Any): Boolean {
-        MDC.put(CORRELATION_ID, request.getHeader(CORRELATION_ID_HEADER) ?: generateCorrelationId())
+        val correlationId = request.getHeader(CORRELATION_ID_HEADER) ?: generateCorrelationId()
+        MDC.put(CORRELATION_ID, correlationId)
+        response.setHeader(CORRELATION_ID_HEADER, correlationId)
```
gurpiarbassi, line 43: It's useful to set the correlation id on the response too to aid in troubleshooting.

src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
```diff
@@ -43,6 +54,46 @@ class CorrelationIdMdcInterceptor : HandlerInterceptor {
     }
 }
 
+/**
+ * Webclient exchange filter that sets the correlation ID MDC variable of either a new value, or the
+ * current value found in the MDC context. This allows for passing and logging a consistent correlation ID between
+ * disparate systems or processes using the spring Webclient.
+ */
+@Component
+class CorrelationIdMdcExchangeFilter(private val mdcExchangeFilter: MDCExchangeFilter) : ExchangeFilterFunction {
```
gurpiarbassi, line 63: This was the bit missing, that makes the Webclient correlation id aware

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -27,6 +27,12 @@ paths:
           type: string
         in: path
         required: true
+      - name: x-correlation-id
```
gurpiarbassi, line 30: It's useful to document at an API level that the contract accepts a correlation Id. I could not find a global way of doing it so had to touch all the endpoints and add a optional header. I'm open to suggestions if there are better accepted ways of documenting it.

src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
```diff
@@ -25,11 +33,14 @@ const val CORRELATION_ID_HEADER = "x-correlation-id"
  * HTTP header `x-correlation-id` if set. This allows for passing and logging a consistent correlation ID between
  * disparate systems or processes.
  */
```
NathanRussellValtech, line 35: I know you didn't add this originally (I did 😁 ), but now we have the Web**Client** code below, I've started to get a bit confused by this interceptor method as to whether its client or server related.
I know its "server" related, as it's a **Handler**Interceptor .... but I had to think about it a bit ...

Do you think its worth adding to the kdoc to this method to make it clear that it's an interceptor for our request handler methods, and therefore is **server** related rather than client

src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
NathanRussellValtech: Agreed. It also means the client (whatever that is - ie. our UI or some external such as EMS' in register checker) can set the same correlation ID on subsequent requests (if it makes sense to that flow/business process)

src/main/resources/openapi/PrintAPIs.yaml
NathanRussellValtech: This is a tricky one, and I dont know the answer
Yes, we need to document it in the openAPI spec somehow. 100%

Do we need to do it on every endpoint? I don't know. For larger APIs (such as VCA) this will add loads to the spec in terms of line count, but I wonder if its adding much (documenting every endpoint). Even in this small API (Print API is one of our smaller ones) you've had to add it to 8 endpoints
And of course if this is the approach we are taking we would need to make sure that all devs know about this to be able to consistently follow the pattern when adding new endpoints etc

Kinda related, but in the VAC repos (VCA, RCA, print-api, notifications-api, ero-management and admin-api) we made a decision a while back **not** to document standard error response codes such as 401/403/400/404/409/5xx etc. unless they added specific context or value.
IE. we don't document 401/403 because if the API returns this we'd like to think the client devs will know this is an authentication issue. Likewise 404 - we all know what that means so felt it unnecessary (to the point of being distracting noise) to explicitly document on every endpoint. 400's and 409's get documented only if there is some very specific context or business case that is not obvious (eg: you'll get a 409 CONFLICT if your application is not in status X, Y or Z)

Also, as well as documenting that it's a request header that we accept; based on the changes you made to the interceptor, its also returned as a response header. We would want to document that as well, but not in every endpoint definition 🤔 

Any thoughts? @matt-wills-valtech @stalwartsolutions @Neil-Massey @vishalgupta-valtech @adacayi @sentioservices

src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
```diff
@@ -1,18 +1,26 @@
...
+import org.springframework.web.reactive.function.client.ExchangeFilterFunction
+import org.springframework.web.reactive.function.client.ExchangeFunction
 import org.springframework.web.servlet.HandlerInterceptor
-import java.lang.Exception
+import reactor.core.publisher.Mono
 import java.util.UUID
 import javax.servlet.http.HttpServletRequest
 import javax.servlet.http.HttpServletResponse
 
+private val logger = KotlinLogging.logger {}
```
NathanRussellValtech, line 22: Is this used?
Seems a bit odd to be logging stuff in beans and components that are all about adding stuff to the MDC for logging ....

src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
NathanRussellValtech: Nice 👍 

Is it worth adding the equivalent bean for `RestTemplate` ? 
I know that `print-api` doesn't use `RestTemplate`, but if the idea/goal is that this class will be moved to a shared library one day (and in the immediate/short term, copy and pasted into all the other repos) .... some of our repos do use `RestTemplate` for some very specific use cases (VCA and RCA I think - possibly others too)

gurpiarbassi: Yes I didn't add the RestTemplate one here because there wasn't a RestTemplate in src/main that I could test. In OAVA we do have a RestTemplate that talks to IER I believe. I guess I could create one in src/test in the repo?

src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
gurpiarbassi: no it's not, need to remove.
I originally had the idea of logging a warning in cases where the inbound request does not have a correlation-id and we have to generate a new one. This is quite useful when viewing logs and you can easily spot outliers who haven't sent a correlation-id. I didn't end up doing this since it would require a bit of a refactor.

src/main/resources/openapi/PrintAPIs.yaml
matt-wills-valtech: >  I could not find a global way of doing it so had to touch all the endpoints and add a optional header

Me neither, which is a shame. So if we want to avoid adding this to every endpoint, which is completely understandable, then perhaps one "better than nothing" option is to add some info to the description at the start of the spec? We currently have:

```
info:
  title: Print APIs
  version: '1.11.0'
  description: Print APIs
```

Whatever we decide, this description could be considerably better than just "Print APIs"! 😆

src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
gurpiarbassi: Wont add RestTemplate interceptor as part of this PR. It'll get added in OAVA and hopefully by then we can move to a shared artefact repo.

src/main/kotlin/uk/gov/dluhc/printapi/config/CorrelationIdMdcConfiguration.kt
gurpiarbassi: Added Kdoc so it's clear which one is for WebClient and which one is for RestControllers

src/main/kotlin/uk/gov/dluhc/printapi/config/WebClientConfiguration.kt
```diff
@@ -9,8 +9,9 @@ import org.springframework.web.reactive.function.client.WebClient
 class WebClientConfiguration {
 
     @Bean
-    fun eroManagementWebClient(@Value("\${api.ero-management.url}") eroManagementApiUrl: String): WebClient =
+    fun eroManagementWebClient(@Value("\${api.ero-management.url}") eroManagementApiUrl: String, correlationIdExchangeFilter: CorrelationIdMdcExchangeFilter): WebClient =
```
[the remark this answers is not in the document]
gurpiarbassi: 1. No other WebClient instances here
2.Updated Kdoc.

NathanRussellValtech (approved):
LGTM 👍 Thanks for making the changes 👍

PR #201, opened 2023-03-20
EIP1-4574: EIP1:4669: Adds (OPTIONAL) deliveryInfo column to anonymous_elector_document entity

vishalgupta-valtech:
Note: 
- The new column is nullable.
- No migrations scripts will be added


src/main/resources/db/changelog/ddl/0020_alter_anonymous_elector_document.xml
```diff
@@ -0,0 +1,22 @@
...
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+
+    <changeSet author="vishal.gupta@valtech.com" id="0020_EIP-4669_anonymous_elector_document add delivery info column" context="ddl">
+
+        <addColumn tableName="anonymous_elector_document">
+            <column name="delivery_id" type="uuid" afterColumn="request_date_time">
+                <constraints nullable="true"
```
vishalgupta-valtech, line 12: Note: It's a nullable column and we are not adding any new migration scripts for this.

matt-wills-valtech: It needs to be nullable as we are removing the delivery related info after 15 months (before the data that's on the printed document itself, which is removed after 10 years).

matt-wills-valtech (approved):
LGTM (especially all the static imports! 🤣)

src/main/resources/db/changelog/ddl/0020_alter_anonymous_elector_document.xml
```diff
@@ -0,0 +1,22 @@
...
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+
+    <changeSet author="vishal.gupta@valtech.com" id="0020_EIP-4669_anonymous_elector_document add delivery info column for data retention" context="ddl">
+
+        <addColumn tableName="anonymous_elector_document">
+            <column name="delivery_id" type="uuid" afterColumn="request_date_time" remarks="This nullable column is required for AED certificate data retention">
+                <constraints nullable="true"
+                             foreignKeyName="fk_aed_certificate_delivery"
+                             references="delivery(id)"/>
```
sentioservices, line 14: It's interesting that declaring the foreignKey results in an index being created.
I'm not sure when that index will be used, but thanks for doing the screenshots so that we are aware.

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/assertj/assertions/AnonymousElectorDocumentCertificateAssert.kt
```diff
@@ -39,4 +60,35 @@ class AnonymousElectorDocumentCertificateAssert
 
         return this
     }
+
+    fun hasDelivery(delivery: Delivery): AnonymousElectorDocumentCertificateAssert {
+        isNotNull
+
+        DeliveryAssert(actual!!.delivery)
```
sentioservices, line 67: minor (non-blocking) - I'm slightly surprised you didn't use your usual pattern of `with(delivery) {...}` :)

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/testdata/entity/CertificateBuilder.kt
```diff
@@ -143,12 +143,12 @@ fun buildElectoralRegistrationOffice(
...
 
 fun buildAddress(
-    street: String = DataFaker.faker.address().streetName(),
-    postcode: String = DataFaker.faker.address().postcode(),
-    property: String? = DataFaker.faker.address().buildingNumber(),
-    locality: String? = DataFaker.faker.address().streetName(),
-    town: String? = DataFaker.faker.address().city(),
-    area: String? = DataFaker.faker.address().state(),
+    street: String = faker.address().streetName(),
+    postcode: String = faker.address().postcode(),
```
sentioservices, line 147: minor (non-blocking) I know you didn't write as part of this pr but this creates a new fake address object, takes one attribute, creates a whole new fake address object, takes one attribute, ...
I think when creating fake addresses we should do something like this 
```
    fakeAddress: net.datafaker.providers.base.Address = faker.address(),
    street: String = fakeAddress.streetName(),
    postcode: String = fakeAddress.postcode(),
```

sentioservices (approved):
LGTM

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/assertj/assertions/AnonymousElectorDocumentCertificateAssert.kt
vishalgupta-valtech: I have `renamed` the function param and I think earlier I did a copy paste of this function.
However without with, it's looking as more readable I think :)

PR #204, opened 2023-03-24
EIP1-3229: EIP1:4618: api spec for Get AED summaries

vishalgupta-valtech:
image

image

### DIFF
--- a/src/main/resources/openapi/PrintAPIs.yaml
+++ b/src/main/resources/openapi/PrintAPIs.yaml
@@ -1115,6 +1115,7 @@ components:
       description: Status of an Anonymous Elector document
       type: string
       enum:
+        - generated
         - printed
 
     GenerateAnonymousElectorDocumentRequest:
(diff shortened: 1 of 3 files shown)


src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1049,32 +1050,71 @@ components:
...
+      required:
+        - addressee
+        - registeredAddress
+
     AnonymousElectorDocumentStatus:
       title: AnonymousElectorDocumentStatus
       description: Status of an Anonymous Elector document
       type: string
       enum:
-        - generated
```
vishalgupta-valtech, line 1077: We never set this value anymore, so removed. All db records were already set to PRINTED in previous PRs

vishalgupta-valtech: Good point, but I am not sure if they will ever get this back from api , I will ask @gavinthomas-valtech

sentioservices (approved):
LGTM, but if it is a breaking change for the FE then obviously worth letting them know

src/main/resources/openapi/PrintAPIs.yaml
vishalgupta-valtech: UI team is away therefore I have reverted change that removes `GENERATED` from api spec and entity

PR #205, opened 2023-03-25
EIP1-4953 - Consistent use of ErrorResponse

NathanRussellValtech:
This PR updates the openAPI spec and various implementations to make the declaration of REST API error statuses and the implementation consistent (consistent both in terms of a consistent API, and also consistent with VCA in respect of implementation)


Neil-Massey (approved):
LGTM

PR #207, opened 2023-03-27
EIP1-3229: EIP1:5006: Mappers to map Get AED summaries entity->dto->api models

vishalgupta-valtech:


matt-wills-valtech (commented):
LGTM

src/main/kotlin/uk/gov/dluhc/printapi/dto/AnonymousElectorDocumentSummaryDto.kt
```diff
@@ -0,0 +1,29 @@
...
+)
+
+class AnonymousElectorDto(
+    val addressee: String,
+    val registeredAddress: AddressDto,
+)
+
+enum class AnonymousElectorDocumentStatus {
+    GENERATED,
+    PRINTED,
```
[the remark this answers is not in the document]
vishalgupta-valtech: Yeah agree, I didn't wanted to write a status mapper to `not` convert `GENERATED` entity status to DTO and API. So it stays unless we remove this entirely, however as you noticed, it's being set to PRINTED in mapper

src/main/kotlin/uk/gov/dluhc/printapi/dto/AnonymousElectorDocumentSummaryDto.kt
```diff
@@ -0,0 +1,29 @@
...
+    val gssCode: String,
+    val certificateLanguage: CertificateLanguage,
+    val supportingInformationFormat: SupportingInformationFormat,
+    val deliveryAddressType: DeliveryAddressType,
+    val elector: AnonymousElectorDto,
+    val status: AnonymousElectorDocumentStatus,
+    val photoLocation: String,
+    val issueDate: LocalDate,
+    val userId: String,
+    val dateTime: Instant,
```
sentioservices, line 18: what happened at this dateTime?

sentioservices (approved):
LGTM, just a minor query around attribute name

src/main/kotlin/uk/gov/dluhc/printapi/dto/AnonymousElectorDocumentSummaryDto.kt
vishalgupta-valtech: This is now renamed to `requestDateTime` (_when document generation was requested_) and the name is now same as the one in entity. 

The `dateTime` name in api spec can't be changed as that's part of `UpdatedBy` object

PR #212, opened 2023-03-31
HOTFIX: EIP1-5198: EIP1-5200: Adds AED supporting format that accepts all values

vishalgupta-valtech:


src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Print APIs
-  version: '1.12.3'
+  version: '1.13.0'
```
matt-wills-valtech, line 4: Minor - why not 1.12.4?

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -786,10 +786,20 @@ components:
         - dateTime
 
     SupportingInformationFormat:
+      title: SupportingInformationFormat
```
vishalgupta-valtech, line 789: Can't extend this as VCA and other flows uses this, so had to add a new format enum

matt-wills-valtech (approved):

sentioservices (approved):
LGTM

src/main/resources/openapi/PrintAPIs.yaml
Neil-Massey: Should the temporary certificate not also support the other supporting information formats?

matt-wills-valtech: Not sure I understand the question? @vishalgupta-valtech's change doesn't affect temp certs.

sentioservices (approved):
LGTM

src/main/resources/openapi/PrintAPIs.yaml
vishalgupta-valtech: @Neil-Massey I don't see temporary certs using `supportingInformationFormat` in spec before and also they are different to AEDs as AEDs needs to be generated (if requested) after actual removal from vca.

image

matt-wills-valtech (approved):
LGTM

src/main/resources/openapi/PrintAPIs.yaml
Neil-Massey: Sorry, missed two things, 1 this is a bug fix and 2 the endpoint I was thinking doesn't take a supporting document format

```
/eros/{eroId}/temporary-certificates/{gssCode}/explainer-document
```

PR #215, opened 2023-04-03
EIP1-5159 - Added CorrelationIdRestTemplateClientHttpRequestInterceptor

NathanRussellValtech:
This PR adds `CorrelationIdRestTemplateClientHttpRequestInterceptor` that can be used with `RestTemplate`; and also updates the integration tests so that the config class and its tests can be used "outside" of print-api (ie. so that we can move it into a shared library)

(not all the tests have been abstracted from print-api - still need to do the tests for the SQS listener and senders, and the scheduled jobs - we can look to do these another time)


adacayi (approved):
LGTM, @NathanRussellValtech , thanks for simplifying the existing MDC integration tests as well.

PR #217, opened 2023-04-04
EIP1-5311 - Refactored CorrelationIdMdcIntegrationTest

NathanRussellValtech:
Refactored CorrelationIdMdcIntegrationTest so it is abstracted away from any print-api specific implementations
This means it will be much easier to pull out into the other API repos, and then into a shared library as and when we get there

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -198,7 +198,7 @@ thread-pool:
 
 logging:
   pattern:
-    level: "%X{correlationId}%5p"
+    level: "%X{correlationId} %5p"
   level:
     com:
       jcraft:
--- a/src/test/kotlin/uk/gov/dluhc/printapi/logging/CorrelationIdMdcIntegrationTest.kt
+++ b/src/test/kotlin/uk/gov/dluhc/printapi/logging/CorrelationIdMdcIntegrationTest.kt
@@ -8,7 +8,7 @@ import com.github.tomakehurst.wiremock.client.WireMock.equalTo
 import com.github.tomakehurst.wiremock.client.WireMock.getRequestedFor
 import com.github.tomakehurst.wiremock.client.WireMock.urlEqualTo
 import com.github.tomakehurst.wiremock.matching.StringValuePattern
-import org.assertj.core.api.Assertions.assertThat
+import io.awspring.cloud.messaging.core.QueueMessagingTemplate
 import org.awaitility.kotlin.await
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Nested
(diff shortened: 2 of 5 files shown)


src/main/resources/application.yml
```diff
@@ -198,7 +198,7 @@ thread-pool:
 
 logging:
   pattern:
-    level: "%X{correlationId}%5p"
+    level: "%X{correlationId} %5p"
```
NathanRussellValtech, line 201: Minor correction to the logging pattern. Without the space it means that there is no space between the correlationID and the logger level when the logger level is 5 characters wide (such as `ERROR`)

src/test/kotlin/uk/gov/dluhc/printapi/logging/CorrelationIdMdcIntegrationTest.kt
```diff
@@ -56,6 +44,12 @@ internal class CorrelationIdMdcIntegrationTest : IntegrationTest() {
     @Autowired
     private lateinit var correlationIdRestTemplateClientHttpRequestInterceptor: CorrelationIdRestTemplateClientHttpRequestInterceptor
 
+    @Autowired
+    private lateinit var testScheduledJob: TestScheduledJob
```
adacayi, line 48: One question, @NathanRussellValtech ,

We usually add the autowired properties to the `IntegrationTest` class to avoid Spring spinning up a new context and slow down our integration tests (I guess that was the reason behind it).

Should we not follow the same pattern here as well?

adacayi (approved):
LGTM, @NathanRussellValtech.

src/test/kotlin/uk/gov/dluhc/printapi/logging/CorrelationIdMdcIntegrationTest.kt
NathanRussellValtech: Yes, an excellent question, and you are right.
In this case I went with putting them straight in the test class because 
a) we want to be able to lift this code, inc this test class, straight out into the other VCA apps, and I'm aiming to be able to do it without having to mess (too much) with their integration test setup
b) longer term (and tbh not too far in the future) we'll be looking at pulling this out entirely into its own shared library artefact ... at which point we'll probably not have an IntegrationTest superclass, and just annotate the test class here  with SpringBootTest

Does that make sense? Or have I waffled a bit? 🤣

adacayi: That definitely makes sense, @NathanRussellValtech. Thanks for the answer.

PR #220, opened 2023-04-11
EIP1-5345 Refactor the logging to be in separate package for reuse.

Neil-Massey:
Refactored the logging code from this module to a [logging module in EIP ERO Portal](https://github.com/cabinetoffice/eip-ero-portal/pull/294) and then copied those refactored tested classes into this microservice.


src/test/kotlin/uk/gov/dluhc/logging/config/CorrelationIdMdcMessageListenerAspectTest.kt
```diff
@@ -0,0 +1,68 @@
...
+
+import ch.qos.logback.classic.Level
+import io.awspring.cloud.messaging.core.QueueMessagingTemplate
+import org.awaitility.kotlin.await
+import org.junit.jupiter.api.BeforeEach
+import org.junit.jupiter.api.Test
+import org.slf4j.MDC
+import org.springframework.beans.factory.annotation.Autowired
+import uk.gov.dluhc.logging.testsupport.assertj.assertions.ILoggingEventAssert
+import uk.gov.dluhc.printapi.config.IntegrationTest
```
NathanRussellValtech, line 11: Hmmm, I had a feeling this would happen .... the `logging` package now has a dependency on the `printapi` package .... 🤔 

Even though we said we'd not want to do this, I wonder if there might be a case for 2 spring boot test contexts? (especially as its short'ish lived, and will one day move into the shared artefact repo)

What if there were an IntegrationTest superclass within the logging package somewhere that just brought in the beans we care about here ?

And perhaps we copy and paste the other dependencies (test log appender, and messageQueue, and anything else the other tests need ??)

We will have to do this anyway as and when we move it into the shared repo?

It might start identifying other elements that could be shared artefacts .... such as some of the test support stuff like the test log appender etc .... 🤔

adacayi (approved):
LGMT, @Neil-Massey ,

There is one minor issue about empty files under the config package in test. 
I think these need to be deleted at some point `TestRestController`, `TestScheduledJob`, `TestSqsListener`.

src/test/kotlin/uk/gov/dluhc/logging/config/CorrelationIdMdcMessageListenerAspectTest.kt
adacayi: > What if there were an IntegrationTest superclass within the logging package somewhere that just brought in the beans we care about here ?

> And perhaps we copy and paste the other dependencies (test log appender, and messageQueue, and anything else the other tests need ??)

I agree @NathanRussellValtech.. Makes sense to me to do it as part of this work.

Neil-Massey: Problem is that the test context will be based on the dependencies defined in the Gradle build file and when we move the code to a logging library the tests need to be re-written.  An example logging library module has been aded to `eip-ero-portal` but its `IntegrationTest` does not copy into this branch (as was hoped) due to the dependencies and Spring Boot auto-configuration.

src/test/kotlin/uk/gov/dluhc/printapi/config/IntegrationTest.kt
```diff
@@ -56,7 +60,11 @@ private val logger = KotlinLogging.logger {}
  */
 @SpringBootTest(
     webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT,
-    classes = [IntegrationTest.IntegrationTestConfiguration::class],
+    classes = [
+        PrintApiApplication::class,
+        TestRestController::class, TestScheduledJob::class, TestSqsListener::class,
```
NathanRussellValtech, line 65: Presumably we no longer need TestRestController::class, TestScheduledJob::class, TestSqsListener::class,
in which case the classes can be deleted and this IntegrationTest class reverted and make no changes to it?

NathanRussellValtech (approved):
This looks great to me, and as discussed happy that we've removed the tests and the reasons why 👍 
All good my friend 👍

PR #221, opened 2023-04-12
EIP1-5345 Refactor logging to use code from library

Neil-Massey:
Logging library is deployed to CodeArtifact and the GitHub user for running the build and tests is updated with permissions to access the library.

Need to verify with a non-admin user that they are able to build this branch locally - James has updated dev permissions that should allow this.


src/test/kotlin/uk/gov/dluhc/printapi/logging/CorrelationIdMdcIntegrationTest.kt
```diff
@@ -18,8 +18,8 @@ import org.springframework.beans.factory.annotation.Autowired
 import org.springframework.boot.test.web.client.TestRestTemplate
```
Neil-Massey, line 18: Not sure whether a version of this is required after code moved to library.

build.gradle.kts
```diff
@@ -31,8 +32,18 @@ allOpen {
     annotations("javax.persistence.Entity", "javax.persistence.MappedSuperclass", "javax.persistence.Embedabble")
 }
 
+val codeArtifactToken = System.getenv("CODEARTIFACT_PAT")
+    ?: "aws codeartifact get-authorization-token --domain erop-artifacts --domain-owner 063998039290 --query authorizationToken --output text --profile code-artifact".runCommand()
```
Neil-Massey, line 36: Simplify if possible, and possibly can pass the profile as environment param instead of the `CODEARTIFACT_PAT`

adacayi (approved):
It works for me @Neil-Massey .

NathanRussellValtech (approved):
LGTM 👍

sentioservices (approved):
LGTM, and thanks for your help spotting the rogue `AWS_REGION` env var 👍

README.md
```diff
@@ -11,6 +11,24 @@ $ ./gradlew ktlintApplyToIdea
...
+aws configure --profile code-artifact
+```
+
+At the prompts configure the `code-artifact` profile as follows:
+* Your AWS Access Key ID
+* Your AWS Secret Access Key
+* Default region name, `eu-west-2`
+* Default output format, `json`
+
+Note: AWS CLI must be installed on the developer workstation as a pre-requisite.
```
RichardBradley, line 30: Will this work with MFA access?

I think currently the EROP accounts don't use MFA on CLI access, but we aim to fix that in the medium term future

Neil-Massey: I know that my account is MFA enabled and this approach works fine, I've not heard any developer having issues and this is the approach outlined in the[ AWS docs](https://docs.aws.amazon.com/codeartifact/latest/ug/maven-gradle.html).

So, I've looked at the [AWS CLI docs](https://docs.aws.amazon.com/cli/latest/reference/codeartifact/index.html) but haven't spotted anything specific to MFA.  

Possibly, this will have to be reviewed when the EROP accounts are updated.

RichardBradley: > my account is MFA enabled and this approach works fine

Do you mean that you use MFA to log into the Web Console, or when you take actions on the CLI or both?

It's a common pattern to use MFA when logging into the Web Console but to use single factor "AWS_SECRET_ACCESS_KEY" creds when working on the CLI, as it's a significant faff to create temp MFA enabled sessions each time you do the latter.

See https://repost.aws/knowledge-center/authenticate-mfa-cli

Neil-Massey: Yes, I meant I have MFA enabled for the AWS console - not for CLI.

Neil-Massey: That does seem like a faff :-(

PR #226, opened 2023-04-18
EIP1-5090: EIP1-5524: send Certificate Failed To Print email

sentioservices:
This PR adds support to send a Certificate Failed To Print email when a `FAILED` print response is received from the printer.
Depends on [infra PR to add SES policy for print API](https://github.com/cabinetoffice/eip-ero-infra/pull/775)


src/main/kotlin/uk/gov/dluhc/printapi/config/EmailContentProperties.kt
```diff
@@ -15,7 +15,8 @@ class EmailContentConfiguration(
     portalBaseUrl: URI,
     vacContextRoot: String,
     aedContextRoot: String,
-    val certificateReturned: EmailContentProperties
+    val certificateReturned: EmailContentProperties,
+    val certificateFailedToPrint: EmailContentProperties,
```
vishalgupta-valtech, line 19: Sorry I didn't understand the need of adding 2 content properties of same type? Can't this be captured in 1 val?

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/AbstractEmailSender.kt
```diff
@@ -0,0 +1,30 @@
+package uk.gov.dluhc.printapi.messaging.service
+
+import uk.gov.dluhc.printapi.client.ElectoralRegistrationOfficeManagementApiClient
+import uk.gov.dluhc.printapi.database.entity.Certificate
+import uk.gov.dluhc.printapi.database.entity.PrintRequest
+import uk.gov.dluhc.printapi.messaging.models.ProcessPrintResponseMessage
+
+abstract class AbstractEmailSender(
+    private val electoralRegistrationOfficeManagementApiClient: ElectoralRegistrationOfficeManagementApiClient,
```
vishalgupta-valtech, line 9: Can you not use `EroService` instead?

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/CertificateFailedToPrintEmailSender.kt
```diff
@@ -0,0 +1,41 @@
...
+import mu.KotlinLogging
+import org.springframework.stereotype.Component
+import uk.gov.dluhc.printapi.client.ElectoralRegistrationOfficeManagementApiClient
+import uk.gov.dluhc.printapi.database.entity.Certificate
+import uk.gov.dluhc.printapi.dto.SendCertificateFailedToPrintEmailRequest
+import uk.gov.dluhc.printapi.messaging.models.ProcessPrintResponseMessage
+
+private val logger = KotlinLogging.logger {}
+
+@Component
```
vishalgupta-valtech, line 12: This is within service package, so should this be `@Service` instead of a `@Component`?
Also Service be appended to class name?


- Also applies to `CertificateNotDeliveredEmailSender` class below

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/CertificateFailedToPrintEmailSender.kt
```diff
@@ -0,0 +1,41 @@
...
+                    sourceReference = sourceReference!!,
+                    applicationReference = applicationReference!!,
+                    localAuthorityEmailAddresses = setOf(getLocalAuthorityEmailAddress(gssCode!!)),
+                    requestingUserEmailAddress = getRequestingUserEmailAddress(printRequests, printResponse.requestId)
+                )
+            }
+            emailService.sendCertificateFailedToPrintEmail(request)
+        } catch (e: Exception) {
+            logger.error(
+                "failed to send Certificate Failed To Print email when processing ProcessPrintResponseMessage for " +
```
vishalgupta-valtech, line 35: Minor: `failed to send [Certificate Failed To Print] email when ...`

Also applies to `CertificateNotDeliveredEmailSender` class below

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/CertificateFailedToPrintEmailSender.kt
```diff
@@ -0,0 +1,41 @@
...
+private val logger = KotlinLogging.logger {}
+
+@Component
+class CertificateFailedToPrintEmailSender(
+    electoralRegistrationOfficeManagementApiClient: ElectoralRegistrationOfficeManagementApiClient,
+    private val emailService: EmailService,
+) : AbstractEmailSender(electoralRegistrationOfficeManagementApiClient) {
+
+    override fun testAndSend(printResponse: ProcessPrintResponseMessage, certificate: Certificate): Boolean {
+        if (printResponse.status != ProcessPrintResponseMessage.Status.FAILED) {
```
vishalgupta-valtech, line 19: My understanding is that this service method will only be called when the print response status has failed, so unsure why do we need to do these checks within these components/services. Same query for `CertificateNotDeliveredEmailSender` if loop check

src/main/kotlin/uk/gov/dluhc/printapi/config/EmailContentProperties.kt
vishalgupta-valtech: This class is called as `EmailContentProperties` whereas it's actually EmailContentConfiguration. Could you please check this for me?

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/PrintResponseProcessingService.kt
```diff
@@ -115,37 +110,13 @@ class PrintResponseProcessingService(
 
         certificateRepository.save(certificate)
 
-        sendEmailIfCertificateNotDelivered(newStatus, certificate, printResponse.requestId)
+        sendOnlyFirstMatchingEmail(
+            { certificateNotDeliveredEmailSender.testAndSend(printResponse, certificate) },
```
vishalgupta-valtech, line 114: I think it's better to check the print status everytime before invoking both these calls here and within those methods, you could still safeguard it if needed

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/AbstractEmailSender.kt
```diff
@@ -0,0 +1,30 @@
...
+) {
+    /**
+     * @param printResponse the printResponse that could trigger an email being sent
+     * @param certificate the certificate associated to the printResponse
+     * @return true if an email was sent otherwise false
+     */
+    abstract fun testAndSend(
```
matt-wills-valtech, line 16: I'm not sure about the name. I was initially alarmed by the thought of "test" code in production, but it's not that kind of test 😆 

Perhaps it would be clearer and quicker to grasp if you did the check in the client, before calling this method. For example:

```
if (printResponse.statusStep != ProcessPrintResponseMessage.StatusStep.NOT_MINUS_DELIVERED) {
   certificateNotDeliveredEmailSender.send(printResponse, certificate)
}
```

src/main/kotlin/uk/gov/dluhc/printapi/config/EmailContentProperties.kt
matt-wills-valtech: This confused me initially - I think the file just needs to be renamed to `EmailContentConfiguration`.

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/CertificateFailedToPrintEmailSender.kt
matt-wills-valtech: I'm on the fence slightly here. In terms of the annotations, it's essentially just syntactic sugar and you could argue this isn't a traditional "service" type class which interacts with a repository and executes transactions etc.

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/CertificateFailedToPrintEmailSender.kt
matt-wills-valtech: Yes, I've added a similar comment 👍

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/CertificateFailedToPrintEmailSender.kt
matt-wills-valtech: That really is minor 😉 😆

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/EmailService.kt
```diff
@@ -6,14 +6,29 @@ import org.springframework.stereotype.Service
...
+        val emailConfig = emailContentConfiguration.certificateReturned
+        sendEmail(request, emailConfig)
+    }
+
+    fun sendCertificateFailedToPrintEmail(request: SendCertificateFailedToPrintEmailRequest) {
+        val emailConfig = emailContentConfiguration.certificateFailedToPrint
+        sendEmail(request, emailConfig)
+    }
+
+    private fun EmailService.sendEmail(
```
matt-wills-valtech, line 28: I assume this extension function on itself is a typo?

src/test/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintResponseMessageListenerIntegrationTest.kt
```diff
@@ -138,7 +139,80 @@ internal class ProcessPrintResponseMessageListenerIntegrationTest : IntegrationT
...
+                    <p>Printing has failed for the application ${certificate.applicationReference}</p>
+                    <p>You can access the application here:</p>
+                    <p><a href="https://erop.ierds.uk/voter-authority-certificate/${certificate.applicationReference}">
+                                https://erop.ierds.uk/voter-authority-certificate/${certificate.applicationReference}</a>
+                    </p>
+                    <br>
+                    <p>This is an automated message from the ERO Portal.  You will not be able to reply to this email.</p>
+                  </body>
+                </html>
+            """.lines().joinToString(separator = "\\s*") { it.trim() }
```
matt-wills-valtech, line 205: Minor - this is very thorough, but also very brittle (perhaps deliberately?). At alternative would be to check for the dynamic values only.

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/AbstractEmailSender.kt
sentioservices: Not with adding a new method that would just invoke the electoralRegistrationOfficeManagementApiClient

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/CertificateFailedToPrintEmailSender.kt
sentioservices: 👍  I've moved these checks up to PrintResponseProcessingService#processPrintResponse

src/test/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintResponseMessageListenerIntegrationTest.kt
sentioservices: I've simplified this to just test for the application reference.

src/main/kotlin/uk/gov/dluhc/printapi/config/EmailContentProperties.kt
sentioservices: Renamed file to EmailContentConfiguration.kt

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/PrintResponseProcessingService.kt
```diff
@@ -115,37 +110,10 @@ class PrintResponseProcessingService(
...
-            } catch (e: EmailNotSentException) {
-                logger.error(
-                    "failed to send Certificate Not Delivered email when processing ProcessPrintResponseMessage for " +
-                        "certificate [${certificate.id}] with requestId [$printResponseRequestId]: ${e.message}"
-                )
-            }
+        if (printResponse.statusStep == ProcessPrintResponseMessage.StatusStep.NOT_MINUS_DELIVERED) {
+            certificateNotDeliveredEmailSenderService.send(printResponse, certificate)
+        } else if (printResponse.status == ProcessPrintResponseMessage.Status.FAILED) {
+            certificateFailedToPrintEmailSenderService.send(printResponse, certificate)
```
matt-wills-valtech, line 116: Nice and clear 👍  

I was about to suggest you could use a `when` statement instead, but unfortunately it's not the same variable being evaluated :(

matt-wills-valtech (approved):
LGTM

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/PrintResponseProcessingService.kt
```diff
@@ -110,13 +110,10 @@ class PrintResponseProcessingService(
...
-            { certificateNotDeliveredEmailSender.testAndSend(printResponse, certificate) },
-            { certificateFailedToPrintEmailSender.testAndSend(printResponse, certificate) },
-        )
-    }
-
-    private fun sendOnlyFirstMatchingEmail(vararg emailSenders: () -> Boolean) {
-        emailSenders.firstOrNull { it.invoke() }
+        if (printResponse.statusStep == ProcessPrintResponseMessage.StatusStep.NOT_MINUS_DELIVERED) {
+            certificateNotDeliveredEmailSenderService.send(printResponse, certificate)
+        } else if (printResponse.status == ProcessPrintResponseMessage.Status.FAILED) {
```
vishalgupta-valtech, line 115: Should there be else if or just 2 ifs? I thought these 2 send emails are not dependant upon each other

vishalgupta-valtech (approved):
LGTM and thanks for addressing comments. Just 1 query around use of `else if` or `if`

src/main/kotlin/uk/gov/dluhc/printapi/messaging/service/PrintResponseProcessingService.kt
sentioservices: When we get a PrintResponse with a StatusStep of NOT_MINUS_DELIVERED then its status will be FAILED. 
I didn't think it made sense to send 2 emails for the 1 event hence the `else if`.

PR #228, opened 2023-04-20
EIP1-5620: Api spec for searching AED applications

vishalgupta-valtech:
**Updated spec**
image

**Old spec**
image

### DIFF
--- a/src/main/resources/openapi/PrintAPIs.yaml
+++ b/src/main/resources/openapi/PrintAPIs.yaml
@@ -469,7 +469,7 @@ paths:
   #
   # Anonymous Elector Document Summaries
   # --------------------------------------------------------------------------------
-  '/eros/{eroId}/anonymous-elector-document-summaries':
+  '/eros/{eroId}/anonymous-elector-documents/search':
     parameters:
       - name: eroId
         description: The ID of the Electoral Registration Office responsible for the Anonymous Elector Documents.
@@ -548,7 +548,7 @@ paths:
         - eroUserCognitoUserPoolAuthorizer: [ ]
       x-amazon-apigateway-integration:
         type: HTTP_PROXY
-        uri: '${base_uri}/eros/{eroId}/anonymous-elector-document-summaries'
+        uri: '${base_uri}/eros/{eroId}/anonymous-elector-documents/search'
         requestParameters:
           integration.request.path.eroId: method.request.path.eroId
         responseParameters:
@@ -1234,7 +1234,7 @@ components:
         applicationReference:
           type: string
           description: The reference number of an Anonymous elector document application
-          example: A3JSZC4CRH
+          example: V3JSZC4CRH
         certificateNumber:
           type: string
           description: The unique identifier of the Anonymous elector document. A 20 character alphanumeric string.
--- a/src/main/resources/openapi/PrintAPIs.yaml
+++ b/src/main/resources/openapi/PrintAPIs.yaml
@@ -467,12 +467,12 @@ paths:
 
 
   #
-  # Anonymous Elector Document Applications
+  # Anonymous Elector Document Summaries
   # --------------------------------------------------------------------------------
-  '/eros/{eroId}/anonymous-applications':
+  '/eros/{eroId}/anonymous-elector-document-summaries':
     parameters:
       - name: eroId
-        description: The ID of the Electoral Registration Office responsible for the application(s).
+        description: The ID of the Electoral Registration Office responsible for the Anonymous Elector Documents.
         schema:
           type: string
         example: 'city-council'
@@ -483,7 +483,7 @@ paths:
       description: |
         Enable CORS by returning correct headers
       tags:
-        - Anonymous Elector Document Applications
+        - Anonymous Elector Document Summaries
       responses:
         200:
           description: Default response for CORS method
@@ -516,38 +516,21 @@ paths:
               application/json: |
                 {}
     get:
-      summary: Search for Anonymous Elector Document Applications
-      description: 'Search for Anonymous Elector Document Applications based on querying parameters'
+      summary: Search for Anonymous Elector Document Summaries
+      description: 'Search for Anonymous Elector Document summaries based on querying parameters'
       tags:
-        - Anonymous Elector Document Applications
+        - Anonymous Elector Document Summaries
       responses:
         '200':
-          $ref: '#/components/responses/SearchApplications'
-      operationId: search-anonymous-elector-document-applications
+          $ref: '#/components/responses/AedSearchResults'
+      operationId: search-anonymous-elector-document-summaries
       parameters:
-        - name: pageSize
-          description: The number of results to return in each page
-          schema:
-            type: integer
-            minimum: 1
-            maximum: 200
-          example: 100
-          in: query
-          required: true
-        - name: page
-          description: The page number being requested (with the first page being 1)
-          schema:
-            type: integer
-            minimum: 1
-          example: 1
-          in: query
-          required: true
         - name: searchBy
           description: |
-            The data field to apply the `searchValue` to. An empty value will return all AED applications.
+            The data field to apply the `searchValue` to. An empty value will return all AED documents.
             This value must be provided when `searchValue` is present
           schema:
-            $ref: '#/components/schemas/SearchBy'
+            $ref: '#/components/schemas/AedSearchBy'
           example: surname
           in: query
           required: false
@@ -565,7 +548,7 @@ paths:
         - eroUserCognitoUserPoolAuthorizer: [ ]
       x-amazon-apigateway-integration:
         type: HTTP_PROXY
-        uri: '${base_uri}/eros/{eroId}/anonymous-applications'
+        uri: '${base_uri}/eros/{eroId}/anonymous-elector-document-summaries'
         requestParameters:
           integration.request.path.eroId: method.request.path.eroId
         responseParameters:
@@ -860,14 +843,12 @@ components:
         - dispatched
         - not-delivered
 
-    SearchBy:
-      title: SearchBy
-      description: An enumeration of search by fields
+    AedSearchBy:
+      title: AedSearchBy
+      description: An enumeration of search by fields for searching AEDs
       type: string
       enum:
         - surname
-        - applicationReference
-        - electoralRollNumber
 
     SourceType:
       title: SourceType
@@ -1221,42 +1202,22 @@ components:
         - generated
         - printed
 
-    ApplicationsSearchResponse:
-      title: ApplicationsSearchResponse
-      description: Response containing the meta data and results of searching for Anonymous Elector document applications
+    AedSearchSummaryResponse:
+      title: AedSearchSummaryResponse
+      description: Response containing the meta data and results of searching for Anonymous Elector document summaries
       type: object
       properties:
-        page:
-          type: integer
-          description: The page number
-          example: 1
-        pageSize:
-          type: integer
-          description: The number of results per page
-          example: 1
-        totalPages:
-          type: integer
-          description: The total number of pages
-          example: 1
-        totalResults:
-          type: integer
-          description: The total number of results
-          example: 1
         results:
           type: array
-          description: Array of anonymous application search summaries that matches the search criteria
+          description: Array of anonymous elector document search summaries that matches the search criteria
           items:
-            $ref: '#/components/schemas/ApplicationSearchSummaryResponse'
+            $ref: '#/components/schemas/AedSearchSummary'
       required:
-        - page
-        - pageSize
-        - totalPages
-        - totalResults
         - results
 
-    ApplicationSearchSummaryResponse:
-      title: ApplicationSearchSummaryResponse
-      description: A response containing summary information of an each Anonymous Elector document application
+    AedSearchSummary:
+      title: AedSearchSummary
+      description: An object containing summary information of an each Anonymous Elector document summary
       type: object
       properties:
         gssCode:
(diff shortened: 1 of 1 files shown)


src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1101,6 +1221,102 @@ components:
...
+          description: The date on which the AED was issued
+          example: '2023-01-15'
+        dateTimeCreated:
+          type: string
+          format: date-time
+          description: The timestamp in ISO-8601 UTC that the application document was originally created
+          example: '2022-07-27T09:39:44Z'
+      required:
+        - gssCode
+        - sourceReference
```
vishalgupta-valtech, line 1310: We could also return `source_type` if required. What do you think? Not sure its value though

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -465,6 +465,117 @@ paths:
...
+      summary: Search for Anonymous Elector Document Applications
+      description: 'Search for Anonymous Elector Document Applications based on querying parameters'
+      tags:
+        - Anonymous Elector Document Applications
+      responses:
+        '200':
+          $ref: '#/components/responses/SearchApplications'
+      operationId: search-anonymous-elector-document-applications
+      parameters:
+        - name: pageSize
```
matt-wills-valtech, line 528: On the basis there's a risk we may not include pagination in this release (I think we will, but there's a chance we won't), I'm not sure about including pagination yet as it suggests it's supported.

One option would be to make it clear it's not currently supported in the description and make sure the UI team know that, but I think it might be better/cleaner to exclude it for now (especially if we forget to update the description!).

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -465,6 +465,117 @@ paths:
...
+            minimum: 1
+          example: 1
+          in: query
+          required: true
+        - name: searchBy
+          description: |
+            The data field to apply the `searchValue` to. An empty value will return all AED applications.
+            This value must be provided when `searchValue` is present
+          schema:
+            $ref: '#/components/schemas/SearchBy'
```
matt-wills-valtech, line 550: Nice idea, but this is really for future stories (EIP1-5591 onwards) 👍

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -749,6 +860,15 @@ components:
...
         - not-delivered
 
+    SearchBy:
+      title: SearchBy
+      description: An enumeration of search by fields
+      type: string
+      enum:
+        - surname
+        - applicationReference
+        - electoralRollNumber
```
matt-wills-valtech, line 870: Aside from the fact it might be best to exclude this enum altogether until we get to EIP1-5591, we might also want to include one at a time (i.e. start with just surname).

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -465,6 +465,117 @@ paths:
...
+            This value must be provided when `searchValue` is present
+          schema:
+            $ref: '#/components/schemas/SearchBy'
+          example: surname
+          in: query
+          required: false
+        - name: searchValue
+          description: |
+            The value to search the field specified by the `searchBy` parameter. 
+            The search is a case insensitive match. We do not support fuzzy or wildcard matching.
```
matt-wills-valtech, line 557: 👍

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -749,6 +860,15 @@ components:
         - dispatched
         - not-delivered
 
+    SearchBy:
```
matt-wills-valtech, line 863: This might be a bit too vague/generic (e.g. if another search endpoint is added in the future which uses different search fields), but I'm struggling to think of a better name! 

 `AedSearchByField`? 🤷‍♂️

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1101,6 +1221,102 @@ components:
...
+          items:
+            $ref: '#/components/schemas/ApplicationSearchSummaryResponse'
+      required:
+        - page
+        - pageSize
+        - totalPages
+        - totalResults
+        - results
+
+    ApplicationSearchSummaryResponse:
```
matt-wills-valtech, line 1257: `AedSearchSummaryResponse`?

src/main/resources/openapi/PrintAPIs.yaml
matt-wills-valtech: Personally I don't see the need.

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1101,6 +1221,102 @@ components:
         - generated
         - printed
 
+    ApplicationsSearchResponse:
```
matt-wills-valtech, line 1224: `AedSearchResponse`?

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -465,6 +465,117 @@ paths:
         connectionId: '${vpc_connection_id}'
         httpMethod: POST
 
+
+  #
+  # Anonymous Elector Document Applications
+  # --------------------------------------------------------------------------------
+  '/eros/{eroId}/anonymous-applications':
```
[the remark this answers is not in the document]
matt-wills-valtech: I've since seen the response object at the bottom of this PR and have suggested `AedSearchSummary` as an alternative. Perhaps therefore this endpoint could be `/eros/{eroId}/aed-search-summaries` which would mean we wouldn't need to change `/eros/{eroId}/anonymous-elector-documents`.

However, I'm not particularly enamoured by that idea and it's making me wonder if "summary" is in danger of becoming an overloaded term? (though not as bad as `PENDING`! 😆). I guess the question is what are these response objects and when does an object stop being the "full" resource and start being a "summary"? 

I'm not saying any of this is easy btw! 😆

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -465,6 +465,117 @@ paths:
...
+          type: string
+        example: 'city-council'
+        in: path
+        required: true
+    options:
+      summary: CORS support
+      description: |
+        Enable CORS by returning correct headers
+      tags:
+        - Anonymous Elector Document Applications
```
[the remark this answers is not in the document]
vishalgupta-valtech: It was deliberate attempt to not put it under the previous tag

src/main/resources/openapi/PrintAPIs.yaml
vishalgupta-valtech: Ok I can remove pagination

src/main/resources/openapi/PrintAPIs.yaml
vishalgupta-valtech: I think keeping these enum now is fine when it's evident that these will be surely supported later on. If ui tries to search by electoral roll and backend hasnt implemented it then it will just return empty list. 

But if i am not too precious about it.

src/main/resources/openapi/PrintAPIs.yaml
vishalgupta-valtech: These are optional fields and ui can just choose to ignore them until we have implemented these in api

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1101,6 +1221,102 @@ components:
...
+          description: The number of results per page
+          example: 1
+        totalPages:
+          type: integer
+          description: The total number of pages
+          example: 1
+        totalResults:
+          type: integer
+          description: The total number of results
+          example: 1
```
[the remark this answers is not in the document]
vishalgupta-valtech: Yes agree. I will remove Pagination related request/response fields

src/main/resources/openapi/PrintAPIs.yaml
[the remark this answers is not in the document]
vishalgupta-valtech: For the same reason that you mentioned with existing get anonymous-elector-documents endpoint , i named this to anonymous-applications. This name is similar to how we search for anonymous-applications within vca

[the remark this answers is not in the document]
vishalgupta-valtech: Could rename exisitng get summaries endpoint to be "history" or something along those lines if UI isnt consuming those.

The current pr is more of "anonymous-elector-documents" then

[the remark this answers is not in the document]
matt-wills-valtech: It's just occurred to me that the existing GET endpoint could have `/{applicationId}` in the path rather than the body, which would make sense conceptually.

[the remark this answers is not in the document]
vishalgupta-valtech: Endpoint and tag renamed to `...../anonymous-elector-document-summaries` @matt-wills-valtech

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -465,6 +465,100 @@ paths:
         connectionId: '${vpc_connection_id}'
         httpMethod: POST
 
+
+  #
+  # Anonymous Elector Document Summaries
+  # --------------------------------------------------------------------------------
+  '/eros/{eroId}/anonymous-elector-document-summaries':
```
NathanRussellValtech, line 472: The problem with `anonymous-elector-document-summaries` is that it's not clear how it is different to the existing `anonymous-elector-documents` endpoint, especially considering the existing endpoint is document as returning "summaries"
I think we need to step back slightly and understand the terminology, resources, use cases etc; otherwise we are at risk of introducing at best a hard to grok API, and at worst a [Big Ball Of Mud](https://www.infoq.com/news/2014/08/microservices_ballmud/) :(

We need to slow down to go faster!!! (sorry, falling into Tim-style management speak!!! :D )

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -749,6 +843,13 @@ components:
         - dispatched
         - not-delivered
 
+    AedSearchBy:
+      title: AedSearchBy
+      description: An enumeration of search by fields for searching AEDs
+      type: string
+      enum:
+        - surname
```
NathanRussellValtech, line 851: I've not checked, but is this the same enum value name/terminology that we use in VAC?
I know they are different search APIs, but we need to be consistent. If it is `surname` in both APIs, then happy days 😁 
But if VCA already uses some other term such as `lastName` then we should use it here too. We just need to check we are being consistent

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1101,6 +1202,82 @@ components:
...
+          type: string
+          pattern: '^[a-fA-F\d]{24}$'
+          description: |
+            A 24 character hex string. 
+            The unique identifier of an Anonymous elector document application.
+          example: 1f0f76a9a66f438b9bb33070
+        applicationReference:
+          type: string
+          description: The reference number of an Anonymous elector document application
+          example: A3JSZC4CRH
```
NathanRussellValtech, line 1237: Minor, but the example should be a "V" number, because all the application references always start with "V"

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1101,6 +1202,82 @@ components:
...
+          example: Fred
+        surname:
+          type: string
+          description: The applicant's surname
+          example: Blogs
+        postcode:
+          type: string
+          description: The applicant's postcode
+          example: AA1 1AA
+        issueDate:
```
NathanRussellValtech, line 1259: I know the UI wants to render this as "Printed Date" (issue == printed), but I'm wondering about the date aspect of it ... 🤔 
Presumably we actually hold it in the DB as a date time ? Should we be returning it as a date time (which will make sorting easier as well), and then the UI can decide for themselves how they want to render the value

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1101,6 +1202,82 @@ components:
...
+          type: string
+          format: date
+          description: The date on which the AED was issued
+          example: '2023-01-15'
+        dateTimeCreated:
```
NathanRussellValtech, line 1264: Whats the difference between when the Anon Elector **Document** was issued vs. created?

matt-wills-valtech: See above.

src/main/resources/openapi/PrintAPIs.yaml
vishalgupta-valtech: yes, from vca,

image

src/main/resources/openapi/PrintAPIs.yaml
vishalgupta-valtech: Copy pasted from vca-api spec, will change it to start with `V` :)

src/main/resources/openapi/PrintAPIs.yaml
NathanRussellValtech: Perfect ; thanks for checking 👍

matt-wills-valtech (approved):
LGTM. Thanks for bearing with us and making the changes 👍

NathanRussellValtech (approved):
LGTM, but wait for a ✅ from Matt before merging, as its likely I will have missed something 😁 👍

sentioservices (approved):
LGTM

PR #231, opened 2023-04-24
EIP1-5615 / EIP1-5627 - openAPI spec to add re-issue endpoint

NathanRussellValtech:
This PR adds the new endpoint definition for the new "re-issue AED" API.

Screenshot 2023-04-24 at 09 09 37

### DIFF
--- a/src/main/resources/openapi/PrintAPIs.yaml
+++ b/src/main/resources/openapi/PrintAPIs.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Print APIs
-  version: '1.13.0'
+  version: '1.14.0'
   description: Print APIs
   contact:
     name: Krister Bone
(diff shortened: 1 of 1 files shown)


src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1359,6 +1443,30 @@ components:
         - registeredAddress
         - delivery
 
+    ReIssueAnonymousElectorDocumentRequest:
+      title: ReIssueAnonymousElectorDocumentRequest
+      description: Request body to re-issue an Anonymous Elector Document (AED)
+      type: object
+      properties:
+        sourceType:
+          $ref: '#/components/schemas/SourceType'
```
NathanRussellValtech, line 1452: I'm not sure about `sourceType` (it's an enum of `voter-card` or `anonymous-elector-document`) ... 🤔 

Half of me says we don't need it as it's implied through the API path .... but there is a semi-precedent set where we already use it in `GenerateAnonymousElectorDocumentRequest` and `GenerateTemporaryCertificateRequest` 🤔 

What do we think? I guess the possible difference here is that the REST APIs that use `GenerateAnonymousElectorDocumentRequest` and `GenerateTemporaryCertificateRequest` are meant to be called by VCA, rather than by the UI, so we can be confident that VCA sets the correct value
Whereas this API is meant to be called by the UI .....

Thoughts anyone ?

vishalgupta-valtech: I agree. I don't think we need `sourceType` at all for generating AEDs. It's implicitly `ANONYMOUS_ELECTOR_DOCUMENT`

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -676,6 +676,90 @@ paths:
...
+        '400':
+          $ref: '#/components/responses/400ErrorResponse'
+        '404':
+          $ref: '#/components/responses/404ErrorResponse'
+      operationId: re-issue-anonymous-elector-document
+      security:
+        - eroUserCognitoUserPoolAuthorizer: [ ]
+      x-amazon-apigateway-integration:
+        type: HTTP_PROXY
+        uri: '${base_uri}/eros/{eroId}/anonymous-elector-documents'
```
vishalgupta-valtech, line 752: Missing `/re-issue`

vishalgupta-valtech (approved):
LGTM, just minor comment

src/main/resources/openapi/PrintAPIs.yaml
NathanRussellValtech: Good spot; done 👍

src/main/resources/openapi/PrintAPIs.yaml
NathanRussellValtech: Done

PR #232, opened 2023-04-24
Feature/EIP1-5628 EIP1-5662 Additional fields in AED response

matt-wills-valtech:
This PR adds additional fields to the response object that's returned from the Get AEDs by applicationId endpoint.

Since each object in the response now contains nearly all the fields for an AED, I've removed any mention of "summary" and instead we should think of them as AEDs in their own right.

The additional fields are:
* sourceReference
* applicationReference
* firstName
* middleNames
* surname
* email
* phoneNumber

image


src/main/kotlin/uk/gov/dluhc/printapi/database/entity/AnonymousElectorDocument.kt
```diff
@@ -51,8 +51,9 @@ class AnonymousElectorDocument(
     @field:Size(max = 255)
     var sourceReference: String,
 
+    @field:NotNull
     @field:Size(max = 255)
-    var applicationReference: String?,
+    var applicationReference: String,
```
matt-wills-valtech, line 56: This is mandatory in the generate endpoint within the swagger spec, but just to be sure we've checked that there aren't any null instances in prod.

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1229,6 +1229,17 @@ components:
...
           description: GSS code of the ERO responsible for the application
           example: E12345678
+        sourceReference:
+          type: string
+          pattern: '^[a-fA-F\d]{24}$'
+          description: |
+            A 24 character hex string. 
+            The unique identifier of an Anonymous elector document application.
+          example: 1f0f76a9a66f438b9bb33070
+        applicationReference:
```
vishalgupta-valtech, line 1239: `sourceReference` and `applicationReference` should be added to `required:` as well

vishalgupta-valtech (approved):
LGTM, just a minor comment around spec

src/main/kotlin/uk/gov/dluhc/printapi/service/aed/AnonymousElectorDocumentService.kt
```diff
@@ -28,25 +28,25 @@ class AnonymousElectorDocumentService(
...
+        with(generateAnonymousElectorDocumentMapper.toAnonymousElectorDocument(dto, filename)) {
             val templateDetails = pdfTemplateDetailsFactory.getTemplateDetails(this)
             val contents = pdfFactory.createPdfContents(templateDetails)
             anonymousElectorDocumentRepository.save(this)
             return PdfFile("anonymous-elector-document-$certificateNumber.pdf", contents)
         }
     }
 
-    fun getAnonymousElectorDocumentSummaries(
+    fun getAnonymousElectorDocuments(
```
vishalgupta-valtech, line 39: Sorry I think I should have done it earlier, but are you ok to add this bit ? 

@Transactional(readOnly = true)

matt-wills-valtech: Good spot, done.

PR #233, opened 2023-04-24
NO-JIRA: Adds dummy endpoint implementations

vishalgupta-valtech:


src/main/kotlin/uk/gov/dluhc/printapi/rest/aed/AnonymousElectorDocumentController.kt
```diff
@@ -60,6 +65,26 @@ class AnonymousElectorDocumentController(
...
     }
 
+    @PostMapping("re-issue")
+    @PreAuthorize(HAS_ERO_VC_ANONYMOUS_ADMIN_AUTHORITY)
+    fun reissueAnonymousElectorDocument(
+        @PathVariable eroId: String,
+        @RequestBody @Valid request: ReIssueAnonymousElectorDocumentRequest,
+        authentication: Authentication
+    ): ResponseEntity<InputStreamResource> {
+        logger.info { "Request received to reissue AED document for eroId [$eroId], request [$request]" }
```
vishalgupta-valtech, line 75: Upon implementating this endpoint, we should change this 2 info statements to be debug

PR #237, opened 2023-04-25
EIP1-5675 - Add collection reason column

matt-wills-valtech:
New optional column which will be populated in subsequent PRs. Also known as "problem reason" in VCA, but "collection reason" is surely better.


src/main/kotlin/uk/gov/dluhc/printapi/database/entity/Delivery.kt
```diff
@@ -48,6 +48,9 @@ class Delivery(
     @Enumerated(EnumType.STRING)
     var deliveryAddressType: DeliveryAddressType,
 
+    @field:Size(max = 1024)
+    var collectionReason: String?,
```
vishalgupta-valtech, line 52: Should it be `= null`

vishalgupta-valtech (approved):
LGTM 👍🏼 I think we need to acknowledge that we will lose the `reason` values for the existing records

PR #238, opened 2023-04-25
Feature/EIP1-5685 Persist collection reason

matt-wills-valtech:
This PR persists the reason the elector gave for having to collect their AED (if applicable).


vishalgupta-valtech (approved):
Lgtm

PR #240, opened 2023-04-25
EIP1-5590: EIP1-5681: Implement search AED service

vishalgupta-valtech:


src/main/kotlin/uk/gov/dluhc/printapi/service/aed/AnonymousElectorDocumentSearchService.kt
```diff
@@ -0,0 +1,34 @@
+package uk.gov.dluhc.printapi.service.aed
+
+import org.springframework.data.domain.Page
```
NathanRussellValtech, line 3: Hmmm .... I'm not sure about the service class knowing (and exposing via the method return type) JPA / DB related stuff .... feels like the wrong coupling ?

What do we do in VCA in this respect?

vishalgupta-valtech: This will eventually be passed on from Controller when we start doing pagination properly. So everything from service will be removed/moved out to mapper etc. In VCA the endpoint query params supplies those values and passes them onto repository. (Also we use specification builder which I don't think we need in this case here and it's a simple search queries) .. I can add a TODO to say remove or move this to mapper when endpoint supports pagination etc.

vishalgupta-valtech: .. or I can pass in `SearchQueryStringParameters` in service which will be populated by Controller?

src/main/kotlin/uk/gov/dluhc/printapi/service/aed/AnonymousElectorDocumentSearchService.kt
```diff
@@ -0,0 +1,34 @@
...
+import uk.gov.dluhc.printapi.service.EroService
+
+@Service
+class AnonymousElectorDocumentSearchService(
+    private val eroService: EroService,
+    private val anonymousElectorDocumentSummaryRepository: AnonymousElectorDocumentSummaryRepository,
+    private val anonymousSearchSummaryMapper: AnonymousSearchSummaryMapper,
+) {
+
+    fun searchAnonymousElectorDocumentSummaries(eroId: String): Page<AnonymousSearchSummaryDto> {
```
NathanRussellValtech, line 22: Re: my comment above, this return type presumably means the controller needs to know what to do with a `org.springframework.data.domain.Page` ?? That doesnt feel right?

NathanRussellValtech: Really? bloomin' eck! I wish we hadn't have done that in VCA 👎 
Never mind, that ship has sailed ... 😢

matt-wills-valtech: Is that right @vishalgupta-valtech? The controller in VCA doesn't know about the `org.springframework.data.domain` package - we use integers to represent pages and pageSize etc.

The service class needs to import `Page` and `PageRequest`, but not the controller.

matt-wills-valtech: Actually you're right but it doesn't explicitly need to import Page, hence my initial confusion. mmm...that's not ideal.

vishalgupta-valtech: VCA Controller 
image

NathanRussellValtech: Yes, I guess this is perhaps one of the bad things about kotlin (😬 , did I really say that? 🤣 )
Where kotlin can infer the type we dont need to have an explicit import, so this level of coupling gets somewhat lost

🤔

matt-wills-valtech: Indeed, I was a bit quick to only examine the imports! 😆

vishalgupta-valtech: @NathanRussellValtech @matt-wills-valtech As suggested, pushed a code that introduces `AnonymousSearchSummaryPageDto` that now will handle all values from pagedResponse within service

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/testdata/entity/AedSummariesPageRequestBuilder.kt
```diff
@@ -0,0 +1,14 @@
+package uk.gov.dluhc.printapi.testsupport.testdata.entity
+
+import org.springframework.data.domain.PageRequest
+import org.springframework.data.domain.Sort
+import org.springframework.data.domain.Sort.Direction
+
+fun withPageRequestAndSortOrder(
```
matt-wills-valtech, line 7: Minor - `buildPageRequest()`?

Not sure if you need to mention the sort order.

matt-wills-valtech (approved):
LGTM. Thanks for addressing our comments 👍

NathanRussellValtech (approved):
LGTM

PR #242, opened 2023-04-26
EIP1-5615 / EIP1-5658 - Added mappings for electoralRollNumber, userId, deliveryAddressType

NathanRussellValtech:
This PR adds in the mappings for `electoralRollNumber`, `userId` and `deliveryAddressType` when re-issuing an AED
(because I forgot about them in my earlier PR! 🙄 )

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/mapper/aed/ReIssueAnonymousElectorDocumentMapper.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/mapper/aed/ReIssueAnonymousElectorDocumentMapper.kt
@@ -1,7 +1,9 @@
 package uk.gov.dluhc.printapi.mapper.aed
 
+import org.mapstruct.AfterMapping
 import org.mapstruct.Mapper
 import org.mapstruct.Mapping
+import org.mapstruct.MappingTarget
 import org.springframework.beans.factory.annotation.Autowired
 import uk.gov.dluhc.printapi.database.entity.AnonymousElectorDocument
 import uk.gov.dluhc.printapi.database.entity.AnonymousElectorDocumentStatus
@@ -26,6 +28,9 @@ abstract class ReIssueAnonymousElectorDocumentMapper {
     @Autowired
     protected lateinit var aedMappingHelper: AedMappingHelper
 
+    @Autowired
+    protected lateinit var deliveryAddressTypeMapper: DeliveryAddressTypeMapper
+
     abstract fun toReIssueAnonymousElectorDocumentDto(
         apiRequest: ReIssueAnonymousElectorDocumentRequest,
         userId: String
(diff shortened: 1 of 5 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/mapper/aed/ReIssueAnonymousElectorDocumentMapper.kt
```diff
@@ -35,13 +40,25 @@ abstract class ReIssueAnonymousElectorDocumentMapper {
...
     @Mapping(target = "delivery.address.id", ignore = true)
     @Mapping(target = "contactDetails.id", ignore = true)
     @Mapping(target = "contactDetails.address.id", ignore = true)
     abstract fun toNewAnonymousElectorDocument(
         previousAed: AnonymousElectorDocument,
+        dto: ReIssueAnonymousElectorDocumentDto,
         aedTemplateFilename: String
     ): AnonymousElectorDocument
+
+    @AfterMapping
```
NathanRussellValtech, line 57: Don't beat me up for this! It's the only way I could get it to work!

If I put `@Mapping(target = "delivery.deliveryAddressType", source = "dto.deliveryAddressType")` on the main mapping method, then the resultant `Delivery` object had all its fields null except `deliveryAddressType` !!!
It seems you have to let MapStruct map it from the previous AED, then mutate it afterwards! ¯\_(ツ)_/¯

vishalgupta-valtech (approved):
LGTM

src/test/kotlin/uk/gov/dluhc/printapi/mapper/aed/ReIssueAnonymousElectorDocumentMapperTest.kt
```diff
@@ -101,15 +107,29 @@ class ReIssueAnonymousElectorDocumentMapperTest {
...
+
+        given(deliveryAddressTypeMapper.mapDtoToEntity(any())).willReturn(DeliveryAddressType.ERO_COLLECTION)
+
         val expected = previousAed.deepCopy().apply {
             certificateNumber = newCertificateNumber
             requestDateTime = FIXED_TIME
             issueDate = FIXED_DATE
             statusHistory = expectedStatusHistory.toMutableList()
+            electoralRollNumber = newElectoralRollNumber
+            delivery?.deliveryAddressType = DeliveryAddressType.ERO_COLLECTION
```
matt-wills-valtech, line 127: How about checking that `delivery?.collectionReason` is also copied (which would be apt given the `DeliveryAddressType` here)?

vishalgupta-valtech: I had similar question, but then I checked the code and found out that the mapper was copying entire delivery from previousAed, so looks good.

NathanRussellValtech: OK , will do .... 👍

PR #243, opened 2023-04-26
Feature/EIP1-5687 Return collection reason

matt-wills-valtech:


NathanRussellValtech (approved):
LGTM 👍

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1250,6 +1250,11 @@ components:
           $ref: '#/components/schemas/AnonymousSupportingInformationFormat'
         deliveryAddressType:
           $ref: '#/components/schemas/DeliveryAddressType'
+        collectionReason:
+          type: string
+          description: |
+            The reason the elector gave for having to collect their voter ID (if applicable).
```
vishalgupta-valtech, line 1256: Minor:  `.... to collect their AED (if applicable)`? Although AED is a voter ID

vishalgupta-valtech (approved):

src/main/resources/openapi/PrintAPIs.yaml
vishalgupta-valtech: Also worth updating api spec version

matt-wills-valtech: Updated. I copied it from another collection reason this spec, which is for AEDs and VACs, but this one isn't 😄

PR #246, opened 2023-04-27
EIP1-4998 - Tidied application properties and readme

NathanRussellValtech:


src/main/resources/application-dev.yml
```diff
@@ -1,10 +1,5 @@
 jobs:
-  process-print-responses:
```
NathanRussellValtech, line 2: These overrides for `dev` are now in `infra` (where they belong! 😁 )

src/test/kotlin/uk/gov/dluhc/printapi/config/LocalStackContainerConfiguration.kt
```diff
@@ -138,7 +138,7 @@ class LocalStackContainerConfiguration {
      * @return a [LocalStackContainerSettings] bean encapsulating the various IDs etc of the configured container and services.
      */
     @Bean
-    fun localStackContainerSqsSettings(
```
NathanRussellValtech, line 141: It's not just SQS related .... (it had the wrong name to start with)

src/main/resources/application-dev.yml
vishalgupta-valtech: QQ: Why isn't the environment vars e.g; `JOBS_BATCH_PRINT_REQUESTS_CRON` and others used within main `application.yml` ?

vishalgupta-valtech:
infra PR https://github.com/cabinetoffice/eip-ero-infra/pull/814/ with newly introduced vars

matt-wills-valtech (commented):
LGTM 👍

src/main/resources/application-dev.yml
vishalgupta-valtech: ok thank you, so the intention is that even in infra, they will never be overridden for prod (without a manual release to change these default cron schedules as per main `application.yml`?)

vishalgupta-valtech (approved):
LGTM

src/main/resources/application-dev.yml
NathanRussellValtech: Not quite.
All AWS deployments (dev, dev2, int, int2, test, tst, live , etc etc) will get these env vars set via `infra`. So in that respect all of our AWS deployments are "overriding" the values in `application.yml`
The default values in `infra` are the same as the default values in `application.yml`:
```
vars.tf

variable "print_api_batch_print_requests_cron" {
  type        = string
  description = "The cron schedule for when the print-api batches print requests to send to the Print Provider"
  default     = "0 0/15 * * * *" # every 15 minutes, starting on the hour - see analysis and recommendations in EIP1-2515"
}

variable "print_api_process_print_responses_cron" {
  type        = string
  description = "The cron schedule for when the print-api polls the Print Provider's OutBound folder to find and process print responses"
  default     = "0 15/30 * * * *" # every 30 minutes starting at 15 minutes past the hour - see analysis and recommendations in EIP1-2515
}
```

Then, specific environments in `infra` can change these values if they like. We do this for dev and dev2 for example:
```
dev2.tfvars

# print-api cron settings
print_api_batch_print_requests_cron = "0 */10 * ? * *" # every 10 minutes
print_api_process_print_responses_cron = "0 */10 * ? * *" # every 10 minutes
```

We are not currently overriding these `infra` variables in `live.tfvars` which means `live` will be deployed with the default cron values by virtue of `vars.tf`. But we could set different values in `live.tfvars` if we needed to, which would be an `infra` deploy, not a change to the application.

Does that answer the question?

PR #250, opened 2023-04-27
EIP1-5615 - Fix mapping of new AED entity

NathanRussellValtech:
This PR fixes the mapping of new AED entities to not copy JPA managed fields from the previous entity


matt-wills-valtech (approved):
LGTM 🚀

PR #251, opened 2023-04-27
EIP1-5616 / EIP1-5761 - Add updatedDate and updatedBy to AED Contact Details

NathanRussellValtech:
This PR adds updatedDate and updatedBy to the AED Contact Details table.

This is to support EIP1-5616 which is to allow the ERO to update the email and/or phoneNumber of an AED (which is the AED Contact Details table).
Without this change it would mean we would lose (overwrite) the createdDate and createdBy fields, thereby losing the history of change

The `print-api` database does not use envers audit tables as it was never intended that database records would ever be updated (by an ERO)
This is a better/faster/pragmatic solution over and above introducing envers at this stage.

### DIFF
--- a/src/main/resources/db/changelog/ddl/0028_alter_aed_contact_details_add_last_updated.xml
+++ b/src/main/resources/db/changelog/ddl/0028_alter_aed_contact_details_add_last_updated.xml
@@ -1,0 +1,32 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+
+    <changeSet author="nathan.russell@valtech.com" id="0028_EIP-5761_aed_contact_details add updated_by and date_updated" context="ddl">
+
+        <addColumn tableName="aed_contact_details">
+            <column name="date_updated" type="timestamp" beforeColumn="version">
+                <constraints nullable="true"/>
+            </column>
+            <column name="updated_by" type="varchar(255)" beforeColumn="version">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+
+        <sql>
+            UPDATE aed_contact_details SET date_updated = date_created;
+            UPDATE aed_contact_details SET updated_by = created_by;
+        </sql>
+
+        <addNotNullConstraint tableName="aed_contact_details" columnName="date_updated" columnDataType="timestamp"/>
+        <addNotNullConstraint tableName="aed_contact_details" columnName="updated_by" columnDataType="varchar(255)"/>
+
+        <rollback>
+            <dropColumn tableName="aed_contact_details" columnName="date_updated"/>
+            <dropColumn tableName="aed_contact_details" columnName="updated_by"/>
+        </rollback>
+    </changeSet>
+</databaseChangeLog>
--- a/src/main/resources/db/changelog/ddl/0028_alter_aed_contact_details_add_last_updated.xml
+++ b/src/main/resources/db/changelog/ddl/0028_alter_aed_contact_details_add_last_updated.xml
@@ -8,10 +8,10 @@
     <changeSet author="nathan.russell@valtech.com" id="0028_EIP-5761_aed_contact_details add updated_by and date_updated" context="ddl">
 
         <addColumn tableName="aed_contact_details">
-            <column name="date_updated" type="timestamp" beforeColumn="version">
+            <column name="date_updated" type="timestamp" afterColumn="created_by">
                 <constraints nullable="true"/>
             </column>
-            <column name="updated_by" type="varchar(255)" beforeColumn="version">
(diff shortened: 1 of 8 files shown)


matt-wills-valtech (approved):

src/main/resources/db/changelog/ddl/0028_alter_aed_contact_details_add_last_updated.xml
```diff
@@ -0,0 +1,32 @@
...
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+
+    <changeSet author="nathan.russell@valtech.com" id="0028_EIP-5761_aed_contact_details add updated_by and date_updated" context="ddl">
+
+        <addColumn tableName="aed_contact_details">
+            <column name="date_updated" type="timestamp" beforeColumn="version">
```
vishalgupta-valtech, line 11: For some reason, I don't think `beforeColumn` works in our liquibase scripts, so you will need to use `afterColumn`. 
If you can run this locally and prove that column is indeed added before version, then that will be appreciated.

src/main/resources/db/changelog/ddl/0028_alter_aed_contact_details_add_last_updated.xml
```diff
@@ -0,0 +1,32 @@
...
+                <constraints nullable="true"/>
+            </column>
+            <column name="updated_by" type="varchar(255)" beforeColumn="version">
+                <constraints nullable="true"/>
+            </column>
+        </addColumn>
+
+        <sql>
+            UPDATE aed_contact_details SET date_updated = date_created;
```
vishalgupta-valtech, line 20: I am not 100% sure but If you would have used this then it copies the value from `created_by` column. Our DBA expert @sentioservices would have known more.

`<column name="updated_by" type="varchar(255)" value="created_by">` 

But don't make this change, (not a blocker)

NathanRussellValtech: I dont think thats how the `value` attribute works - I think it means it will set the value from the string literal value `created_by` (in your example), rather than "take the value from this column name"

See the liquibase docs here - but of interest there is a `value` attribute (which I think means "set this string literal value"), and also `valueBoolean`, `valueDate`, `valueNumeric` etc ... which further support my thinking/understanding of how this works.

I based by approach on this answer: https://stackoverflow.com/questions/35172172/how-to-add-new-column-with-default-value-from-existing-column-in-liquibase

src/main/resources/db/changelog/ddl/0028_alter_aed_contact_details_add_last_updated.xml
NathanRussellValtech: You are right; `beforeColumn` does not work. I'll try `afterColumn` .... (but I'm not going to lose any sleep over it - I don't think it _really_ matters in what order the columns appear 👍 )

NathanRussellValtech: OK, I've changed it to `afterColumn`, which results in this 👍 
Screenshot 2023-04-27 at 19 02 50

src/main/resources/db/changelog/ddl/0028_alter_aed_contact_details_add_last_updated.xml
NathanRussellValtech: Also worth noting that the approach I've taken (create a nullable column, SQL to set its value, add non-nullable constraint) is exactly what Neil did in `0015_refactor_anonymous_elector_document_tables.xml`
```
    <changeSet author="neil.massey@valtech.com" id="0015-EIP1-4179-refactor-anonymous-elector-document-table"
               context="ddl">
        <addColumn tableName="anonymous_elector_document">
            <column name="electoral_roll_number" type="varchar(7)">
                <constraints nullable="true"/>
            </column>
        </addColumn>

        <addColumn tableName="anonymous_elector_document">
            <column name="aed_template_filename" type="varchar(255)">
                <constraints nullable="true"/>
            </column>
        </addColumn>

....

        <sql>
            <comment>Copy AED details from Print Request</comment>
            UPDATE anonymous_elector_document, aed_print_request
            SET anonymous_elector_document.electoral_roll_number = aed_print_request.electoral_roll_number, anonymous_elector_document.aed_template_filename = aed_print_request.aed_template_filename, 
....
            WHERE anonymous_elector_document.id = aed_print_request.aed_id;
        </sql>

....

        <addNotNullConstraint tableName="anonymous_elector_document" columnName="electoral_roll_number"
                              columnDataType="varchar(7)"/>
        <addNotNullConstraint tableName="anonymous_elector_document" columnName="aed_template_filename"
                              columnDataType="varchar(255)"/>

```

vishalgupta-valtech: Yep your approach is absolutely fine no doubts, i just thought if there was a clever way of just prepopulating new columns with existing column values. All is good 👍🏼

vishalgupta-valtech (approved):
Lgtm and thanks for addressing comments 👍🏼

PR #252, opened 2023-04-28
EIP1-4998 - Application ymls for new environments

NathanRussellValtech:


matt-wills-valtech (approved):
👍

src/main/resources/application-tst2.yml
```diff
@@ -0,0 +1,16 @@
+
+email:
```
vishalgupta-valtech, line 2: Should this file be `test2.yml`

PR #253, opened 2023-04-28
No jira/rename test2 yml

vishalgupta-valtech:


NathanRussellValtech (commented):
No; this is wrong.
The file should be called `application-tst2.yml`

PR #255, opened 2023-04-28
EIP1-4998 - Add logging for email sending

NathanRussellValtech:
Adds logging to understand who emails are being sent to

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/emailnotifications/SesEmailClient.kt
+++ b/src/main/kotlin/uk/gov/dluhc/emailnotifications/SesEmailClient.kt
@@ -22,6 +22,7 @@ class SesEmailClient(
         val filteredEmailToAddresses = filterRecipientsIfAllowListConfigured(emailToRecipients, subject)
         val filteredEmailCcAddresses = filterRecipientsIfAllowListConfigured(emailCcRecipients, subject)
         if (filteredEmailToAddresses.isEmpty() && filteredEmailCcAddresses.isEmpty()) {
+            logger.warn { "Both toAddresses and ccAddresses have been filtered down to empty sets. No recipients to send to." }
             return emptySet()
         }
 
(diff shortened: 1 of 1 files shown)


src/test/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintResponseMessageListenerIntegrationTest.kt
```diff
@@ -115,7 +115,9 @@ internal class ProcessPrintResponseMessageListenerIntegrationTest : IntegrationT
         val expectedEmailMessage = buildLocalstackEmailMessage(
             timestamp = OffsetDateTime.now(UTC).toLocalDateTime().truncatedTo(SECONDS),
             emailSender = "noreply_erouser@erop.ierds.uk",
-            toAddresses = setOf("a-user@valtech.com"),
+            // TODO - remove this as soon as EIP1-4998 and EIP1-5090 have been tested
+            toAddresses = setOf("joshua.ayoade@valtech.com"),
```
vishalgupta-valtech, line 119: I suggest we also temporarily change log.debug to log.info or add few log info before and after email is actually sent

vishalgupta-valtech (approved):
Looks hacky to me 👍🏼

src/main/kotlin/uk/gov/dluhc/emailnotifications/SesEmailClient.kt
```diff
@@ -22,6 +22,7 @@ class SesEmailClient(
         val filteredEmailToAddresses = filterRecipientsIfAllowListConfigured(emailToRecipients, subject)
         val filteredEmailCcAddresses = filterRecipientsIfAllowListConfigured(emailCcRecipients, subject)
         if (filteredEmailToAddresses.isEmpty() && filteredEmailCcAddresses.isEmpty()) {
+            logger.warn { "Both toAddresses and ccAddresses have been filtered down to empty sets. No recipients to send to." }
             return emptySet()
```
NathanRussellValtech, line 26: I do wonder why this returns "successfully", rather than throwing an exception (which the other fail cases do .... but I'm not going to change that)

matt-wills-valtech (approved):

src/main/kotlin/uk/gov/dluhc/emailnotifications/SesEmailClient.kt
vishalgupta-valtech: I think it's for a scenario If the recipients are not from `allowedList` then after filtering to/cc, these could be empty, but I agree with you. I remember I raised too in PR that this could return an exception but don't remember the real reason why it just returns an empty set

PR #262, opened 2023-05-02
EIP1-5838 - Service to create a pre-signed URL to retrieve an AED photo

matt-wills-valtech:
This PR completes the endpoint to generate a pre-signed URL to an AED's photo in S3.

I'm not completely sure about the path to retrieve the URL (personally I'm leaning towards `/eros/{eroId}/anonymous-elector-documents/{certificateNumber}/photo`), but this is something @NathanRussellValtech and I have agreed to put aside for the time being and first make sure this works.

### DIFF
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -29,8 +29,6 @@ sqs:
   remove-certificate-queue-name: ${SQS_REMOVE_CERTIFICATE_QUEUE_NAME}
 
 s3:
-  # TODO - EIP1-5839 - infra PR (including S3 permissions)
-  # TODO - add properties to README
   certificate-photo-access-duration: 1M
   certificate-photos-target-bucket: ${S3_CERTIFICATE_PHOTOS_TARGET_BUCKET_NAME}
   certificate-photos-target-bucket-proxy-endpoint: ${S3_CERTIFICATE_PHOTOS_TARGET_BUCKET_PROXY_ENDPOINT}
--- a/src/main/resources/application.yml
+++ b/src/main/resources/application.yml
@@ -28,6 +28,13 @@ sqs:
   application-removed-queue-name: ${SQS_APPLICATION_REMOVED_QUEUE_NAME}
   remove-certificate-queue-name: ${SQS_REMOVE_CERTIFICATE_QUEUE_NAME}
 
+s3:
+  # TODO - EIP1-5839 - infra PR (including S3 permissions)
+  # TODO - add properties to README
+  certificate-photo-access-duration: 1M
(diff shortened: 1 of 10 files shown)


src/main/resources/application.yml
```diff
@@ -28,6 +28,11 @@ sqs:
   application-removed-queue-name: ${SQS_APPLICATION_REMOVED_QUEUE_NAME}
   remove-certificate-queue-name: ${SQS_REMOVE_CERTIFICATE_QUEUE_NAME}
 
+s3:
+  certificate-photo-access-duration: 1M
+  certificate-photos-target-bucket: ${S3_CERTIFICATE_PHOTOS_TARGET_BUCKET_NAME}
+  certificate-photos-target-bucket-proxy-endpoint: ${S3_CERTIFICATE_PHOTOS_TARGET_BUCKET_PROXY_ENDPOINT}
```
matt-wills-valtech, line 34: EIP1-5839 for the infra related changes.

src/test/kotlin/uk/gov/dluhc/printapi/service/S3PhotoServiceTest.kt
```diff
@@ -0,0 +1,88 @@
...
+        val expectedUri = URI.create(transformedUrl)
+
+        // When
+        val location = s3PhotoService.generatePresignedGetCertificatePhotoUrl(s3Arn)
+
+        // Then
+        assertThat(location).isEqualTo(expectedUri)
+        verify(s3Presigner).presignGetObject(expectedPresignRequest)
+    }
+}
```
ReedSoftware, line 88: It would be good to add the unit test that's in VCA named `should generate photo URL given S3Arn containing all symbols one might normally find on a keyboard, and ensures they are encoded` which ensures that stuff is all encoded properly in a filename

NathanRussellValtech (approved):
LGTM, especially if we can add the test or similar that James has mentioned 👍

vishalgupta-valtech (approved):
LGTM. So now you a S3 expert too 😉 Good stuff 😬

NathanRussellValtech:
> LGTM. So now you a S3 expert too 😉 Good stuff 😬

It's not good stuff ; it's "Great Stuff" !!! 🤣

src/test/kotlin/uk/gov/dluhc/printapi/service/S3PhotoServiceTest.kt
matt-wills-valtech: I've addressed this in the integration test, rather than the unit test, since I'm mocking some of the S3 classes in the latter.

PR #267, opened 2023-05-04
EIP1-5853: Api spec update to include pagination while searching AEDs

vishalgupta-valtech:
image

### DIFF
--- a/src/main/resources/openapi/PrintAPIs.yaml
+++ b/src/main/resources/openapi/PrintAPIs.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Print APIs
-  version: '1.17.0'
+  version: '1.18.0'
   description: Print APIs
   contact:
     name: Krister Bone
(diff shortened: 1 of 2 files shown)


src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -525,6 +525,23 @@ paths:
...
+        - name: page
+          description: The page number being requested (with the first page being 1). Defaulted to 1 if not supplied
+          schema:
+            type: integer
+            minimum: 1
+          example: 1
+          in: query
+          required: true
+        - name: pageSize
+          description: The number of results to return in each page. Defaulted to 100 if not supplied
```
matt-wills-valtech, line 537: Is this on the assumption the UI is going to be ready?

vishalgupta-valtech: Even if UI doesn't supply page or pageSize, the controller defaults them to 1 and 100, so shouldn't break UI.

vishalgupta-valtech: image

matt-wills-valtech: What if there are > 100 results? I thought we were going to set it to something high, like 1000 to start wtih?

vishalgupta-valtech: Yes sure, Could change that in next subtasks.

vishalgupta-valtech: As per great @matt-wills-valtech there are only around 3570 anonymous electors, so chances of AEDs among some ero's applications going above are going to be less than 100, except one of two EROS which has more AEDs. 😉

matt-wills-valtech: Perhaps the best thing is to make a decision on this tomorrow, depending on how the API & UI dev goes, so I'll approve this 👍

matt-wills-valtech (approved):

PR #270, opened 2023-05-05
EIP1-5918 - New endpoint to update AED contact details

matt-wills-valtech:
This PR introduces a new endpoint to update the contact details of one or more AEDs, belonging to an application.

I've merged these PRs into this one:

* https://github.com/cabinetoffice/eip-ero-print-api/pull/276
* https://github.com/cabinetoffice/eip-ero-print-api/pull/273

**Endpoint definition**
image

**Schema**
image

### DIFF
--- a/src/main/resources/openapi/PrintAPIs.yaml
+++ b/src/main/resources/openapi/PrintAPIs.yaml
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Print APIs
-  version: '1.18.0'
+  version: '1.19.0'
   description: Print APIs
   contact:
     name: Krister Bone
@@ -659,7 +659,35 @@ paths:
         connectionType: VPC_LINK
         connectionId: '${vpc_connection_id}'
         httpMethod: POST
-
+    patch:
+      summary: Updates specific details within an Anonymous Elector Document (AED).
+      description: Updates specific details (e.g. the elector's phone number) within an Anonymous Elector Document (AED). Note that if more than one AED has been issued for an application, then all of them will be updated.
+      tags:
+        - Anonymous Elector Documents (AED)
+      requestBody:
+        $ref: '#/components/requestBodies/UpdateAnonymousElectorDocument'
+      responses:
+        '204':
+          description: AED updated successfully
+        '400':
+          $ref: '#/components/responses/400ErrorResponse'
+        '404':
+          $ref: '#/components/responses/404ErrorResponse'
+      operationId: update-anonymous-elector-document-details
+      security:
+        - eroUserCognitoUserPoolAuthorizer: [ ]
+      x-amazon-apigateway-integration:
+        type: HTTP_PROXY
+        uri: '${base_uri}/eros/{eroId}/anonymous-elector-documents'
+        requestParameters:
+          integration.request.path.eroId: method.request.path.eroId
+        responseParameters:
+          method.response.header.Access-Control-Allow-Headers: '''Content-Type,X-Amz-Date,Authorization,X-Api-Key'''
+          method.response.header.Access-Control-Allow-Methods: '''*'''
+          method.response.header.Access-Control-Allow-Origin: '''*'''
+        connectionType: VPC_LINK
+        connectionId: '${vpc_connection_id}'
+        httpMethod: PATCH
     get:
       summary: Returns the Anonymous Elector Documents for an Anonymous application
       description: Returns the Anonymous Elector Documents for an Anonymous application
@@ -1631,6 +1659,29 @@ components:
         - electoralRollNumber
         - deliveryAddressType
 
+    UpdateAnonymousElectorDocumentRequest:
+      title: UpdateAnonymousElectorDocumentRequest
+      description: Request body to update specific details of an Anonymous Elector Document (AED).
+      type: object
+      properties:
+        sourceReference:
+          type: string
+          example: 1f0f76a9a66f438b9bb33070
+          description: Reference in the source application that this Anonymous Elector Document is for.
+        email:
+          type: string
+          maxLength: 1024
+          format: email
+          description: The applicant's email address. This will only be updated if a non-null value is provided.
+          example: fred.blogs@some-domain.co.uk
+        phoneNumber:
+          type: string
+          maxLength: 50
+          description: The applicant's phone number. This will only be updated if a non-null value is provided.
+          example: 01234 567890
+      required:
+        - sourceReference
+
     PreSignedUrlResourceResponse:
       title: PreSignedUrlResourceResponse
       description: An object detailing the pre-signed URL to use in order to retrieve the resource
@@ -1818,6 +1869,12 @@ components:
         application/json:
           schema:
             $ref: '#/components/schemas/ReIssueAnonymousElectorDocumentRequest'
+    UpdateAnonymousElectorDocument:
+      description: Request body to update specific details belonging to one or more Anonymous Elector Documents (AEDs)
+      content:
+        application/json:
+          schema:
+            $ref: '#/components/schemas/UpdateAnonymousElectorDocumentRequest'
 
   securitySchemes:
     eroUserCognitoUserPoolAuthorizer:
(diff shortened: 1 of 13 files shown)


src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1631,6 +1665,29 @@ components:
         - electoralRollNumber
         - deliveryAddressType
 
+    UpdateAnonymousElectorDocumentRequest:
+      title: UpdateAnonymousElectorDocumentRequest
+      description: Request body to update specific details on one or more Anonymous Elector Documents (AEDs). Note that explicitly setting any of the values to an empty string ("") indicates it should be removed. Otherwise, omitting the field means it will not be updated.
+      type: object
+      properties:
+        sourceReference:
```
vishalgupta-valtech, line 1673: I think this `sourceReference` should be part of the path and not request body. 

`/eros/{eroId}/anonymous-elector-documents/{$applicationId}/`

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1631,6 +1665,29 @@ components:
...
+          type: string
+          example: 1f0f76a9a66f438b9bb33070
+          description: Reference in the source application that this Anonymous Elector Document is for.
+        email:
+          type: string
+          maxLength: 1024
+          format: email
+          description: The applicant's email address. This will only be updated if a non-null value is provided. Note that it will be removed if an empty string ("") is sent (if applicable).
+          example: fred.blogs@some-domain.co.uk
```
vishalgupta-valtech, line 1682: Unsure if we ever want to *update* the email or phone to be blank?

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1631,6 +1665,29 @@ components:
...
+        phoneNumber:
+          type: string
+          maxLength: 50
+          description: The applicant's phone number. This will only be updated if a non-null value is provided. Note that it will be removed if an empty string ("") is sent (if applicable).
+          example: 01234 567890
+      required:
+        - sourceReference
```
vishalgupta-valtech, line 1689: What happens when a call to this endpoint is made without email and phone in body? Are they both set to blank/empty. Don't think that should be allowed

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -659,7 +659,41 @@ paths:
         connectionType: VPC_LINK
         connectionId: '${vpc_connection_id}'
         httpMethod: POST
-
+    patch:
+      summary: Updates specific details within an Anonymous Elector Document (AED).
+      description: Updates specific details (e.g. the elector's phone number) within an Anonymous Elector Document (AED). Note that if more than one AED has been issued for an application, then all of them will be updated.
```
vishalgupta-valtech, line 664: Minor: Updates specific `contact` details of an AED...

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -659,7 +659,41 @@ paths:
...
+        $ref: '#/components/requestBodies/UpdateAnonymousElectorDocument'
+      responses:
+        '201':
+          description: created
+          headers:
+            Content-Disposition:
+              description: Content-Disposition header
+              schema:
+                type: string
+                example: attachment; filename="anonymous-elector-document-DGFF4E6420YHWPJYX560.pdf"
```
vishalgupta-valtech, line 677: I don't think this is correct, we aren't returning any pdfs for this patch endpoint? It should just be No Content (204) or something

src/main/resources/openapi/PrintAPIs.yaml
matt-wills-valtech: @NathanRussellValtech and I have debated this and believe me it's not easy to decide on the best approach! 

Note that `applicationId` is not a unique identifier for an AED, so I don't think that exact path works. Possibly `/eros/{eroId}/anonymous-elector-documents/applications/{$applicationId}/` might be better, but that goes against the existing pattern, which uses a query param (which is fine for GET methods).

The existing POST endpoints have `sourceReference` in the request body, so I've decided to follow suit here.

src/main/resources/openapi/PrintAPIs.yaml
matt-wills-valtech: I need to check the validity of this use case, but in theory an elector could want us to remove the email/phone number we have on record.

src/main/resources/openapi/PrintAPIs.yaml
matt-wills-valtech: The current plan is that nothing will happen (you explicitly have to set them to `""` to remove them, but I need to check that use case). 

Alternatively, if nothing is sent in the request, then we could return a 400. That might be better actually 🤔

src/main/resources/openapi/PrintAPIs.yaml
matt-wills-valtech: I don't want to be too specific in case other fields are added in future.

src/main/resources/openapi/PrintAPIs.yaml
matt-wills-valtech: Oops, I'll fix this.

src/main/resources/openapi/PrintAPIs.yaml
vishalgupta-valtech: To me whatever gets updated _(phone and email)_ goes to body of `PATCH` request. Yeah `/eros/{eroId}/anonymous-elector-documents/applications/{$applicationId}/` seems better than current endpoint

src/main/resources/openapi/PrintAPIs.yaml
vishalgupta-valtech: Yeah throwing `400` error would be better when both contact details values are not present in body.

src/main/resources/openapi/PrintAPIs.yaml
matt-wills-valtech: @NathanRussellValtech - any thoughts on this?

src/main/resources/openapi/PrintAPIs.yaml
matt-wills-valtech: I plan to do that, but I haven't mentioned it in the spec yet. I'm going to wait until we have consensus re the applicationId in the path vs body.

src/main/resources/openapi/PrintAPIs.yaml
matt-wills-valtech: I've removed this as I think it's unlikely or very rare. If absolutely necessary, a support request could be raised, or the ERO could update the phone number to something like 0777777777!

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -659,7 +659,35 @@ paths:
...
+    patch:
+      summary: Updates specific details within an Anonymous Elector Document (AED).
+      description: Updates specific details (e.g. the elector's phone number) within an Anonymous Elector Document (AED). Note that if more than one AED has been issued for an application, then all of them will be updated.
+      tags:
+        - Anonymous Elector Documents (AED)
+      requestBody:
+        $ref: '#/components/requestBodies/UpdateAnonymousElectorDocument'
+      responses:
+        '204':
+          description: AED updated successfully
```
vishalgupta-valtech, line 671: Minor: All AEDs with given `sourceReference` updated successfully as it could update more than 1 AEDs

matt-wills-valtech: I've just gone for "Updated successfully" :)

I know I briefly mentioned that more than one AED may be updated, but it feels like an implementation detail. Aside from anything else, I assume the vast majority of applications will only have one AED.

src/main/resources/openapi/PrintAPIs.yaml
NathanRussellValtech: I'm not going to lose any sleep one way or the other. Neither option is perfect, and short of implementing the path as something like `eros/{eroId}/anonymous-elector-documents/applications/{$applicationId}/` (which wont match our other endpoints, and will no doubt cause more discussion amongst us as to what is "best") - I'd be happy with either approach.

If I were to be pushed for an answer I'd perhaps go with the `sourceReference` in the body - but I really dont mind one way or the other ¯\_(ツ)_/¯

vishalgupta-valtech (approved):

matt-wills-valtech:
I'm going to hold off merging this until after the release.

PR #271, opened 2023-05-05
EIP1-5594: EIP1-5854: Adds pagination response in controller + updates integration tests

vishalgupta-valtech:


src/main/kotlin/uk/gov/dluhc/printapi/mapper/aed/AnonymousSearchSummaryMapper.kt
```diff
@@ -10,13 +10,13 @@ import uk.gov.dluhc.printapi.database.entity.AnonymousElectorDocumentSummary as
 import uk.gov.dluhc.printapi.models.AedSearchSummary as AedSearchSummaryApi
 
 @Mapper(uses = [InstantMapper::class])
-interface AnonymousSearchSummaryMapper {
+abstract class AnonymousSearchSummaryMapper {
```
matt-wills-valtech, line 13: Out of interest, what's the reason for this change (and the protected abstract method below)?

matt-wills-valtech (approved):

src/main/kotlin/uk/gov/dluhc/printapi/mapper/aed/AnonymousSearchSummaryMapper.kt
vishalgupta-valtech: Intention was to show that protected function is used within the mapper impl itself and not by any of kotlin classes written by us. 
The function `toAedSearchSummaryResponse` internally uses this protected function. With interface we couldn't mark these as protected.
Here is the autogenerated class for this Mapper 
image

PR #275, opened 2023-05-09
EIP1-5612 / EIP1-5859 - Wired specificationBuilder into AnonymousElectorDocumentSearchService

NathanRussellValtech:
### Note
This PR builds upon (and targets) my [previous PR ](https://github.com/cabinetoffice/eip-ero-print-api/pull/272)which added the specificationBuilder
When that PR is merged to eip-main, this one can be re-targetted to eip-main

---

This PR refactors `AnonymousElectorDocumentSearchService` to use the new specification builder.
This completes the database and service level work; and now all we need to do in order to allow querying by application reference is to update the openAPI spec to define it as a query string param enum (along with surname), and update a mapper 👍


src/main/kotlin/uk/gov/dluhc/printapi/database/repository/AnonymousElectorDocumentSummaryRepository.kt
```diff
@@ -1,29 +1,12 @@
...
 import uk.gov.dluhc.printapi.database.entity.AnonymousElectorDocumentSummary
-import uk.gov.dluhc.printapi.database.entity.SourceType
 import java.util.UUID
 
 @Repository
 interface AnonymousElectorDocumentSummaryRepository :
     PagingAndSortingRepository<AnonymousElectorDocumentSummary, UUID>,
-    JpaSpecificationExecutor<AnonymousElectorDocumentSummary> {
-
-    fun findAllByGssCodeInAndSourceType(
```
NathanRussellValtech, line 17: These methods are no longer necessary or used, now that the service class uses  the repository `findAll` method with the build specification and pageable 👍

src/main/kotlin/uk/gov/dluhc/printapi/service/aed/AnonymousElectorDocumentSearchService.kt
```diff
@@ -22,51 +18,33 @@ class AnonymousElectorDocumentSearchService(
...
+            page = criteria.page,
+            pageSize = criteria.pageSize
+        )
+        val specification = specificationBuilder.buildSpecification(gssCodes, criteria)
+
+        val aedSummaries = anonymousElectorDocumentSummaryRepository.findAll(specification, pageRequest)
+            .map { anonymousSearchSummaryMapper.toAnonymousSearchSummaryDto(entity = it) }
+        return with(aedSummaries) {
+            AnonymousSearchSummaryResults(
+                page = if (totalPages > 0) criteria.page else totalPages,
```
NathanRussellValtech, line 39: The tests highlighted a small bug here, in the case where the query returns zero results. What this service used to return in that case was a `AnonymousSearchSummaryResults` with fields:
```
page: 1
pageSize: 100
totalPages: 0
totalResults: 0
results: []
```
This didnt feel quite right (ie. saying it was page 1 of 0 pages). How can it be page 1 of 0 ?
So I added the if expression to fix it.

src/test/kotlin/uk/gov/dluhc/printapi/service/aed/AnonymousElectorDocumentSearchServiceIntegrationTest.kt
```diff
@@ -1,168 +1,188 @@
...
+import uk.gov.dluhc.printapi.testsupport.assertj.assertions.Assertions.assertThat
 import uk.gov.dluhc.printapi.testsupport.testdata.aValidSourceReference
 import uk.gov.dluhc.printapi.testsupport.testdata.dto.aed.buildAnonymousSearchCriteriaDto
 import uk.gov.dluhc.printapi.testsupport.testdata.entity.buildAnonymousElectorDocument
+import uk.gov.dluhc.printapi.testsupport.testdata.model.buildElectoralRegistrationOfficeResponse
+import uk.gov.dluhc.printapi.testsupport.testdata.model.buildLocalAuthorityResponse
 import java.time.Instant
 import java.time.LocalDate
 
-@Deprecated("This is a temporary test, and will be removed when EIP1-5612 uses the specification builder in the search service")
```
NathanRussellValtech, line 16: The diff is a bit hard to read here - basically what I've done is taken the temporary test `AnonymousElectorDocumentSummarySpecificationBuilderIntegrationTest` (which was always temporary), and deleted it.
And I've created a new test `AnonymousElectorDocumentSearchServiceIntegrationTest` which has the exact same test scenarios that the previous test had, but instead of hitting the repository class directly, it hits the service class and method 👍

src/test/kotlin/uk/gov/dluhc/printapi/service/aed/AnonymousElectorDocumentSearchServiceTest.kt
```diff
@@ -38,141 +38,50 @@ internal class AnonymousElectorDocumentSearchServiceTest {
...
     private lateinit var anonymousSearchSummaryMapper: AnonymousSearchSummaryMapper
 
+    @Mock
+    private lateinit var specificationBuilder: AnonymousElectorDocumentSummarySpecificationBuilder
+
     @InjectMocks
     private lateinit var anonymousElectorDocumentSearchService: AnonymousElectorDocumentSearchService
 
     @Test
-    fun `should return empty Anonymous Elector Document summaries given null searchBy`() {
```
NathanRussellValtech, line 45: I've removed all but one of the previous unit test methods in this class (`AnonymousElectorDocumentSearchServiceTest`)

The implementation has changed, so clearly I needed to make some change to this test. Now that the implementation is using the specification builder it actually becomes hard to test. The key part of testing the implementation is the `Specification` which is built by the SpecificationBuilder.
I could setup mocks to return a `Specification` in various states to reflect querying by surname/appRef/nothing etc etc .... but in the context of unit testing the service it adds nothing

So I chose to have 1 simple test that basically just asserts that the various collaborators are called with the correct/expected values .... and then leave testing of the actual logic wrt returned values etc to the new `AnonymousElectorDocumentSearchServiceIntegrationTest` above ☝️

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/assertj/assertions/AnonymousSearchSummaryResultsAssert.kt
```diff
@@ -0,0 +1,74 @@
+package uk.gov.dluhc.printapi.testsupport.assertj.assertions
+
+import org.assertj.core.api.AbstractObjectAssert
+import org.assertj.core.api.Assertions.assertThat
+import uk.gov.dluhc.printapi.dto.aed.AnonymousSearchSummaryDto
+import uk.gov.dluhc.printapi.dto.aed.AnonymousSearchSummaryResults
+
+class AnonymousSearchSummaryResultsAssert(actual: AnonymousSearchSummaryResults?) :
```
NathanRussellValtech, line 8: Who doesn't love an assertJ custom assertion class? 😁

matt-wills-valtech (approved):

PR #276, opened 2023-05-09
EIP1-5925 - Update AED contact details - happy path

matt-wills-valtech:


src/test/kotlin/uk/gov/dluhc/printapi/rest/aed/UpdateAnonymousElectorDocumentByApplicationIdIntegrationTest.kt
```diff
@@ -92,4 +102,151 @@ internal class UpdateAnonymousElectorDocumentByApplicationIdIntegrationTest : In
...
+        wireMockService.stubEroManagementGetEroByEroId(eroResponse, ERO_ID)
+        val originalEmailAddress = aValidEmailAddress()
+        val originalPhoneNumber = aValidPhoneNumber()
+        val aed = buildAnonymousElectorDocument(
+            gssCode = GSS_CODE,
+            sourceReference = APPLICATION_ID,
+            contactDetails = buildAedContactDetails(email = originalEmailAddress, phoneNumber = originalPhoneNumber)
+        )
+        anonymousElectorDocumentRepository.save(aed)
+        Thread.sleep(2000)
```
vishalgupta-valtech, line 123: Can we not thread sleep this for just a second (same comment for other tests in this class)

src/test/kotlin/uk/gov/dluhc/printapi/rest/aed/UpdateAnonymousElectorDocumentByApplicationIdIntegrationTest.kt
```diff
@@ -92,4 +102,151 @@ internal class UpdateAnonymousElectorDocumentByApplicationIdIntegrationTest : In
...
+            .uri(URI_TEMPLATE, ERO_ID, APPLICATION_ID)
+            .bearerToken(getVCAnonymousAdminBearerToken(eroId = ERO_ID))
+            .contentType(APPLICATION_JSON)
+            .withBody(updateAedRequest)
+            .exchange()
+            .expectStatus().isNoContent
+
+        // Then
+        val updated = anonymousElectorDocumentRepository.findById(aed.id!!).get()
+        assertThat(updated.contactDetails!!.email).isEqualTo(newEmailAddress)
```
vishalgupta-valtech, line 148: Could use `AnonymousElectorDocumentCertificateAssert` to assert these values? (same comment for test below)

src/main/kotlin/uk/gov/dluhc/printapi/service/aed/AnonymousElectorDocumentService.kt
```diff
@@ -60,11 +60,19 @@ class AnonymousElectorDocumentService(
...
         with(updateAedDto) {
-            val anonymousElectorDocuments = getAnonymousElectorDocuments(eroId, sourceReference)
+            val gssCodes = eroService.lookupGssCodesForEro(eroId)
+            val anonymousElectorDocuments = getAnonymousElectorDocumentsSortedByDate(gssCodes, sourceReference)
             if (anonymousElectorDocuments.isEmpty()) {
                 throw CertificateNotFoundException(eroId, ANONYMOUS_ELECTOR_DOCUMENT, sourceReference)
             }
-            // TODO EIP1-5925
+            anonymousElectorDocuments.forEach {
+                if (valueHasChanged(email, it.contactDetails!!.email)) {
```
vishalgupta-valtech, line 69: Can we not just update an email and phone for the latest record?

src/main/kotlin/uk/gov/dluhc/printapi/service/aed/AnonymousElectorDocumentService.kt
```diff
@@ -100,4 +108,8 @@ class AnonymousElectorDocumentService(
         val contents = pdfFactory.createPdfContents(templateDetails)
         return PdfFile(PDF_FILE_NAME.format(certificateNumber), contents)
     }
+
+    private fun valueHasChanged(newValue: String?, existingValue: String?): Boolean {
+        return !newValue.isNullOrBlank() && newValue != existingValue
```
vishalgupta-valtech, line 113: 1) What if user is trying to set the email to `null/blank` and phone to some sensible value? 
This logic might not update/blank out the email for latest AED

2) Could support that by adding a unit test in service

vishalgupta-valtech (commented):
Do you NAH ? 😜

src/main/kotlin/uk/gov/dluhc/printapi/service/aed/AnonymousElectorDocumentService.kt
matt-wills-valtech: We think it's better to update all the AEDs. The contact details do not really belong to an AED per se (though they've been modelled that way) and I can't see any benefit in keeping out of date details.

src/test/kotlin/uk/gov/dluhc/printapi/rest/aed/UpdateAnonymousElectorDocumentByApplicationIdIntegrationTest.kt
matt-wills-valtech: Happy to do that for email and phone number 👍  

I could do it for most of the others, though the test for `dateUpdated` is different in each case.

src/main/kotlin/uk/gov/dluhc/printapi/service/aed/AnonymousElectorDocumentService.kt
vishalgupta-valtech: If we do want to update all the records, then does it make sense to check only the latest record and see if it has changed, and if yes then we change all email records (same for phone)?

matt-wills-valtech: I'm not so sure to be honest. Realistically there will usually only be one and at most two or three(?). Once they're retrieved from the DB, iterating over them is negligible in terms of performance and I still think it's best to check them all.

NathanRussellValtech (approved):
GS 👍

PR #286, opened 2023-06-14
EROPSPT-123: Change link in emails to use id instead of reference

kirstenland:
This will change the links in emails for both certificate-returned and certificate-failed-to-print emails - I assume both are supposed to have the application id rather than the application reference? The ticket only mentions the certificate returned case but the links are the same in both cases.

It also adds the application reference into the body of the certificate-returned email. The application reference was already in the email subject, and in the body of the certificate-failed-to-print email.


beccar97 (approved):
Assuming that `sourceReference` is the application id, then looks good to me

PR #291, opened 2023-09-01
EROPSPT-142 - API changes for Manage Certificates feature

scl-softwire:


src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1,7 +1,7 @@
 openapi: 3.0.0
 info:
   title: Print APIs
-  version: '1.20.0'
+  version: '1.21.0'
```
scl-softwire, line 4: The changes to this spec are breaking. Nonetheless, I've only used a minor version bump because the only consumer of the endpoints which have been removed was the manage certificates page, which was never released against the old spec.

PR #295, opened 2023-09-25
EROPSPT-169 - Send message to queue when statistics update is required

scl-softwire:


src/main/kotlin/uk/gov/dluhc/printapi/aspect/StatisticsUpdateAspect.kt
```diff
@@ -0,0 +1,71 @@
...
+        saved.sourceReference?.also {
+            statisticsUpdateService.triggerVoterCardStatisticsUpdate(it)
+        }
+    }
+
+    @AfterReturning(
+        pointcut = "execution(* org.springframework.data.repository.CrudRepository.save(..)) && args(anonymousElectorDocument)",
+        returning = "saved"
+    )
+    fun afterSaveTemporaryCertificate(joinPoint: JoinPoint, anonymousElectorDocument: AnonymousElectorDocument, saved: AnonymousElectorDocument) {
```
kirstenland, line 41: Is this named correctly?

src/main/kotlin/uk/gov/dluhc/printapi/aspect/StatisticsUpdateAspect.kt
```diff
@@ -0,0 +1,71 @@
...
+        statisticsUpdateService.triggerVoterCardStatisticsUpdate(saved.sourceReference)
+    }
+
+    @AfterReturning(
+        pointcut = "execution(* org.springframework.data.repository.CrudRepository.saveAll(..))",
+        returning = "saved"
+    )
+    fun afterSaveAll(joinPoint: JoinPoint, saved: Iterable<Any>) {
+        // Type is erased and AOP is passed an iterable containing any type
```
kirstenland, line 50: I'm not sure I understand this - what's AOP? What type is erased?

src/test/kotlin/uk/gov/dluhc/printapi/jobs/BatchPrintRequestsJobIntegrationTest.kt
```diff
@@ -0,0 +1,94 @@
...
+                    )
+                )
+            )
+        )
+        certificateRepository.save(certificate1)
+        certificateRepository.save(certificate2)
+        TestTransaction.flagForCommit()
+        TestTransaction.end()
+
+        // Clear rogue messages from the queue in order to make the test valid
```
kirstenland, line 74: Rogue maybe makes it sound more worrying than it is - presumably these messages come from creating the certificate in the first place? Is there a reason why this is needed here and not elsewhere?

kirstenland (commented):
Looks fairly reasonable to me.

src/main/kotlin/uk/gov/dluhc/printapi/aspect/StatisticsUpdateAspect.kt
scl-softwire: Hmm this comment came over from VCA where I ~copied~ got inspiration for this code from. AOP is "Aspect Oriented Programming", the comment is saying that while the `afterSave` methods have explicit types in the signatures this one has an `Any` type, so we need to check types explicitly. Honestly I don't think the comment adds anything - happy to remove?

src/test/kotlin/uk/gov/dluhc/printapi/jobs/BatchPrintRequestsJobIntegrationTest.kt
scl-softwire: Yep, will remove "rogue" and add detail. Calling `certificateRepository.save` is what triggers these messages to go out, so it's only needed on tests where we call that explicitly.

src/main/kotlin/uk/gov/dluhc/printapi/aspect/StatisticsUpdateAspect.kt
kirstenland: Is it worth maybe mentioning how we know all objects are of the same type?

kirstenland (approved):

PR #296, opened 2023-09-26
EROPSPT-171: Add endpoints for VCA to call with stats info for VACs and AEDs

kirstenland:


src/main/kotlin/uk/gov/dluhc/printapi/rest/CertificateStatisticsController.kt
```diff
@@ -0,0 +1,44 @@
...
+import uk.gov.dluhc.printapi.database.entity.SourceType
+import uk.gov.dluhc.printapi.mapper.CertificateStatisticsStatusMapper
+import uk.gov.dluhc.printapi.models.StatisticsResponse
+import uk.gov.dluhc.printapi.service.CertificateFinderService
+import uk.gov.dluhc.printapi.service.temporarycertificate.TemporaryCertificateFinderService
+
+private val logger = KotlinLogging.logger {}
+
+@RestController
+@CrossOrigin
```
kirstenland, line 20: Not sure what the `@CrossOrigin`s do and if we need them for this controller.

src/main/kotlin/uk/gov/dluhc/printapi/service/aed/AnonymousElectorDocumentService.kt
```diff
@@ -86,6 +92,14 @@ class AnonymousElectorDocumentService(
             .map { anonymousElectorDocumentMapper.mapToAnonymousElectorDocumentDto(it) }
     }
 
+    @Transactional(readOnly = true)
+    fun getAnonymousElectorDocumentsByApplicationId(applicationId: String): List<AnonymousElectorDocumentDto> {
+        return anonymousElectorDocumentRepository.findBySourceTypeAndSourceReference(
```
scl-softwire, line 97: I think given the data model we have, this code is right
But I am still sad about having a `sourceType` field and then splitting the data into different _tables_ based on that field!

src/main/kotlin/uk/gov/dluhc/printapi/rest/CertificateStatisticsController.kt
scl-softwire: Probably not needed - we're not expecting requests to come from a browser to this endpoint. Although I'm not completely clear about how this interacts with the CORS config we've set at the API Gateway level.

Fine to leave.

src/main/kotlin/uk/gov/dluhc/printapi/rest/CertificateStatisticsController.kt
```diff
@@ -0,0 +1,44 @@
...
+        logger.debug { "Collecting VAC statistics for applicationId $applicationId" }
+
+        val certificate = certificateFinderService.getCertificate(SourceType.VOTER_CARD, applicationId)
+        val temporaryCertificates = temporaryCertificateFinderService.getTemporaryCertificates(
+            SourceType.VOTER_CARD,
+            applicationId,
+        )
+
+        return StatisticsResponse(
+            certificateStatus = statusMapper.fromEntityPrintRequestStatus(certificate.status!!),
```
scl-softwire, line 39: Have we checked that this coercion is safe?

scl-softwire (approved):
Looks good modulo one question about null-safety and confirming that the APIG config is correct

src/main/kotlin/uk/gov/dluhc/printapi/rest/CertificateStatisticsController.kt
kirstenland: This field is marked as NotNull on the entity so I think we are okay.

PR #297, opened 2023-10-17
EROPSPT-169 - Send statistics update after transaction commit

scl-softwire:


kirstenland (commented):
Looks good to me - we should probably leave a comment somewhere because this is going to be highly confusing to someone at some stage.

kirstenland (approved):

PR #308, opened 2024-07-03
EROPSPT-336: Return error and upload response to S3 when temporary VAC too large

kirstenland:
I don't know if either or both these changes would be worth doing or would be likely to be approved for release:

- Not returning anything more than 9MB should stop VAC from getting stuck hopefully - this seems sensible
- Uploading any temporary certificates that are too big to S3 would make it much easier for us to access them - we'd want to revert this change after Enhanced Support ends and clear out this folder in S3. This seems more risky and would require giving print api write permissions to the bucket, but derisks having to send temporary certificates over until Thursday - otherwise I'm not sure whether we'll be able to keep up with the workload.

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/rest/TemporaryCertificateController.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/rest/TemporaryCertificateController.kt
@@ -96,7 +96,7 @@ class TemporaryCertificateController(
         )
         return temporaryCertificateService.generateTemporaryCertificate(eroId, dto).also {
             if (it.contents.size > MAX_RESPONSE_FILE_SIZE && generateTemporaryCertificateRequest.allowLargeResponse != true) {
-                val s3Path = "temporary_certificates/${dto.gssCode}/${dto.applicationReference}"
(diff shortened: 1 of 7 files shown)


src/main/kotlin/uk/gov/dluhc/printapi/rest/TemporaryCertificateController.kt
```diff
@@ -84,6 +95,20 @@ class TemporaryCertificateController(
             userId
         )
         return temporaryCertificateService.generateTemporaryCertificate(eroId, dto).also {
+            if (it.contents.size > MAX_RESPONSE_FILE_SIZE && generateTemporaryCertificateRequest.allowLargeResponse != true) {
+                val s3Path = "temporary_certificates/${dto.gssCode}/${dto.applicationReference}.pdf"
```
scl-softwire, line 99: I think the bucket has versioning enabled, and they probably shouldn't be generating multiple certificates for one application anyway, but could we just add a unique ID into this path anyway?

scl-softwire (approved):

scl-softwire (approved):

PR #311, opened 2024-08-08
EROPSPT-339: Upload temporary certificates to S3

kirstenland:
I haven't done this in a backwards compatible way as temporary certificates aren't downloaded very often and I'm expecting this all to be released at the same time.


scl-softwire (approved):
Nice!

PR #314, opened 2024-10-23
EROPSPT-396: Reset initial removal date when reissuing AED

kirstenland:
I haven't covered the final data retention date here, only the initial date.

~Having had a call with policy, it is no longer clear whether this date should be reset, so I've made the PR into a draft and will hold off merging until we have more clarity.~
Following further calls with policy, we will be going ahead with this change.


ollylarkin-softwire (commented):
Looks good to me! Can't see the original ticket so will leave it up to Alex to approve or not

Out of interest, why only the initial date in this change?

kirstenland:
> Out of interest, why only the initial date in this change?

I'm keeping them separate for now so we can get this change through sooner if there are complexities with the final retention date - we know there is something to sort out with the certificate photo and we are waiting on a call with policy to confirm our understanding is correct. I expect to make both changes soon though.

aleyipsoftwire (approved):

PR #316, opened 2024-11-01
EROPSPT-405: Exclude AEDs from endpoints once past initial data retention

kirstenland:
At present, AEDs past the initial data retention will show up on the GetAnonymousElectorDocument endpoint, but will cause errors as some of the fields are null. We can either exclude them from such endpoints or allow these fields to be null - excluding them is likely to be the simpler solution.

This is required now because if there are multiple reissues of a certificate, and some have been through initial data retention while others haven't, then currently attempting to view the AED on the Manage AEDs tab will cause an error.

This change also excludes AEDs that have been through initial data retention from being updated by the update endpoints, which is important as they could add back contact details that was deleted during initial data retention.

These AEDs are also excluded from the reissue endpoint and the photo endpoint, to prevent any errors, though this is probably less important.

In the future, we are likely to want to display AEDs beyond the 15 month period, all the way up to 10 years. That may require undoing some of this, but that won't be very difficult and it is too early to work out what the exact technical requirements will be.


ollylarkin-softwire (approved):
Lgtm!

PR #324, opened 2024-12-03
EIP1-11169: Update service to support req/res messages from/to Applic…

EsengulGungor:
…ations API

## Describe your changes
We will need to enable the Applications API to send messages to any existing queues in Print API and receive a response back on its own SQS queue.

As part of this, we will need to configure the Print API service so it supports receiving requests and sending responses from/to the Applications API.
## Issue ticket number and link

https://mhclgdigital.atlassian.net/browse/EIP1-11169
## Checklist before requesting a review

- [x ] I double checked that ACs on the ticket are met by this code update
- [ x] I have added testing steps to the ticket
- [ x] I have updated the relevant yml versions
- [ ] I have formatted the code
- [ x] I have added tests to new code and updated existing tests where needed
- [ ] I have added any corresponding changes to the UI portal

### DIFF
--- a/src/main/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintRequestBatchMessageListener.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintRequestBatchMessageListener.kt
@@ -24,7 +24,10 @@ class ProcessPrintRequestBatchMessageListener(
 
             val certificates = processPrintBatchService.processBatch(batchId, printRequestCount)
             certificates.forEach {
-                statisticsUpdateService.triggerVoterCardStatisticsUpdate(it.sourceReference!!)
+                if (payload.isFromApplicationsApi == true)
+                    statisticsUpdateService.triggerApplicationStatisticsUpdate(it.sourceReference!!)
+                else
+                    statisticsUpdateService.triggerVoterCardStatisticsUpdate(it.sourceReference!!)
             }
 
             logger.info("Successfully processed print request for batchId: $batchId")
--- a/src/main/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintRequestBatchMessageListener.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintRequestBatchMessageListener.kt
@@ -24,10 +24,7 @@ class ProcessPrintRequestBatchMessageListener(
 
             val certificates = processPrintBatchService.processBatch(batchId, printRequestCount)
             certificates.forEach {
-                if (payload.isFromApplicationsApi == true)
-                    statisticsUpdateService.triggerApplicationStatisticsUpdate(it.sourceReference!!)
-                else
-                    statisticsUpdateService.triggerVoterCardStatisticsUpdate(it.sourceReference!!)
+                statisticsUpdateService.updateStatistics(it.sourceReference!!, payload.isFromApplicationsApi)
             }
 
             logger.info("Successfully processed print request for batchId: $batchId")
--- a/src/main/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintResponseMessageListener.kt
+++ b/src/main/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintResponseMessageListener.kt
@@ -24,10 +24,7 @@ class ProcessPrintResponseMessageListener(
     override fun handleMessage(@Payload payload: ProcessPrintResponseMessage) {
         logger.info { "Begin processing PrintResponse with requestId ${payload.requestId}" }
         printResponseProcessingService.processPrintResponse(payload)?.also {
-            if (payload.isFromApplicationsApi == true)
-                statisticsUpdateService.triggerApplicationStatisticsUpdate(it.sourceReference!!)
-            else
-                statisticsUpdateService.triggerVoterCardStatisticsUpdate(it.sourceReference!!)
+            statisticsUpdateService.updateStatistics(it.sourceReference!!, payload.isFromApplicationsApi)
         }
     }
 }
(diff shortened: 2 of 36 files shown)


src/main/resources/openapi/sqs/vca-api-sqs-messaging-erop.yaml
```diff
@@ -267,6 +276,17 @@ components:
           example: 1f0f76a9a66f438b9bb33070
       required:
         - voterCardApplicationId
+    UpdateApplicationStatisticsMessage:
+      title: TriggerApplicationStatisticsUpdateMessage
+      type: object
+      description: SQS Message for initiating a statistics update for an application api
```
TomosSlater, line 282: I don't think this description is correct, how about "SQS message for initiating a statistics update for an application"

src/main/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintRequestBatchMessageListener.kt
```diff
@@ -24,7 +24,10 @@ class ProcessPrintRequestBatchMessageListener(
 
             val certificates = processPrintBatchService.processBatch(batchId, printRequestCount)
             certificates.forEach {
-                statisticsUpdateService.triggerVoterCardStatisticsUpdate(it.sourceReference!!)
+                if (payload.isFromApplicationsApi == true)
```
TomosSlater, line 27: I don't think we need the == true here, can just have `if(payload.isFromApplicationsApi)`

src/main/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintResponseMessageListener.kt
```diff
@@ -24,7 +24,10 @@ class ProcessPrintResponseMessageListener(
     override fun handleMessage(@Payload payload: ProcessPrintResponseMessage) {
         logger.info { "Begin processing PrintResponse with requestId ${payload.requestId}" }
         printResponseProcessingService.processPrintResponse(payload)?.also {
-            statisticsUpdateService.triggerVoterCardStatisticsUpdate(it.sourceReference!!)
+            if (payload.isFromApplicationsApi == true)
```
TomosSlater, line 27: We're reusing this `if(payload.isFromApplicationsApi)` logic in a bunch of places, can we instead have this handled in the statistics update service to avoid replicating code

src/test/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintRequestBatchMessageListenerIntegrationTest.kt
```diff
@@ -94,6 +94,71 @@ internal class ProcessPrintRequestBatchMessageListenerIntegrationTest : Integrat
         }
     }
 
+    @Test
```
TomosSlater, line 97: This test is identical to the one above except for the `isFromApplicationsApi` bit and the assert around the update statistics message, can we merge the two of them into a parameterised test instead please

src/test/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintResponseFileMessageListenerIntegrationTest.kt
```diff
@@ -46,4 +46,37 @@ internal class ProcessPrintResponseFileMessageListenerIntegrationTest : Integrat
 
         // todo assert db updates after completing service implementation
     }
+
+    @Test
```
TomosSlater, line 50: Can we parameterise this test, too, please

src/test/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintResponseMessageListenerIntegrationTest.kt
```diff
@@ -74,6 +74,55 @@ internal class ProcessPrintResponseMessageListenerIntegrationTest : IntegrationT
         }
     }
 
+    @Test
```
TomosSlater, line 77: Can we parameterise this test also

src/test/kotlin/uk/gov/dluhc/printapi/messaging/SendApplicationToPrintMessageListenerIntegrationTest.kt
```diff
@@ -101,6 +101,59 @@ internal class SendApplicationToPrintMessageListenerIntegrationTest : Integratio
         }
     }
 
+    @Test
```
TomosSlater, line 104: Can we parameterise this one also, please

src/main/kotlin/uk/gov/dluhc/printapi/service/StatisticsUpdateService.kt
```diff
@@ -18,4 +20,14 @@ class StatisticsUpdateService(
             )
         )
     }
+
+    fun triggerApplicationStatisticsUpdate(applicationId: String) {
```
TomosSlater, line 24: We should add some tests for this in `StatisticsUpdateServiceTest`

TomosSlater (changes requested):
Looks good, have added a few mostly minor comments

src/main/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintRequestBatchMessageListener.kt
EsengulGungor: But it is Boolean?, so cany just check if(payload.isFromApplicationsApi)

TomosSlater: Ah yeah you're right, we do want this given it can be null

src/test/kotlin/uk/gov/dluhc/printapi/messaging/ProcessPrintResponseFileMessageListenerIntegrationTest.kt
```diff
@@ -11,10 +14,13 @@ import uk.gov.dluhc.printapi.testsupport.testdata.entity.buildCertificate
...
 import java.util.concurrent.TimeUnit
 
+@TestInstance(Lifecycle.PER_CLASS)
 internal class ProcessPrintResponseFileMessageListenerIntegrationTest : IntegrationTest() {
 
-    @Test
-    fun `should fetch remote print response file and delete it`() {
+    // @Test
+    @ParameterizedTest
+    @MethodSource("uk.gov.dluhc.printapi.testsupport.testdata.ApplicationsApiTestSource#isFromApplicationsApiTestSource")
```
TomosSlater, line 22: I think this is a bit overkill, can we use a CSV source instead? Should be able to test true/false/null this way. Same for the other tests where you've added this

TomosSlater (approved):
Looks good, thanks Esen!

PR #328, opened 2025-02-28
EROPSPT-397: Return AEDs post-initial retention period for AED search endpoint

aleyipsoftwire:
## Describe your changes

To keep the PR small-ish, I'm splitting the changes up into multiple PRs. This one is for returning AEDs that have had their first retention period data removed when searched, i.e., showing it in the "Manage AEDs" page.

There will need to be changes to the UI as well, and further changes to the backend for the page that shows individual applications.

## Issue ticket number and link

https://mhclgdigital.atlassian.net/browse/EROPSPT-397

## Checklist before requesting a review

- [ ] I double checked that ACs on the ticket are met by this code update
- [ ] I have added testing steps to the ticket
- [ ] I have updated the relevant yml versions
- [ ] I have formatted the code
- [ ] I have added tests to new code and updated existing tests where needed
- [ ] I have added any corresponding changes to the UI portal

### DIFF
--- a/src/main/resources/db/changelog/ddl/0032_EROPSPT-397_update_aed_summary_view_include_aeds_after_15_months.xml
+++ b/src/main/resources/db/changelog/ddl/0032_EROPSPT-397_update_aed_summary_view_include_aeds_after_15_months.xml
@@ -22,7 +22,8 @@
             contact.first_name,
             contact.surname,
             contact.sanitized_surname,
-            addr.postcode
+            addr.postcode,
+            aed.initial_retention_data_removed
             FROM anonymous_elector_document aed
             JOIN aed_contact_details contact ON aed.aed_contact_details_id = contact.id
             LEFT JOIN address addr on contact.address_id = addr.id
--- a/src/main/resources/db/changelog/ddl/0032_EROPSPT-397_update_aed_summary_view_include_aeds_after_15_months.xml
+++ b/src/main/resources/db/changelog/ddl/0032_EROPSPT-397_update_aed_summary_view_include_aeds_after_15_months.xml
@@ -1,0 +1,81 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<databaseChangeLog
+    xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-4.1.xsd">
+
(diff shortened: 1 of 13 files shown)


src/main/resources/db/changelog/ddl/0032_EROPSPT-397_update_aed_summary_view_include_aeds_after_15_months.xml
```diff
@@ -0,0 +1,81 @@
...
+            aed.application_reference,
+            aed.issue_date,
+            aed.date_created,
+            contact.first_name,
+            contact.surname,
+            contact.sanitized_surname,
+            addr.postcode
+            FROM anonymous_elector_document aed
+            JOIN aed_contact_details contact ON aed.aed_contact_details_id = contact.id
+            LEFT JOIN address addr on contact.address_id = addr.id
```
aleyipsoftwire, line 28: The only update to this view is changing the `JOIN` to the address table a `LEFT JOIN`, so it includes any AEDs that don't have address info anymore.

src/test/kotlin/uk/gov/dluhc/printapi/testsupport/testdata/entity/AnonymousElectorDocumentSummaryBuilder.kt
```diff
@@ -50,28 +49,3 @@ fun buildAnonymousElectorDocumentSummaryEntity(
         postcode = postcode
     )
 }
-
-fun buildAnonymousElectorDocumentSummaryViewFromAedEntity(
```
aleyipsoftwire, line 54: This wasn't used anywhere

src/main/resources/openapi/PrintAPIs.yaml
```diff
@@ -1893,7 +1894,7 @@ components:
           example: Blogs
         postcode:
           type: string
-          description: The applicant's postcode
+          description: The applicant's postcode. A null postcode indicates that the AED has passed it's initial retention period.
```
scl-softwire, line 1897: nit: it's --> its

aleyipsoftwire: Done

scl-softwire (commented):
Using the postcode does work, but feels a bit "off" to me. 

It's more effort, but how about pulling the `initialRetentionDataRemoved` field through to the summary view and using that to determine the status?

aleyipsoftwire:
> Using the postcode does work, but feels a bit "off" to me.
> 
> It's more effort, but how about pulling the `initialRetentionDataRemoved` field through to the summary view and using that to determine the status?

Yea I think that's fair enough.

I did consider doing that, but thought that basing the status on a single variable was quite nice in a way, it means I don't have to consider the case were (somehow) the front-end receives status=EXPIRED and postcode != null or vice-versa.

scl-softwire (approved):

